PR #473, opened 2014-08-15
Add support for the new quality options

patrickli:
This adds support for the following new quality options:

 * jpeg_quality
 * png_compression_level
 * png_compression_filter


lsmith77:
needs a rebase .. also tests would be good.

makasim:
We really need this feature but I am not sure about the implementation.

patrickli:
Rebased. However I am not sure if unit test is any helpful since these options are just passed own to imagine.

makasim:
hm, the PR looks unfinished. there is not any tests and you just defined properties but do not use them anywhere

lsmith77:
@makasim this PR just adds options for Imagine, so they are passed along with the other options.

@patrickli in terms of tests you just need to ensure that they are properly set https://github.com/liip/LiipImagineBundle/tree/master/Tests/DependencyInjection

makasim:
@lsmith77 thanks for the explanation

patrickli:
Unit test added.

makasim:
@patrickli thanks

Issue #516, opened 2014-10-24
Bug at cache lookup


ghostika:
I came across with an interesting bug with the cache lookup.
My setup:
storage on S3
cache is Redis

The main problem is the first `/` character. 

In my example, I have a file with this path:
`/userImage/234/13123jfk24.jpg`

Using the twig filter, the CacheManager will look in the Redis cache with this path, and it will be a miss, so it's generate a resolving url. After this, the resolving url will be called, and this is where the problem comes from. In the ImagineController:filterAction the cachemanager is looking for the path, but at this time, the path will be this: `userImage/234/13123jfk24.jpg` without the first `/` because that is the route parameter delimiter and after resolve, the entry will be stored under the cachekey for the path without `/`. With the FileSystemLoader, it's not a problem, cause it takes care of this, and append a `/`.
Part of my workflow is that if a user creates a new cut of an avatar, I remove the stored one from S3, at the same time deleting the Redis entry( actually this one is not happening cause it's not finding the key ) and then a new image should be generated.
 At this point I made a shortcut, at the remove process I ltrim() the path with the `/` so it will delete the right key from redis but this one doesn't solve the problem of the double lookup.

digitalkaoz:
could you create a PR for that?

ghostika:
Done

PR #541, opened 2015-01-05
Fix for cache lookup bug

ghostika:
Fix for the #516 issue.

### DIFF
--- a/Controller/ImagineController.php
+++ b/Controller/ImagineController.php
@@ -145,7 +145,7 @@ class ImagineController
 
             return new RedirectResponse($this->cacheManager->resolve($rcPath, $filter), 301);
         } catch (RuntimeException $e) {
-            throw new \RuntimeException(sprintf('Unable to create image for path "%s" and filter "%s". Message was "%s"', $hash.$path, $filter, $e->getMessage()), 0, $e);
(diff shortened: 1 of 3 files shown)


makasim:
ci build failed - https://travis-ci.org/liip/LiipImagineBundle/jobs/46003325. Could you please fix it?

akostottos:
Fixed

ghostika:
Fixed

lsmith77:
@makasim good to merge now?

Controller/ImagineController.php
```diff
@@ -67,6 +67,8 @@ public function __construct(
      */
     public function filterAction(Request $request, $path, $filter)
     {
+        $path = '/'.ltrim($path, '/');
```
makasim: what if the path is an id in a storage, I mean it is not real path on the filesystem but something like mongo id or so.

IN this case it will break all such storages. Maybe we have to put it only to those resolvers\loaders which rely on real filesystem, like WebPathResolver.

Also if the path contains a `/` symbol it must be urlencoded and added to the resolve path.

makasim: @ghostika ping

makasim:
The change may break existing resolvers, especially ones which are not filesystem based, mongo etc

PR #559, opened 2015-01-27
Add a short introductory paragraph about the bundle

javiereguiluz:
Now that this documentation is integrated in symfony.com, it's better to provide some information about what does the bundle do.

**Before**

before

**After**

after


makasim:
thanks.

PR #561, opened 2015-01-29
fix upscale size not being calculated correctly

scuben:
I had trouble to upsize a 50x50 image to 201x201 because of the float to int casting in the Box Class.

$originWidth * $ration produces a float(201). Passing that to the Box constructor which will cast it to int will produce int(200) which is not what i wanted/configured. Rounding them before passing to the Box constructor solves this issue.


Imagine/Filter/Loader/UpscaleFilterLoader.php
```diff
@@ -35,7 +35,7 @@ public function load(ImageInterface $image, array $options = array())
 
             $ratio = $widthRatio > $heightRatio ? $widthRatio : $heightRatio;
 
-            $filter = new Resize(new Box($origWidth * $ratio, $origHeight * $ratio));
+            $filter = new Resize(new Box(round($origWidth * $ratio), round($origHeight * $ratio)));
```
makasim: We are dealing with pixels this is why we have to round it, did I get right?

scuben: Yes. The problem is that if we just pass the float value to the Box object it will be casted in the Box constructor which will lead to this issue that float(201) become float(200) which is not what i configured in the filter.

Tests/Imagine/Filter/Loader/UpscaleFilterLoaderTest.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @covers Liip\ImagineBundle\Imagine\Filter\Loader\UpscaleFilterLoader
+ */
+class UpscaleFilterLoaderTest extends AbstractTest
+{
+    public function testLoad()
+    {
+        $loader = new UpscaleFilterLoader();
+        $imagine = new Imagine();
+        $image = $imagine->open(__DIR__.'/../../../Fixtures/assets/square.png');
```
makasim: could you name an image so that it contains width and height square-200x200.png. and rename the test so it is clear what it test.

scuben:
Did the rename and added an explaination in the class doc-block

Tests/Imagine/Filter/Loader/FloatToIntCastByRoundUpscaleFilterLoaderTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Imagine\Gd\Image;
+use Imagine\Gd\Imagine;
+use Liip\ImagineBundle\Imagine\Filter\Loader\UpscaleFilterLoader;
+use Liip\ImagineBundle\Tests\AbstractTest;
+
+/**
+ * @covers Liip\ImagineBundle\Imagine\Filter\Loader\UpscaleFilterLoader
+ *
+ * Due to int casting in Imagine\Image\Box which can lead to wrong pixel
+ * numbers ( e.g. float(201) casted to int(200) ). Solved by round the
```
makasim: the phrase `float(201) casted to int(200)` does not look correct. float 201 is cast to int 201. there must be a decimal part which we loose. something like: `float(200.7) casted to int(200)`. 

Could you confirm that?

scuben: This is what is rely strange. By executing the test on master while var_dump'ing ratio and the arguments passed to the constructor i get the following:

    var_dump($ratio, $origHeight * $ratio);
    float(4.02)
    float(201)

PR #565, opened 2015-02-21
fix invalid in_array

digitalkaoz:
under some circumstances it seems `$index` is not always an array. therefore we do the default stuff


lsmith77:
why not just cast to an array before hand?

digitalkaoz:
obviously simple, so yeah :)

makasim:
thanks.

digitalkaoz:
can we tag a new release?

makasim:
Sure: https://github.com/liip/LiipImagineBundle/releases/tag/1.2.3

digitalkaoz:
merci :)

PR #567, opened 2015-02-28
use isDefined method for OptionsResolver instead of isKnown  (when possible)

adam187:
isKnown is deprecated since version 2.6


makasim:
thanks.

PR #574, opened 2015-03-20
Remove some overhead at ResolveCacheCommand.php

igormukhingmailcom:
When more than one filter used there are some overhead with getting same binary content several times.
So, I propose use that optimization.


igormukhingmailcom:
Sorry, I did not realize that there are defferent loaders in filters.

Issue #587, opened 2015-04-07
Add image rotate filter


bocharsky-bw:
Hi there!

@lsmith77 I think, there is missing rotate filter, for example, to rotate image 90 (45, 180, etc.) degrees? As for me, the `auto_rotate` filter look like good, but custom `rotate` filter will be very useful, isn't it?

PR #588, opened 2015-04-07
Add image rotate filter

bocharsky-bw:
PR for #587 issue.

### DIFF
--- a/Imagine/Filter/Loader/RotateFilterLoader.php
+++ b/Imagine/Filter/Loader/RotateFilterLoader.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Liip\ImagineBundle\Imagine\Filter\Loader;
+
+use Imagine\Image\ImageInterface;
+use Imagine\Image\ManipulatorInterface;
+
+/**
+ * Loader for Imagine's basic rotate method
+ *
+ * @author Bocharsky Victor <bocharsky.bw@gmail.com>
+ */
+class RotateFilterLoader implements LoaderInterface
+{
+    /**
+     * Loads and applies a filter on the given image.
+     *
+     * @param ImageInterface $image
+     * @param array $options
+     *
+     * @return ManipulatorInterface
+     */
+    public function load(ImageInterface $image, array $options = array())
+    {
+        $angle = isset($options['angle']) ? (int) $options['angle'] : 0;
+
+        return 0 === $angle ? $image : $image->rotate($angle);
+    }
+}
--- a/Tests/Imagine/Filter/Loader/RotateFilterLoaderTest.php
+++ b/Tests/Imagine/Filter/Loader/RotateFilterLoaderTest.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Liip\ImagineBundle\Tests\Filter;
+
+use Liip\ImagineBundle\Imagine\Filter\Loader\RotateFilterLoader;
+use Liip\ImagineBundle\Tests\AbstractTest;
+
+/**
+ * Test cases for RotateFilterLoader class
+ *
+ * @covers Liip\ImagineBundle\Imagine\Filter\Loader\RotateFilterLoader
+ * @author Bocharsky Victor <bocharsky.bw@gmail.com>
+ */
+class RotateFilterLoaderTest extends AbstractTest
+{
+    public function testLoadRotate0Degrees()
+    {
+        $loader = new RotateFilterLoader();
+
+        $image = $this->getMockImage();
+
+        $result = $loader->load($image, array('angle' => 0));
+        $this->assertSame($image, $result);
+    }
+
+    public function testLoadRotate90Degrees()
+    {
+        $loader = new RotateFilterLoader();
+
+        $image = $this->getMockImage();
+        $rotatedImage = $this->getMockImage();
+
+        $image
+            ->expects($this->once())
+            ->method('rotate')
+            ->with(90)
+            ->willReturn($rotatedImage)
+        ;
+
+        $result = $loader->load($image, array('angle' => 90));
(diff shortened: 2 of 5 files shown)


bocharsky-bw:
If it's OK, I can add a few usage examples for it

makasim:
Looks good to me, could you please add some tests for it and fix currently broken.

would be good if you can add some docs too.

Koc:
Private method and variable looks useless

bocharsky-bw:
@Koc It unload `load` method and a bit improve readability and understanding in my opinion.
@makasim  What do you think, better to refactor code without `private` method and property in this case?

makasim:
There is not much need of private method. The code is pretty simple to understand in one place. :+1: Let's do as @Koc  suggested. Same for the property.

bocharsky-bw:
@makasim I added some tests and usage example to docs.

Imagine/Filter/Loader/RotateFilterLoader.php
```diff
@@ -0,0 +1,33 @@
...
+     * Loads and applies a filter on the given image.
+     *
+     * @param ImageInterface $image
+     * @param array $options
+     *
+     * @return ManipulatorInterface
+     */
+    public function load(ImageInterface $image, array $options = array())
+    {
+        $angle = isset($options['angle']) ? intval(round(floatval($options['angle']), 0)) : 0;
```
makasim: `intval(round(floatval($x)` is it really should be that complicated?

makasim: $angle = isset($options['angle']) ? (int) $options['angle'] : 0;

Tests/Imagine/Filter/Loader/RotateFilterLoaderTest.php
```diff
@@ -0,0 +1,33 @@
...
+class RotateFilterLoaderTest extends AbstractTest
+{
+    public function testLoad()
+    {
+        $loader = new RotateFilterLoader();
+
+        $image = $this->getMockImage();
+
+        $result = $loader->load($image, array('angle' => 0));
+        $this->assertTrue($image === $result);
```
makasim: $this->assertSame($image, $result);

Tests/Imagine/Filter/Loader/RotateFilterLoaderTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $image
+            ->expects($this->once())
+            ->method('rotate')
+            ->with(90)
+            ->willReturn($image)
+        ;
+
+        $result = $loader->load($image, array('angle' => 90));
+        $this->assertInstanceOf('Imagine\Image\ManipulatorInterface', $result);
+    }
```
makasim: you test two different cases in one test case, could you please split them into two? one for zero degree and one for 90 one.

makasim: it would also good to add a functional test for the filter service, like this one: https://github.com/liip/LiipImagineBundle/blob/master/Tests/Functional/Imagine/Filter/Loader/InterlaceFilterLoaderTest.php#L9

Imagine/Filter/Loader/RotateFilterLoader.php
```diff
@@ -0,0 +1,33 @@
...
+
+        if (0 === $angle) {
+            return $image;
+        }
+
+        return $image->rotate($angle);
```
makasim: return 0 === $angle ? $image : $image->rotate($angle);

Imagine/Filter/Loader/RotateFilterLoader.php
bocharsky-bw: OK, I want to maintain a good precision in this place, but most probably it's unnecessary because angle in 0.5 degree will not be visible to the naked eye )

makasim: I am thinking we should not do anything here. The angle comes from config? A developer set it? So it is ok if he wants 0.5 angle. Let's just pass it as is to rotate method...

wdyt?

bocharsky-bw: In my case, I use this filter at runtime in my twig templates, and my angle evaluate from several params and can be `float`. But `Imagine\Image\ManipulatorInterface::rotate` method expect angle as `int`, so better to type cast to `int`.

makasim: okay

Tests/Imagine/Filter/Loader/RotateFilterLoaderTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $image = $this->getMockImage();
+        $image
+            ->expects($this->once())
+            ->method('rotate')
+            ->with(90)
+            ->willReturn($image)
+        ;
+
+        $result = $loader->load($image, array('angle' => 90));
+        $this->assertInstanceOf('Imagine\Image\ManipulatorInterface', $result);
```
makasim: You have to create one more image mock and return it from rotate method. and here you have to check that it is that image, like you did in previous test 

```
$this->assertSame($rotatedImage, $result);
```

bocharsky-bw: OK, I did it

makasim: Do not check for instance of and keep only one load method with image and rotatedImage.

bocharsky-bw: Check again, please

makasim:
Thanks

PR #598, opened 2015-04-29
Add optipng post processor

franmomu:
This PR adds optipng as a post processor. The implementation is based on the existing [JpegOptimPostProcessor](https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php) and [OptiPngFilter from Assetic](https://github.com/kriswallsmith/assetic/blob/master/src/Assetic/Filter/OptiPngFilter.php).

The test are also based on [Assetic](https://github.com/kriswallsmith/assetic/tree/master/tests/Assetic/Test/Filter).

The Process component added to `composer.json` was needed in [JpegOptimPostProcessor](https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php).

I'm not sure what to do with `gif` files. It seems like if it's animated it doesn't work unless you use the `-snip` option, in which case it will cut one image out. I've seen some ways to [detect if a gif file is animated](http://it.php.net/manual/en/function.imagecreatefromgif.php#59787), but for me it's better to not support gif files and if someone wants this behaviour, they can implement it themselves.


Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -0,0 +1,113 @@
...
+use Liip\ImagineBundle\Model\Binary;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\ProcessBuilder;
+
+class OptiPngPostProcessor implements PostProcessorInterface
+{
+    const MIN_LEVEL = 0;
+    const MAX_LEVEL = 7;
+
+    private static $allowedTypes = array(
```
makasim: I think that this array could be used as default but there is a way to extend it, what about second constructor argument?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -0,0 +1,113 @@
...
+    public function setLevel($level)
+    {
+        $this->level = $this->sanitizeLevel($level);
+
+        return $this;
+    }
+
+    private function sanitizeLevel($level)
+    {
+        if ($level > self::MAX_LEVEL) {
```
makasim: there is no need for a separate method... could we move it to setter?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -0,0 +1,113 @@
...
+    public function process(BinaryInterface $binary)
+    {
+        $type = strtolower($binary->getMimeType());
+        if (!in_array($type, self::$allowedTypes)) {
+            return $binary;
+        }
+
+        $pb = new ProcessBuilder(array($this->optipngBin));
+
+        if ($this->level) {
```
makasim: what `level` property can take, is the `0` allowed value?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -0,0 +1,113 @@
...
+
+        if (0 !== $code) {
+            unlink($input);
+            throw new ProcessFailedException($proc);
+        }
+
+        $result = new Binary(file_get_contents($output), $binary->getMimeType(), $binary->getFormat());
+
+        unlink($input);
+        unlink($output);
```
makasim: this should be also done in `catch` block

Tests/Imagine/Filter/PostProcessor/FilterTest.php
```diff
@@ -0,0 +1,36 @@
...
+        $paths = array(__DIR__ . '/../../../../node_modules/.bin');
+        if ($current = ini_get('open_basedir')) {
+            ini_set('open_basedir', $this->ensurePaths($current, $paths));
+        } else {
+            $varname = getenv('PATH') ? 'PATH' : 'Path';
+            putenv(sprintf('%s=%s', $varname, $this->ensurePaths(getenv($varname), $paths)));
+        }
+        $finder = new ExecutableFinder();
+        return $finder->find($name);
+    }
```
makasim: looks overcomplicated

lsmith77:
ping .. needs a rebase .. where are we with the the review?

makasim:
closed in favor https://github.com/liip/LiipImagineBundle/pull/692

PR #599, opened 2015-04-30
Add filter to auto rotate images which uploads only from mobile devices

### LINKED ISSUE #532: Rotation issue
Hello.

I'm experiencing a weird bug.
I'm using the thumbnail filters and when my users upload a picture for example this one:
pic

Automatically is rotated... I've checked the exif data for this one and in orientation I found
Orientation Rotate 90 CW
So, I'm guessing that the bundle is taking the orientation on the exif data and applying automatically, but this behaviour can be deactivated? Because in this case this image is vertical but after uploading it, the thumbnail is displayed as landscape...

Thanks!

Edit --> After updating the picture here, is also rotated.

bocharsky-bw:
Possible resolve #532

### DIFF
--- a/Imagine/Filter/Loader/MobileAutoRotateFilterLoader.php
+++ b/Imagine/Filter/Loader/MobileAutoRotateFilterLoader.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Liip\ImagineBundle\Imagine\Filter\Loader;
+
+use Imagine\Image\ImageInterface;
+use Liip\ImagineBundle\Imagine\Filter\Loader\AutoRotateFilterLoader;
+
+/**
+ * Rotates Image based on its EXIF Data which loads only from mobile devices
+ *
+ * @author Victor Bocharsky <bocharsky.bw@gmail.com>
+ */
+class MobileAutoRotateFilterLoader extends AutoRotateFilterLoader
+{
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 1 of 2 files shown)


makasim:
not really sure, what problem it solves?

bocharsky-bw:
@makasim Hi! I think it resolve the problem, described in #532 issue.

The problem is that `auto_rotate` filter should rotate **only** images, which load from mobile devices directly. If user load image from mobile to his laptop first, and then load this image from laptop to the website, this image already rotated by laptop browser and shouldn't rotate by website.

It's probably needs more tests. I tested this image (http://server.ndv.net.ua/IMG_0556.JPG) on iPhone and Ubuntu laptop. Photo was made by iPhone. When I load image directly from iPhone - `auto_rotate` filter work correctly. But when I drop this image to laptop and then from laptop load it to website, image was rotated incorrectly (The correct result was achieved when I turn of `auto_rotate` filter when load image from laptop).

I think we need to modified `AutoRotateFilterLoader` to rotate only images which loads dicrectly from mobile devices. Or we can create `MobileAutoRotateFilterLoader` which would do this things.

I create `MobileAutoRotateFilterLoader` for other people can easily to test it. But Maybe better to modify `AutoRotateFilterLoader` to rotate only images which loads dicrectly from mobile devices. What do you think?

makasim:
Shouldn't ubuntu update exif info when doing rotate? This does not sound like the bundle's bug. And the implementation is way specific.

PR #566, opened 2015-02-28
use setFactory service definition method for symfony >= 2.6 (when possible)

adam187:
setFactoryClass, setFactoryMethod is deprecated since 2.6

PR #602, opened 2015-05-15
make configuration compatible with sf3.0

ip512:
As this will break compatibility with sf2.3, is it possible to put this change in a separate branch ? I'm not a version management expert, so if you have better suggestion, feel free to do something else

### DIFF
--- a/Resources/config/imagine.xml
+++ b/Resources/config/imagine.xml
@@ -245,19 +245,13 @@
 
         <!-- Guessers -->
 
-        <service
-            id="liip_imagine.mime_type_guesser"
-            class="Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesserInterface"
-            factory-class="Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesser"
-            factory-method="getInstance"
-        />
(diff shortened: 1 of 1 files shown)


makasim:
The current approach does not work in sf3.0, does it?

ip512:
Yes, the setFactoryClass method of Symfony\Component\DependencyInjection\Definition class is marked depracated and will be removed inf sf3.0.
With version 2.6+, there's some deprecated warnings only.

makasim:
so, maybe we can introduce a new imagine_sf30.xml file which is loaded only for sf3.0 version and overwrites services?  wdyt?

In this case it will work for both 2.x and 3.x version at the same time

ip512:
I looked how it's managed in other bundles, in fosuser there is a test in the DI configuration :
https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/DependencyInjection/FOSUserExtension.php#L46-51
I think it's better to avoid config duplication. 
In the meantime, I tried to launch the tests and I realized there's a lot of deprecated warnings and tests failures. Maybe this will require additional work to fix all that.

Koc:
similar to #566

lsmith77:
seems like #566 is indeed a more BC friendly approach

PR #605, opened 2015-05-28
Make AwsS3Resolver compatible with SDK v3

cdaguerre:
AwsS3Resolver is not compatible with AWS SDK v3 released earlier today because `Enum` constants were removed (refer: http://docs.aws.amazon.com/aws-sdk-php/v3/guide/guide/migration.html#enum-classes-have-been-removed)

This simple commit should make the resolver compatible with v3.

### DIFF
--- a/Imagine/Cache/Resolver/AwsS3Resolver.php
+++ b/Imagine/Cache/Resolver/AwsS3Resolver.php
@@ -2,7 +2,6 @@
 
 namespace Liip\ImagineBundle\Imagine\Cache\Resolver;
 
-use Aws\S3\Enum\CannedAcl;
 use Aws\S3\S3Client;
 use Liip\ImagineBundle\Binary\BinaryInterface;
 use Liip\ImagineBundle\Exception\Imagine\Cache\Resolver\NotStorableException;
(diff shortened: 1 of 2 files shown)


lsmith77:
looks good to me .. @makasim ?

lsmith77:
BTW I guess this line should also be updated then to include v3:
https://github.com/liip/LiipImagineBundle/blob/master/composer.json#L45

cdaguerre:
@lsmith77 `composer.json` updated.

lsmith77:
thx

makasim:
`require-dev` rely on 2.4 version of SDK, not sure we have to keep or update, any thoughts?

cdaguerre:
@makasim Actually I wasn't sure either.

I updated `aws/aws-sdk-php` to v3 locally and ran the unit test suite (excluding functional). I only got a bunch of symfony deprecation warnings due to the new version of the debug component, so I guess it works as expected and as it used to with v2.

According to the AWS migration guide:
*  Given the ***"data-driven and API version agnostic nature of Version 3"***, I guess this means from v3.0 on, the resolver should work with any new release and API version.
*  They also state they will follow strict SemVer standards from now on.

So my guess is: If it works with v2 and there should be no BC breaks going forward (at least in the v3.0 branch), we might aswell update the composer requirement.

PR #610, opened 2015-06-19
introduced DownscaleFilterLoader

sascha-meissner:
Added Feature 
DownscaleFilterLoader

basically it´s the inversed upscale filter

I missed that one,  as it is as useful as upscale filter,
so here you go


makasim:
Could you add some tests please?

lsmith77:
ping

PR #612, opened 2015-06-25
StreamLoader : try urlencode() if not found

bonswouar:
I had a problem fetching a file path, **with spaces and accents**, on S3 (`Source image ... not found.`), while this image URL *(the one thrown in the error message)* was working from my web browser.

Using `urlencode` on the path fixed the issue.

If you don't think it's a good idea please don't hesitate to leave a comment !

### DIFF
--- a/Binary/Loader/StreamLoader.php
+++ b/Binary/Loader/StreamLoader.php
@@ -52,7 +52,11 @@ class StreamLoader implements LoaderInterface
          * file_exists() is not used as not all wrappers support stat() to actually check for existing resources.
          */
         if (($this->context && !$resource = @fopen($name, 'r', null, $this->context)) || !$resource = @fopen($name, 'r')) {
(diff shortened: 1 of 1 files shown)


bonswouar:
What's happening with this TravisCI build by the way ?
```
 [Composer\Downloader\TransportException]
"https://getcomposer.org/composer.phar" appears broken, and returned an empty 200 response 
```

lsmith77:
question is why not do the url_encode by default then instead of retrying?

bonswouar:
@lsmith77 I just wasn't sure if there was no side effect

makasim:
I guess it is not needed any more, since we do decode in the controller: https://github.com/liip/LiipImagineBundle/blob/master/Controller/ImagineController.php#L80

Reopen in you think the issue is still there.

PR #613, opened 2015-06-30
Make it easier to get a dev environment up and running

lstrojny:
- Moves the ODM extension to suggest
- Ignores a few more directories
- Installs ODM for CI

### DIFF
--- a/Tests/Binary/Loader/GridFSLoaderTest.php
+++ b/Tests/Binary/Loader/GridFSLoaderTest.php
@@ -23,6 +23,9 @@ class GridFSLoaderTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
+        if (!extension_loaded('mongodb')) {
+            $this->markTestSkipped('ext/mongodb not installed');
+        }
         $this->repo = $this->getMockBuilder('Doctrine\ODM\MongoDB\DocumentRepository')->disableOriginalConstructor()->getMock();
 
         $dm = $this->getMockBuilder('Doctrine\ODM\MongoDB\DocumentManager')->disableOriginalConstructor()->getMock();


.travis.yml
```diff
@@ -16,6 +16,7 @@ before_script:
   - sh -c 'if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then echo "memory_limit = -1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi;'
   - composer require symfony/symfony:${SYMFONY_VERSION} --prefer-source
   - composer install --dev --prefer-source
+  - composer require doctrine/mongodb-odm ~1.0.3
```
lsmith77: why not put this into ``require-dev`` ?

.gitignore
```diff
@@ -4,3 +4,4 @@ composer.lock
 composer.phar
 vendor/
 Tests/Functional/app/web/media/cache
+/.idea/
```
lsmith77: this should be in your own gitignore file

.travis.yml
lsmith77: ah .. the point is not forcing mongodb odm for local dev?

lstrojny: Yes

.gitignore
lstrojny: Will do

lsmith77:
what is the problem with the ODM being installed by default? does it cause issues during composer install if you do not have the mongo ext installed? seems like in the tests everything is mocked anyway ..

makasim:
Personally I dont like this approach (having partial environments)

makasim:
The env has to be the same for all developers and CI.

lstrojny:
@lsmith77 you can't run composer install as ODM depends on ext-mongodb which might not be installed. For dev-dependencies I would try to not depend on dependencies, that are hard(er) to install, like PECL extensions. I am probably a good example for the issue, as everytime I want to do something on LiipImagineBundle I have to first edit the composer.json at least temporarily to get a dev env up an running. Not very optimal.

lsmith77:
I see your point .. or more importantly .. people tend not to be aware of ``--ignore-platform-reqs`` option in composer. but I do like the "skip test" but it should be in the extension

lstrojny:
@lsmith77 not sure I understand what you mean by "but I do like the 'skip test' but it should be in the extension"

Tests/Binary/Loader/GridFSLoaderTest.php
```diff
@@ -23,6 +23,10 @@ class GridFSLoaderTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
+        if (!class_exists('Doctrine\ODM\MongoDB\DocumentRepository')) {
+            $this->markTestSkipped('Doctrine ODM not installed');
```
lsmith77: in case our tests require the mongo pecl ext as well to run, we should also check for their existence to determine if to skip

lstrojny: Ah, I thought it would be sufficient to check for ODM, as that’s the one that requires ext/mongodb. My idea was to check for the highest abstraction and not all the way down. But I don’t mind adding it.

lstrojny:
OK, basically what’s left is the test for the extension. The rest has been reverted. Should be good to go now.

lsmith77:
thx!

PR #614, opened 2015-06-30
Use is_file() instead of Filesystem::exists()

lstrojny:
`Filesystem::exists()` will return true no matter what, even if the file is actually a directory. This happens e.g. if you pass an empty path to the Twig filter. I know this breaks the Filesystem encapsulation but there is no `is_file()` equivalent in Filesystem.

### DIFF
--- a/Imagine/Cache/Resolver/WebPathResolver.php
+++ b/Imagine/Cache/Resolver/WebPathResolver.php
@@ -63,7 +63,7 @@ class WebPathResolver implements ResolverInterface
      */
     public function isStored($path, $filter)
     {
-        return $this->filesystem->exists($this->getFilePath($path, $filter));
+        return is_file($this->getFilePath($path, $filter));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Imagine/Cache/Resolver/WebPathResolver.php
```diff
@@ -63,7 +63,7 @@ public function resolve($path, $filter)
      */
     public function isStored($path, $filter)
     {
-        return $this->filesystem->exists($this->getFilePath($path, $filter));
+        return is_file($this->getFilePath($path, $filter));
```
KalleZ: (Hopping in here from Twitter! Hi Lukas!)

Wouldn't you want to call it as \is_file() instead (same for the sys_get_temp_dir() call in WebPathResolverTest.php) so that the call shaves off the resolve time instead of looking for:

> \Liip\ImagineBundle\Imagine\Cache\Resolver\is_file()
> \Liip\ImagineBundle\Imagine\Cache\is_file()
> \Liip\ImagineBundle\Imagine\is_file()
> \Liip\ImagineBundle\is_file()
> \Liip\is_file()
> \is_file()

Although I can't remember if the lookup attempts between \Liip\ImagineBundle\Imagine\Cache\Resolver\is_file() and \is_file() are still the case.

lsmith77: I do not have a link handy but afaik this isn't relevant in terms of performance.

KalleZ: Indeed, although I would consider it good practice all-in-all (:

makasim:
:+1:

PR #615, opened 2015-06-30
Return string, not Twig_Markup object in Twig extension

lstrojny:
Returning a markup object breaks interoperability with the Symfony assets extension. More details here: https://github.com/symfony/symfony/pull/15158

### DIFF
--- a/Templating/ImagineExtension.php
+++ b/Templating/ImagineExtension.php
@@ -42,10 +42,7 @@ class ImagineExtension extends \Twig_Extension
      */
     public function filter($path, $filter, array $runtimeConfig = array())
     {
-        return new \Twig_Markup(
-            $this->cacheManager->getBrowserPath($path, $filter, $runtimeConfig),
-            'utf8'
-        );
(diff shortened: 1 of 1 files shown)


makasim:
This is a BC break IMO. I believe it was done (wrapping to Twig_Markup) on purpose to solve the problem we had, cannot recall which one though

lsmith77:
i guess https://github.com/liip/LiipImagineBundle/commit/54cfe94d0e31a6a66313cb5c8aa0fe0a82c906f0

lsmith77:
@stof / @tobion ?

makasim:
I recall, We had double url encoding and to avoid this Twig_Markup was added.

makasim:
I guess I found: https://github.com/liip/LiipImagineBundle/pull/314

lstrojny:
Wouldn’t it be better to mark the function as HTML safe then?

stof:
@lstrojny avoiding autoescaping looks really weird here, because the method **does not** return some HTML, and it is **not** safe for HTML (it returns an URL, which is not safe for HTML context as ``&`` needs to be escaped in it).

PR #619, opened 2015-07-17
Use identity instead equality

piotrantosik:


makasim:
why? array_key_exists always returns boolean

piotrantosik:
I know but is more readable for psr.
Basically this should be.

```php
if (!array_key_exists($filter, $this->filters)) {
    throw new NonExistingFilterException(sprintf('Could not find configuration  for a filter: %s', $filter));
}
```

Is acceptable?

trsteel88:
I prefer false ===

PR #620, opened 2015-07-19
Update AwsS3ResolverFactory.php

glenn--:
I got the AwsS3Resolver only working with such a configuration:

```
liip_imagine:
    resolvers:
       default_resolver:
          aws_s3:
              client_config:
                  credentials: { key: %aws_key%, secret: %aws_secret% }
                  region: %aws_s3_region%
                  version: %aws_s3_version%
```
aws-sdk-php (s3/v3) expects key and secret as array with credentials as key. So my proposal would be to allow client_config to be variable.

Best regards,
Glenn


lsmith77:
sorry for the late response .. we still should have some validation .. the Symfony config system does allow to embedd some validation here .. potentially the user will need to configure the version of the SDK explicitly to enable proper validation .. alternative we can try to detect which version of the SDK is installed .. you can look at the FOSRestBundle Configuration class to see some examples of using configuration validation callbacks.

makasim:
The config from the previous project I worked on. It works:

```
liip_imagine:
    loaders:
        stream_image:
            stream:
                wrapper: gaufrette://image/

    resolvers:
       web_path:
          web_path: ~

       aws_s3:
          aws_s3:
              client_config:
                  key:    %amazon.s3_key%
                  secret: %amazon.s3_secret%
                  region: %amazon.s3_region%
              bucket:     %amazon.s3_cache_bucket%
              cache:      liip_doctrine_cache.ns.imagine_cache
```

Why do you have to put `key`and `secret` inside `credentials`. Doesn't it work the way I posted it  above?

glenn--:
Sorry for the late answer. I didn't had the time to implement and better solution for handling the different configurations.

I've commited a new version of the resolver factory.

glenn--:
@makasim, this configuration only works with AWS SDK versions < 3

lsmith77:
what about https://github.com/liip/LiipImagineBundle/pull/636 ?

glenn--:
That isn't working for me.

In AwsS3ResolverFactory::create(), line 27, it seems that the client_config is set as argument as-is. So client_config has to be in the correct format in the config.yml. Credentials is needed as element and has to be an array. This leads to the structure:

```
              client_config:
                  credentials:
                    key: %aws_key%
                    secret: %aws_secret%
                  region: %aws_s3_region%
                  version: %aws_s3_version%
```

But that isn't working as described: "Invalid type for path "liip_imagine.resolvers.default_resolver.aws_s3.client_config.xyz". Expected scalar, but got array."

DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
```diff
@@ -84,17 +85,20 @@ public function getName()
      */
     public function addConfiguration(ArrayNodeDefinition $builder)
     {
+		if (class_exists('\Aws\Sdk')) {
```
dincho: this check looks a little bit weird, what about: 

```
        if (defined('\Aws\Sdk::VERSION') && version_compare(\Aws\Sdk::VERSION, '3.0.0', '>=')) {
            // v3
        } else {
            // v2
        }
```

glenn--:
Yes, you're right. This is a lot better. I've changed it.

lsmith77:
@glenn-- there are some CS issues

glenn--:
Sorry, I don't know how to fix that. What do I have to do?

lsmith77:
click on details on the line with "StyleCI" .. it will show you the issues .. it will even allow you to create a PR for this PR to fix all issues with a few clicks.

dincho:
> PR for this PR to fix all issues with a few clicks.

 great feature :lipstick:

glenn--:
Is this one OK now?

dincho:
It fails on PHP 5.3 because of the short array syntax. @lsmith77 if you support 5.3, I think the StyleCI config should be fixed to not force short array syntax.

makasim:
fixed in 1.4.1

PR #621, opened 2015-07-20
Updated docs for features introduced in Symfony 2.4

foaly-nr1:
Symfony 2.4 introduced constants for HTTP status codes to make code more readable. See: http://symfony.com/blog/new-in-symfony-2-4-using-constants-for-http-status-code.

Also introduced was the request stack with the request service being deprecated for removal in 3.0. See: http://symfony.com/blog/new-in-symfony-2-4-the-request-stack. As the filterAction() doesn't actually depend on the request, I suggest injecting an empty request object.

### DIFF
--- a/Resources/doc/filters.rst
+++ b/Resources/doc/filters.rst
@@ -263,7 +263,7 @@ A simple example showing how to change the filter configuration dynamically.
             $this->cacheManager->store($filteredBinary, $path, $filter);
         }
 
-        return new RedirectResponse($this->cacheManager->resolve($path, $filter), 301);
+        return new RedirectResponse($this->cacheManager->resolve($path, $filter), Response::HTTP_MOVED_PERMANENTLY);
     }
 
 Post-Processors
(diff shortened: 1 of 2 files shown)


makasim:
we still [support symfony 2.3](https://github.com/liip/LiipImagineBundle/blob/master/composer.json#L21). So there is no way to use new symfony features without providing BC.

foaly-nr1:
Ok. The second suggestion still makes sense though: As the filterAction() doesn't actually depend on the request, I suggest injecting an empty request object

Resources/doc/filters.rst
```diff
@@ -263,7 +263,7 @@ A simple example showing how to change the filter configuration dynamically.
             $this->cacheManager->store($filteredBinary, $path, $filter);
         }
 
-        return new RedirectResponse($this->cacheManager->resolve($path, $filter), 301);
+        return new RedirectResponse($this->cacheManager->resolve($path, $filter), Response::HTTP_MOVED_PERMANENTLY);
```
lsmith77: please add a ``note`` here to say that for Symfony 2.3 devs should use ``301`` instead of the constant.

lsmith77:
ping ..

foaly-nr1:
Done. Sorry for the delay

PR #631, opened 2015-08-25
Fix deprecated twig filter syntax

Rattler3:
With the release of Twig 1.21, deprecated calls in twig are being logged. Noticed this one.

### DIFF
--- a/Templating/ImagineExtension.php
+++ b/Templating/ImagineExtension.php
@@ -27,7 +27,7 @@ class ImagineExtension extends \Twig_Extension
     public function getFilters()
     {
         return array(
-            new \Twig_SimpleFunction('imagine_filter', array($this, 'filter')),
+            new \Twig_SimpleFilter('imagine_filter', array($this, 'filter'))
         );
     }
 
--- a/Templating/ImagineExtension.php
+++ b/Templating/ImagineExtension.php
@@ -27,7 +27,7 @@ class ImagineExtension extends \Twig_Extension
     public function getFilters()
     {
         return array(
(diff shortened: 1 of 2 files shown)


makasim:
It is not available since 1.0, right? So we have to raised a minimum twig version, Which one should it be?

Templating/ImagineExtension.php
```diff
@@ -27,7 +27,7 @@ public function __construct(CacheManager $cacheManager)
     public function getFilters()
     {
         return array(
-            'imagine_filter' => new \Twig_Filter_Method($this, 'filter'),
+            new \Twig_SimpleFunction('imagine_filter', array($this, 'filter')),
```
Koc: this is wrong. Should be `\Twig_SimpleFilter` instead

Koc:
@makasim it is available since `1.12.0` https://github.com/twigphp/Twig/commit/1918edefa937b81233087aa79297e54ea0462f9b

makasim:
@Koc thanks, @Rattler3  Could you update a twig section in composer json, saying that `1.12.0` is minimum version we require.

Rattler3:
@makasim I left twig requiring less than version 2 as this has not yet been released and could require more amends.

makasim:
@Rattler3 I meant the description of suggest section https://github.com/liip/LiipImagineBundle/pull/631/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L45

Rattler3:
@makasim I have added a note about the version needing to be greater than 1.12

makasim:
Thanks

PR #638, opened 2015-10-07
Implement Imagine Grayscale filter

gregumo:


Imagine/Filter/Loader/GrayscaleFilterLoader.php
```diff
@@ -0,0 +1,24 @@
...
+ * GrayscaleFilterLoader - apply grayscale filter
+ *
+ * @author Gregoire Humeau <gregoire.humeau@gmail.com>
+ */
+class GrayscaleFilterLoader implements LoaderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function load(ImageInterface $image, array $options = [])
```
makasim: PR looks good, though you use `[]` which is not supported on php5.3, but we still supports this version, could you changes this to be BC?

makasim: Tests failed on travis.

gregumo:
The fix is done but Travis is still failing. I think it fail for another reason.

lsmith77:
can you also add something to the docs?

lsmith77:
please rebase on master to get the test fixed

makasim:
@gregumo Could you please do a rebase?

gregumo:
@lsmith77 The doc for Grayscale Filter has been added

@makasim I just made the rebase but Imagine Grayscale Filter results are not the same between imagine/imagine 0.6.1 and 0.6.3. And the class Imagine\Image\Palette\RGB that I used doesn't exist before 0.6.1. So I propose to upgrade imagine dependency to 0.6.3 minimum. I made the change in the composer.json file but I let you add a new release tag for Liip if you think it's necessary.

Sorry for my late reply :/

makasim:
The bundle support imagine 0.5 now. Should we drop it? Is it widely used?

gregumo:
If you ask me if Grayscale filter is widely used, I really don't know but I think this is a really interesting filter to add to LiipImagineBundle.

For Imagine 0.5.0, my test can't work because the class 'Imagine\Image\Palette\RGB' that I load has been created in Imagine 0.6.0. So the continuous integration test with the composer tag --prefer-lowest failed if I don't change imagine dependency tags in composer.json.

Then I have a color issue between 0.6.1 and 0.6.3:
- for Imagine 0.6.0 and 0.6.1, the color returned after my Grayscale filter test is #145af0.
- for Imagine 0.6.3, the color returned is #565656.

This is why I propose to upgrade Imagine dependency to ^0.6.3

So I think we should drop Imagine 0.5 for 0.6.3 and upper.

makasim:
@lsmith77 @havvg wdyt? Can we drop support of 0.5.x imagine?

havvg:
I think so, just for semantic versioning: This is a BC and would increase the major version :)

lenybernard:
Hello,
:+1: 
When do you plan to merge this ?

gregumo:
Hi ! Do you plan to merge this PR ?

lsmith77:
I think we can bump the imagine dependency to 0.6.3 without a major new version

PR #639, opened 2015-10-12
Fixed wrong rotation for flipped images in auto_rotate filter

Heshyo:
The original code was not rotating (nor flipping) flipped images.

The code has been tested using those images https://github.com/recurser/exif-orientation-examples

### DIFF
--- a/Imagine/Filter/Loader/AutoRotateFilterLoader.php
+++ b/Imagine/Filter/Loader/AutoRotateFilterLoader.php
@@ -86,7 +86,7 @@ class AutoRotateFilterLoader implements LoaderInterface
             return isset($data['Orientation']) ? $data['Orientation'] : null;
         }
 
-        return null;
+        return;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lsmith77:
can you add a test case?

Heshyo:
I was able to setup my test environment but now I'm having issues faking the metadata object.
```php
$metaData = $this->getMock('Metadata\MetadataInterface');

$metaData
	->expects($this->once())
	->method('offsetGet')
	->willReturn("1")
;

$metaData->offsetGet();
// Call to undefined method Mock_MetadataInterface_7f3f507c::offsetGet()
```

Am I missing something obvious? I don't even know where ```MetadataInterface``` is defined (I see in ```ImageInterface``` that ```metadata``` is supposed to return a ```Metadata\MetadataInterface```).

lsmith77:
this interface is part of the "jms/metadata" composer package I guess .. but what makes you believe you need to mock this interface for a test?

Heshyo:
The autorotate needs to get the metadata from the image to know the rotation. So I want to mock it, to avoid having to provide real pictures.

I found the info on https://imagine.readthedocs.org/en/latest/usage/metadata.html. The ```metadata``` function returns a ```Imagine\Image\Metadata\MetadataBag```.

Heshyo:
I added the test cases and followed StyleCI recommendations.

lsmith77:
the style issues were already fixed in master .. it would be best if you start a new branch from current master and cherry pick your changes into there.

Heshyo:
Actually I only fixed the StyleCI issues for the files I modified.
I still created a new PR https://github.com/liip/LiipImagineBundle/pull/654

PR #654, opened 2015-10-29
- Task: correctly handles all rotations, even those involving flippin…

Heshyo:
Correctly handles all rotations, even those involving flipping horizontally.

### DIFF
--- a/Imagine/Filter/Loader/AutoRotateFilterLoader.php
+++ b/Imagine/Filter/Loader/AutoRotateFilterLoader.php
@@ -11,15 +11,15 @@ use Imagine\Image\ImageInterface;
  */
 class AutoRotateFilterLoader implements LoaderInterface
 {
-    protected $orientationKeys = [
+    protected $orientationKeys = array(
         'exif.Orientation',
         'ifd0.Orientation',
-    ];
+    );
 
     /**
      * {@inheritdoc}
      */
-    public function load(ImageInterface $image, array $options = [])
+    public function load(ImageInterface $image, array $options = array())
     {
         if ($orientation = $this->getOrientation($image)) {
             // Rotates if necessary.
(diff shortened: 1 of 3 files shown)


lsmith77:
please do not use short open tags .. I assume styleci initially told you that you need to use it before you rebased .. because the default config specifies short open tags, while the config in master defines that it should not be used.

Heshyo:
```AutoRotateFilterLoader.php``` and ```AutoRotateFilterLoaderTest.php``` both start with ```<?php```. I'm not sure where you're seeing ```<?```?

**Comment by lsmith77** [comment]
argh .. I meant short array syntax :)

File: Imagine/Filter/Loader/AutoRotateFilterLoader.php [hunk 1]
```diff
@@ -11,22 +11,27 @@
  */
 class AutoRotateFilterLoader implements LoaderInterface
 {
-    protected $orientationKeys = array(
+    protected $orientationKeys = [
```
lsmith77: please do not use short array syntax

Imagine/Filter/Loader/AutoRotateFilterLoader.php
```diff
@@ -11,22 +11,27 @@
...
         'exif.Orientation',
         'ifd0.Orientation',
-    );
+    ];
 
     /**
      * {@inheritdoc}
      */
-    public function load(ImageInterface $image, array $options = array())
+    public function load(ImageInterface $image, array $options = [])
```
lsmith77: please do not use short array syntax

Heshyo:
Yes I just noticed about the short array syntax. Should I do another commit or you already modified it?

lsmith77:
yeah please fix in another commit or by force pushing on the last commit

lsmith77:
thanks!

PR #655, opened 2015-10-30
Fix broken CacheResolver tests (#650)

### LINKED ISSUE #650: fix unit tests
```
There were 2 failures:
1) Liip\ImagineBundle\Tests\Imagine\Cache\Resolver\CacheResolverTest::testRemoveSinglePathCacheOnRemove
Failed asserting that actual size 2 matches expected size 3.
/home/travis/build/liip/LiipImagineBundle/Tests/Imagine/Cache/Resolver/CacheResolverTest.php:140
2) Liip\ImagineBundle\Tests\Imagine\Cache\Resolver\CacheResolverTest::testRemoveAllFilterCacheOnRemove
Failed asserting that actual size 6 matches expected size 7.
/home/travis/build/liip/LiipImagineBundle/Tests/Imagine/Cache/Resolver/CacheResolverTest.php:175
```

This seems to be a regression of some sort since things are working fine when we test lower dependencies: https://travis-ci.org/liip/liip/LiipImagineBundle/builds/87425790

kamazee:
Fixes #650 

[Since September'2015 Doctrine doesn't always save namespace version in
the cache itself] (https://github.com/doctrine/cache/commit/ecc4af15dc8d198ab6fc7b6ccecba1728d47b0da)

Relying on such kind of implementation detail in a test that is supposed to check if several keys are saved in cache is kind of pointless, so it makes sense to ignore the cache key that holds a namespace version -- that's exactly what's implemented in the PR.

### DIFF
--- a/Tests/Imagine/Cache/Resolver/CacheResolverTest.php
+++ b/Tests/Imagine/Cache/Resolver/CacheResolverTest.php
@@ -132,17 +132,16 @@ class CacheResolverTest extends AbstractTest
         $cacheResolver->resolve($this->path, $this->filter);
 
         /*
-         * Three items:
+         * Checking 2 items:
          * * The result of one resolve execution.
          * * The index of entity.
-         * * The array cache meta info
          */
-        $this->assertCount(3, $this->readAttribute($cache, 'data'));
+        $this->assertCount(2, $this->getCacheEntries($cache));
 
         $cacheResolver->remove(array($this->path), array($this->filter));
 
         // Cache including index has been removed.
-        $this->assertCount(1, $this->readAttribute($cache, 'data'));
+        $this->assertCount(0, $this->getCacheEntries($cache));
     }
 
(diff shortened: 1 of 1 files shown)


lsmith77:
can you look at the prefer-lowest test too? https://travis-ci.org/liip/LiipImagineBundle/jobs/88415483 
we might just have to bump up the lower dependencies

kamazee:
@lsmith77 yeah, I noticed that. That test fails with Imagine < 0.6, but it's not the code, it's the test that tests the code for old versions of Imagine improperly. I fixed the test and one ofther issue with `AutoRotateFilterLoader` (it didn't handle exceptional situations properly), see #656

kamazee:
So, when this PR and #657 are merged together, the build should get green. I'll probably put them together and create a dummy PR with both commits just to make Travis build them both and check my assumption.

PR #657, opened 2015-10-31
Fix handling invalid orientation in AutoRotateFilterLoader & test exceptions

### LINKED ISSUE #656: Fix AutoRotateFilterLoader and tests for it
There are 2 issues with `AutoRotateFilterLoader`:

* It fails miserably when an image with an invalid orientation tag is passed (tries to throw `Exception` that doesn't exist in the current namespace (it's not tested with any existing test)

* Tests fail when it falls back to `exif_read_data` due to outdated Imagine (even though class itself handles this case correctly, tests don't account for this case at all, hence [this build fail](https://travis-ci.org/liip/LiipImagineBundle/jobs/88086170))

kamazee:
Fixes #656


Tests/Imagine/Filter/Loader/AutoRotateFilterLoaderTest.php
```diff
@@ -26,30 +27,42 @@ private function loadExif($exifValue, $expectCallRotateValue, $expectCallFlip)
...
-        // If $exifValue is null, it means the image doesn't contain any metadata.
-        $metaData = $this->getMockMetaData();
+        // Mocks the image and makes it use the fake meta data.
+        $image = $this->getMockImage();
 
-        $metaData
-            ->expects($this->atLeastOnce())
-            ->method('offsetGet')
-            ->willReturn($exifValue);
+        if (method_exists(ImageInterface::class, 'metadata')) {
```
lsmith77: we still need to be 5.3 friendly

kamazee: OOPS. Will fix.

PR #658, opened 2015-10-31
Fix all test fails in master (just to check)

kamazee:
Just a dummy pull request to check #655 and #657 together against travis


lsmith77:
awesome!

Issue #633, opened 2015-09-04
Memory leak inside the FilterManager


cmodijk:
I'm currently debugging a memory leak we have inside of our application. We have a long running php script which generates video's from a serie of images and we are using the LiipImageBundle to convert al the images to the correct sizes before we start generating the video using ffmpeg. We use Imagick to covert the images. After looking into this problem I found a [bug report](https://bugs.php.net/bug.php?id=59878) at PHP and I upgraded to the latest version's of Imagick (3.3.0RC2) and ImageMagick (ImageMagick 6.9.2-0) but the problem still remains. 

After a debugging session this evening i found that the https://github.com/avalanche123/Imagine/ library has implemented a [```__destruct```](https://github.com/avalanche123/Imagine/blob/develop/lib/Imagine/Imagick/Image.php#L82-L88) method but this method is only getting called after the script ended. But in our long running situation this never happens. 

I tried adding a ```unset($image)``` inside the [apply](https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/FilterManager.php#L92) method but that does not help destructing the object and destructing imagick to free the memory used. I just added a call to ```$image->__destruct();``` and that does the trick to free memory. 

I'm not sure if this is something to be implemented inside this bundle or if we should implement it ourself with a custom FilterManager.

lsmith77:
related to this see http://paul-m-jones.com/archives/262

cmodijk:
@lsmith77 Tnx great artikel! We currently are not using LiipImagineBundle any more and implement our own filters using https://github.com/avalanche123/Imagine with the custom `__destruct` methods. I'm closing this because it's not an issue for us anymore.

PR #682, opened 2015-12-21
Destruct image to cleanup memory

cmodijk:
In #633 we where encountered a lot of memory leak using the imagick extension. Imagick and Imagine fix that by using the `__destruct` method but because we use this in a long running process the `__destruct` is never getting called.  This pull requests fixes that issues.

### DIFF
--- a/Imagine/Filter/FilterManager.php
+++ b/Imagine/Filter/FilterManager.php
@@ -109,7 +109,14 @@ class FilterManager
                 ));
             }
 
+            $prevImage = $image;
             $image = $this->loaders[$eachFilter]->load($image, $eachOptions);
+
+            // If the filter returns a different image object destruct the old one because imagick keeps consuming memory if we don't
+            // See https://github.com/liip/LiipImagineBundle/pull/682
+            if ($prevImage !== $image && method_exists($prevImage, '__destruct')) {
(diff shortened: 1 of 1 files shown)


makasim:
I dont like it, kind of hack. We can add an event dispatch there, like POST_FILTER or so and you can add a listener with custom logic in your app.

cmodijk:
@makasim I don't like it to but it don't know if an event is going to help because  in the event you don't know of the imagick object is different than the previous one. And if you `__destruct ` the images that is still being used you get other problems and errors from that.

makasim:
Okay, can you add a comment there (with a link to the issue) cuz it is not obvious from the code what it solves.

makasim:
@cmodijk ping

cmodijk:
@makasim Sorry! To what issue do you want to point the link? This pull request or the original problem?

cmodijk:
@makasim I added a comment but I'm not so good with words in English could you please check the comments?

mattjanssen:
Thank you! I spent hours yesterday debugging this typical-PHP memory leak. I was about to submit a PR, but I see you took care of it. :+1:

I also used the __destruct() method, even though it feels gross; I could think of no better way. The comments are valuable.

PR #683, opened 2015-12-23
Fixed config issue with AWS SDK v3 in AwsS3ResolverFactory

kieste:
I just modified the pull request from glenn-- so that his changes pass PHP 5.3 checks.

### DIFF
--- a/LiipImagineBundle.php
+++ b/LiipImagineBundle.php
@@ -9,6 +9,7 @@ use Liip\ImagineBundle\DependencyInjection\Compiler\ResolversCompilerPass;
 use Liip\ImagineBundle\DependencyInjection\Factory\Loader\FileSystemLoaderFactory;
 use Liip\ImagineBundle\DependencyInjection\Factory\Loader\StreamLoaderFactory;
 use Liip\ImagineBundle\DependencyInjection\Factory\Resolver\AwsS3ResolverFactory;
+use Liip\ImagineBundle\DependencyInjection\Factory\Resolver\AwsS3SdkV3ResolverFactory;
 use Liip\ImagineBundle\DependencyInjection\Factory\Resolver\WebPathResolverFactory;
 use Liip\ImagineBundle\DependencyInjection\LiipImagineExtension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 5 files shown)


makasim:
What if we add a new factory for v3 sdk? Same time we can deprecate curent aws factory. 

It would be good if you can add some docs on when to use what factory.

Sounds good?

makasim:
@kieste are you planning to finish this PR?

kieste:
Yes I will fix this before end of the year.

kieste:
@makasim separated old and new SDK client config code into different factory classes. Added a note to the documentation.

LiipImagineBundle.php
```diff
@@ -32,7 +33,11 @@ public function build(ContainerBuilder $container)
         $extension = $container->getExtension('liip_imagine');
 
         $extension->addResolverFactory(new WebPathResolverFactory());
-        $extension->addResolverFactory(new AwsS3ResolverFactory());
+        if (defined('\Aws\Sdk::VERSION') && version_compare(\Aws\Sdk::VERSION, '3.0.0', '>=')) {
```
makasim: register both, and let a developer decided which one to use. Also would be good to add an exception if a resolver is used with wrong sdk.

Resources/doc/cache-resolver/aws_s3.rst
```diff
@@ -42,6 +42,18 @@ Create resolver using factory
                   put_options:
                       CacheControl: 'max-age=86400'
 
+If you use `aws-sdk-php`_ library version >= 3.0.0 client config credentials
+must be an associative array containing key and secret.
+
+.. code-block:: yaml
+
+    aws_s3:
```
makasim: I meant the resolver for sdk v3 has to have separate name, like aws_s3_sdkv3

kieste:
@makasim after taking a closer look at my current setup I realized that I am not using the factories any more. (Currently I am creating the AwsS3Resolver as a service and injecting it directly into a proxy resolver.)

So I can't finish this pull request now with your suggested changes. (I must progress with my current project.) But I aggree to all of your comments. This totally makes sense. But it's too much effort for me at the moment.

makasim:
Thanks for your contribution, I took care of your PR. Soon in the master

PR #692, opened 2016-01-08
Adding optipng post transformer

gouaille:


makasim:
tests fails, could @gouaille please fix them?

https://travis-ci.org/liip/LiipImagineBundle/jobs/101123613

gouaille:
Edited.

StyleCI seems to be ok if we check the details ?

makasim:
Have you tested it in real world? Does it work for you?

gouaille:
Yes, but my project is private atm and i can't show you an example

makasim:
I do not need an example, just need confirmation that the solution was tested in real world.

makasim:
@gouaille Could you please add some words to the documentation about the new post proceesor

https://github.com/liip/LiipImagineBundle/blob/master/Resources/doc/filters.rst#post-processors

gouaille:
It is working in real world.

Doc updated.

makasim:
Thanks

PR #693, opened 2016-01-14
Update WebPathResolver.php

gonzalovilaseca:
Missing empty line


makasim:
StylieCI does not like the change: https://styleci.io/analyses/8QbPgX

gonzalovilaseca:
Something is wrong with StyleCI then.

makasim:
I cannot merge it then

javiereguiluz:
@gonzalovilaseca the problem is that your new line is not empty but it contains three white spaces. If you remove those white spaces, everything should be fine.

Issue #640, opened 2015-10-12
Background filter broken


peterpoe:
Error:  Attempted to call method "palette" on class "Imagine\Imagick\Image" in vendor/liip/imagine-bundle/Liip/ImagineBundle/Imagine/Filter/Loader/BackgroundFilterLoader.php at line 22

```php
$background = $image->palette()->color(
    isset($options['color']) ? $options['color'] : '#fff',
    isset($options['transparency']) ? $options['transparency'] : null
);
```
should be
```php
$background = new Color(
    isset($options['color']) ? $options['color'] : '#fff',
    isset($options['transparency']) ? $options['transparency'] : 0
);
````

PR #699, opened 2016-01-22
Fix background filter

jockri:
This solves #640

### DIFF
--- a/Imagine/Filter/Loader/BackgroundFilterLoader.php
+++ b/Imagine/Filter/Loader/BackgroundFilterLoader.php
@@ -3,6 +3,7 @@
 namespace Liip\ImagineBundle\Imagine\Filter\Loader;
 
 use Imagine\Image\Box;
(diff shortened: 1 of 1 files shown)


makasim:
Does it work for you?

jockri:
Yes, before this change I was receiving this error: 
`Error: Attempted to call method "palette" on class "Imagine\Imagick\Image" in vendor/liip/imagine-bundle/Liip/ImagineBundle/Imagine/Filter/Loader/BackgroundFilterLoader.php at line 22`
And now it works again.

makasim:
What is the version of imagine library you have?

jockri:
```
        "name": "imagine/imagine",
            "version": "v0.5.0",
            "source": {
                "type": "git",
                "url": "https://github.com/avalanche123/Imagine.git",
                "reference": "f64ec666baaa800edcbf237db41121a569230709"
            },
```

PR #705, opened 2016-02-04
Add FileBinary[Interface] to support large files without loading them in memory unnecessarily

Seldaek:
This works in combination with https://github.com/avalanche123/Imagine/pull/479 to allow me to generate thumbnails for very large images (>1GB) with low memory usage (tested with 50MB) vs needing to increase the memory usage to a few gigs.

It should be BC as far as I can tell.

### DIFF
--- a/Binary/FileBinaryInterface.php
+++ b/Binary/FileBinaryInterface.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Liip\ImagineBundle\Binary;
+
+interface FileBinaryInterface extends BinaryInterface
+{
+    /**
+     * @return string
+     */
+    public function getPath();
+}
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -2,7 +2,7 @@
 
 namespace Liip\ImagineBundle\Binary\Loader;
 
-use Liip\ImagineBundle\Model\Binary;
+use Liip\ImagineBundle\Model\FileBinary;
 use Symfony\Component\HttpFoundation\File\MimeType\ExtensionGuesserInterface;
 use Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesserInterface;
 use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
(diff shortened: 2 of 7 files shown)


makasim:
I understand the problem, but solution looks a bit hacky. I have to think of a better approach. If nothing comes in my mind I merge this one.

@Seldaek  Maybe you already have some ideas?

Seldaek:
I don't really see how it is hacky.. sometimes we have files, sometimes we just have data blobs, and converting it all to data blobs is extremely inefficient if a path to a file does the job. The only other way would be to always convert to file and write data blobs in a temp dir, but that's also less efficient and tends to cause new types of issues with permissions and whatnot. The current patch takes care of both of these in the best possible way as far as I can tell.

lsmith77:
@makasim can you elaborate about your concerns?

makasim:
First of all I am a big fan of things like this but since we have a BC promise we have to develop a good solution and here's what do not like in current approach: 

* It adds if else logic to every place where we have to create image from binary object
* It works only for filesystem loader. 
* What about other loaders? 
* Rename `FileBinary` to `BinarySteram`

Seldaek:
- The if else logic I agree isn't great but I don't see a way around it, it's not even a repeating one that we could easily abstract.
- I don't see how it can apply to other loaders.. if your files aren't stored locally then you will need to read them into memory somehow before processing them, or we could migrate them to a temp dir and create a FileBinary with that.
- The name issue if there is a consensus I don't mind renaming it, it's a detail. I don't really see this as a stream though because as mentioned above it can't be guaranteed to work with any php stream. For example if you have a flysystem or s3 stream wrapper and pass that as a file to Imagine, I have no idea what will happen.

makasim:
https://github.com/liip/LiipImagineBundle/tree/1.5.0

Seldaek:
Ok then ;) Thanks

makasim:
Thank you too

MAXakaWIZARD:
@Seldaek Last stable release of https://github.com/avalanche123/Imagine was on 19 Sep 2015. As I got it, we need new stable release of imagine to be tagged or install it as `dev-master`.

MAXakaWIZARD:
However, even after plugging optimized version, I am still not able to resize 5Mb image and getting:
`Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 71663617 bytes) in /home/proj/vendor/imagine/imagine/lib/Imagine/Gd/Image.php on line 617`

Seldaek:
Yeah? Maybe something else broke it then, or maybe that doesn't work with GD. I used imagick.

But yes there needs a new tag either way. I'm running on my forks for now as I don't really trust people here since there is no regression test in place I am afraid it'll break in the future. Writing tests would be good but no time.

MAXakaWIZARD:
@Seldaek I've used GD, maybe it is the culprit.

PR #707, opened 2016-02-13
Fix regression introduced in bb8e4109672902e37931e0a491ff55ebac93d8e9

Seldaek:
Sorry about that, missed it when doing my patch.

Fixes https://github.com/liip/LiipImagineBundle/commit/bb8e4109672902e37931e0a491ff55ebac93d8e9#commitcomment-16066858


xavierlacot:
Another solution, which keeps using FileBinary: https://github.com/xavierlacot/LiipImagineBundle/commit/b763eaa3b780574808c547e916774e9ebfc0f9e1

Seldaek:
Yeah that seems better, I am not familiar enough with the bundle so wasn't sure how these classes were being used.

PR #717, opened 2016-03-20
Introduce mozjpeg and pngquant post-processors, add transform options.

antoligy:
Implementing mozjpeg and pngquant post processors, as both allow for significantly greater _lossy_ compression to the current options.  This also allows for options to be passed at run-time, as different image filter sets may require different degrees of compression.

Concerning mozjpeg in particular I'd like to revisit this later on, as the current implementation can be improved to account for the level of detail required in an image (and dpr), but the current default is fairly safe and I'm already seeing reductions of 80-90% compared to the raw images.

### DIFF
--- a/Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Liip\ImagineBundle\Imagine\Filter\PostProcessor;
+
+use Liip\ImagineBundle\Binary\BinaryInterface;
+use Liip\ImagineBundle\Binary\FileBinaryInterface;
+use Liip\ImagineBundle\Model\Binary;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\ProcessBuilder;
+
+/**
+ * mozjpeg post-processor, for noticably better jpeg compression
+ * 
+ * @see http://calendar.perfplanet.com/2014/mozjpeg-3-0/
+ * @see https://mozjpeg.codelove.de/binaries.html
+ *
+ * @author Alex Wilson <a@ax.gy>
+ */
+class MozJpegPostProcessor implements PostProcessorInterface
+{
+    /** @var string    Path to the mozjpeg cjpeg binary */
+    protected $mozjpegBin;
+
+    /** @var null|int  Quality factor */
+    protected $quality;
+
+    /**
+     * Constructor.
+     *
+     * @param string   $mozjpegBin Path to the mozjpeg cjpeg binary
+     * @param int|null $quality    Quality factor
+     */
+    public function __construct(
+        $mozjpegBin = '/opt/mozjpeg/bin/cjpeg',
+        $quality = null
+    ) {
+        $this->mozjpegBin = $mozjpegBin;
+        $this->setQuality($quality);
+    }
+
+    /**
+     * @param int $quality
+     *
+     * @return MozJpegPostProcessor
+     */
+    public function setQuality($quality)
+    {
+        $this->quality = $quality;
+
+        return $this;
+    }
+
+    /**
+     * @param BinaryInterface $binary
+     *
+     * @param array           $options
+     *
+     * @throws ProcessFailedException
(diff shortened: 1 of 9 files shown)


Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
```diff
@@ -0,0 +1,95 @@
...
+    protected $quality;
+
+    /**
+     * Constructor.
+     *
+     * @param string   $mozjpegBin Path to the mozjpeg cjpeg binary
+     * @param int|null $quality    Quality factor
+     */
+    public function __construct(
+        $mozjpegBin = '/opt/mozjpeg/bin/cjpeg',
```
crevillo: could be this made configurable somehow?

antoligy: This is already configurable here as a parameter - the value in the constructor shouldn't ever need to be used. https://github.com/antoligy/LiipImagineBundle/blob/implement-pngquant-mozjpeg/Resources/config/imagine.xml#L79

antoligy:
Hey @lsmith77, any updates on this?

antoligy:
Hiya @lsmith77, any updates with this PR?  I'd love to start rolling this out to production as it really has been helping cut back on page weight.  Please let me know! 😄

machoo:
Hey @lsmith77. Are you actually going to respond to this PR anytime soon? It's been over a month.. Would love to see this merged sometime soon so i can take advantage of better compression 👯

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -25,14 +25,15 @@ public function __construct($optipngBin = '/usr/bin/optipng')
...
+     * @param array           $options
      *
      * @throws ProcessFailedException
      *
      * @return BinaryInterface
      *
      * @see      Implementation taken from Assetic\Filter\optipngFilter
      */
-    public function process(BinaryInterface $binary)
+    public function process(BinaryInterface $binary, array $options)
```
makasim: this is BC break, not sure we can do it,

makasim: do you have any idea BC way?

makasim: I am the second argument was added to the interface

makasim:
@antoligy Hi, I am fine to merge this if you can do it BC way.

antoligy:
@makasim Cheers for reviewing!  Just looking at this, instead of making the change to `process`, does it make more sense to introduce a new `setOptions` setter?  IMO the processors should remain stateless, which I think is why I originally made the change to `process`...
Can you think of any (better) cleaner way of allowing per-operation configuration whilst maintaining the stateless nature of the PostProcessors?

antoligy:
One alternative might be to introduce a new `ConfigurablePostProcessor` interface enforcing a new method - `processWithConfiguration(BinaryInterface $binary, array $options)`, which could be callable from the older `process`...

That said I'm not so sure that this does break BC - Looking at the iterator in FilterManager, an array is always passed to the processors and I can't see anywhere else from where these are called.  I suppose this affects custom post processors, however if this change is documented somewhere or there's a new minor revision then to me this wouldn't be so much of an issue.

What do you think

makasim:
Let's create a new interface with a note "This for BC and so on" so while working on the 2.x version we can merge them into one again.

antoligy:
How's this? (Separate interface, versus an extended one)

StyleCI has failed on the Test bootstrap.php.

makasim:
I'll do a deeper review on wednesday when I finally come back to work.

antoligy:
Hey @makasim, have you had a chance to look at this yet? 😄

makasim:
https://github.com/liip/LiipImagineBundle/releases/tag/1.5.3

@antoligy  Thanks for the contribution

ceesvanegmond:
Is there an option to always add (the same) post processor to every filter_set?
Now, in every filter_set I have to define the same post processor.

robfrawley:
@ceesvanegmond There is not.

PR #721, opened 2016-03-30
background filter: allow image positioning

uvoelkel:
allow positioning of the image on the background. uses the same position options as the watermark filter.


uvoelkel:
Scrutinizer - rightfully - complains about code duplication.
Should I move the position calculation into its own class? If so, where to put it and what to name it.

PR #751, opened 2016-07-04
Ignore invalid exif orientations

lstrojny:
While exif orientation values should theoretically only be between 1 and 8, the reality is different. We’ve seen all kinds of values from 0 to 256 to 65535.

### DIFF
--- a/Imagine/Filter/Loader/AutoRotateFilterLoader.php
+++ b/Imagine/Filter/Loader/AutoRotateFilterLoader.php
@@ -24,9 +24,7 @@ class AutoRotateFilterLoader implements LoaderInterface
     {
         if ($orientation = $this->getOrientation($image)) {
             if ($orientation < 1 || $orientation > 8) {
-                throw new InvalidArgumentException(
-                    sprintf('The image has wrong EXIF orientation tag (%d)', $orientation)
-                );
+                return $image;
             }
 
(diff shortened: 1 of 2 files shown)


lsmith77:
@lstrojny can you look into the CS issues?

furthermore .. do you have any link that would support the claim that other values are possible "in the wild"?

lstrojny:
@lsmith77 the CS issues aren’t introduced with this PR but have been there before. Should we fix them with this PR?
On sources: quick googling revealed this issue by the Wikimedia project (https://phabricator.wikimedia.org/T34868) which isn't about out of range exif tags but duplicate exif tags.

lsmith77:
regarding the CS issues: doh .. fixed via https://github.com/liip/LiipImagineBundle/pull/757

as for the validation/exception .. maybe the sensible thing to do is make it a config option?

lstrojny:
@lsmith77 I am against making it a configuration option as I don't see the use case for it. Why would anyone want an image to not be rendered at all instead of the current behavior which strips the (invalid) exif information and presents the image as-is.

lsmith77:
not sure .. but I assume if you use the auto rotate filter, you want to know the image is rotated accordingly and if an image cannot be rotated then you have a problem. then again I guess this should be validated before it even gets to this Bundle.

PR #753, opened 2016-07-08
clean filename for WebPathResolver

DenisMedved:
This PR help us make filename more friendly for file system. It is can be helpful when we user url with some parameters and path.

Example of url:
```
    http://gravatar.com/avatar/e4fb8bac7961505d218aba493994e1d8?s=80&r=g&d=mm
```

It will be converted into `avatar/e4fb8bac7961505d218aba493994e1d8` and that name can be successfully saved.

Example of config:
```
liip_imagine:
    loaders:
        ...
        stream.url:
            stream:
                wrapper: http://www.gravatar.com/

    ...
    filter_sets:
        ...
        cache_url_medium:
            data_loader: stream.url
            quality: 70
            filters:
                thumbnail: { size: [77, 81], mode: outbound }

```

### DIFF
--- a/Imagine/Cache/Resolver/WebPathResolver.php
+++ b/Imagine/Cache/Resolver/WebPathResolver.php
@@ -69,7 +69,11 @@ class WebPathResolver implements ResolverInterface
      */
     public function isStored($path, $filter)
     {
-        return is_file($this->getFilePath($path, $filter));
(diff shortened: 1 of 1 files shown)


DenisMedved:
I have found some mistakes, so decide close this PR

PR #754, opened 2016-07-11
Remove url part with parameters from filename.

DenisMedved:
This PR help us make filename more friendly for file system. It can be helpful when we use url with some parameters and path.

Example of url:
```
    http://gravatar.com/avatar/e4fb8bac7961505d218aba493994e1d8?s=80&r=g&d=mm
```

It will be converted into `avatar/e4fb8bac7961505d218aba493994e1d8` and that name can be successfully saved.

Example of config:
```
liip_imagine:
    loaders:
        ...
        stream.url:
            stream:
                wrapper: http://www.gravatar.com/

    ...
    filter_sets:
        ...
        cache_url_medium:
            data_loader: stream.url
            quality: 70
            filters:
                thumbnail: { size: [77, 81], mode: outbound }

```

This PR  was recreated form PR #753

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -100,7 +100,9 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('driver')->defaultValue('gd')
                     ->validate()
-                        ->ifTrue(function ($v) { return !in_array($v, array('gd', 'imagick', 'gmagick')); })
+                        ->ifTrue(function ($v) {
+                            return !in_array($v, array('gd', 'imagick', 'gmagick'));
+                        })
                         ->thenInvalid('Invalid imagine driver specified: %s')
                     ->end()
                 ->end()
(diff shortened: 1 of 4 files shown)


trsteel88:
This will fail as soon as a second request is made with a different parameter. E.g what happens when s=80 changes to s=200? The filename would conflict.

DenisMedved:
@trsteel88 , can we store filename as sha1 from path ? 


```
    public function cleanPath($path)
    {
        $dir = pathinfo($path, PATHINFO_DIRNAME);
        $filename = pathinfo($path, PATHINFO_FILENAME);

        return $dir.DIRECTORY_SEPARATOR.sha1($filename);
    }
```

trsteel88:
I don't think this logic belongs here. If you need to fetch an image from a remote location, you should be writing a loader for it?

Then doing something like {{ 'http://gravatar.com/avatar/e4fb8bac7961505d218aba493994e1d8?s=80&r=g&d=mm'|imagine_filter('thefilter') }}

DenisMedved:
Ok, I think you are right. Thanks you for response.

PR #720, opened 2016-03-24
Enable configuration of post processors using parameters

notrix:

PR #759, opened 2016-07-21
Enable configuration of post processors using parameters (closes #720)

antoligy:
Closes #720

### DIFF
--- a/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
@@ -38,8 +38,11 @@ class JpegOptimPostProcessor implements PostProcessorInterface, ConfigurablePost
      * Constructor.
      *
      * @param string $jpegoptimBin Path to the jpegoptim binary
+     * @param bool   $stripAll     Strip all markers from output
+     * @param int    $max          Set maximum image quality factor
+     * @param bool   $progressive  Force output to be progressive
      */
-    public function __construct($jpegoptimBin = '/usr/bin/jpegoptim', $stripAll = true, $max = 100, $progressive = true)
+    public function __construct($jpegoptimBin = '/usr/bin/jpegoptim', $stripAll = true, $max = null, $progressive = true)
     {
         $this->jpegoptimBin = $jpegoptimBin;
         $this->stripAll = $stripAll;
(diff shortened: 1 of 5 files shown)


lsmith77:
please rebase

antoligy:
Done, tests re-running now

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
```diff
@@ -38,8 +38,11 @@ class JpegOptimPostProcessor implements PostProcessorInterface, ConfigurablePost
      * Constructor.
      *
      * @param string $jpegoptimBin Path to the jpegoptim binary
+     * @param bool   $stripAll     Strip all markers from output
+     * @param int    $max          Set maximum image quality factor
+     * @param bool   $progressive  Force output to be progressive
      */
-    public function __construct($jpegoptimBin = '/usr/bin/jpegoptim', $stripAll = true, $max = 100, $progressive = true)
+    public function __construct($jpegoptimBin = '/usr/bin/jpegoptim', $stripAll = true, $max = null, $progressive = true)
```
lsmith77: is this change intentional?

antoligy: I was just looking at this myself during the rebase - it's originally from here https://github.com/liip/LiipImagineBundle/pull/720/commits/81ee48a241259660d0288343721e41da31ac6ffa#diff-9cd4d6a634315c7cf9b3483cb8b37db8R45 so I chose to keep it.

Setting --max [disables lossless mode](https://www.mankier.com/1/jpegoptim#--max) and will reduce the quality factor of jpegs (if greater than its current value) to the current setting - My assumption then is that it doesn't matter, as 100 is the max quality factor for the jpeg format anyway, so removing the flag shouldn't matter much.

I can run a few test cases using jpegoptim by hand to see if there's any difference, if that'd be preferred?

lsmith77: yeah I guess defaulting to ``null`` is better then

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -44,7 +64,16 @@ public function process(BinaryInterface $binary)
...
+
+        if ($this->level !== null) {
+            $pb->add(sprintf('--o%d', $this->level));
+        }
+
+        if ($this->stripAll) {
+            $pb->add('--strip=all');
+        }
+
+        $pb->add($input = tempnam(sys_get_temp_dir(), 'imagine_optipng'));
```
notrix: Sorry for my late comment, but I have noticed that there is merge conflict. $input gets added twice.

antoligy: Oops, nice catch!

lsmith77: can you do a follow up PR?

PR #762, opened 2016-07-22
Input is added twice in the OptiPngProcessor.

antoligy:
Thanks to @notrix for spotting this!

### DIFF
--- a/Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
@@ -73,7 +73,6 @@ class OptiPngPostProcessor implements PostProcessorInterface
             $pb->add('--strip=all');
         }
 
-        $pb->add($input = tempnam(sys_get_temp_dir(), 'imagine_optipng'));
         $pb->add($input);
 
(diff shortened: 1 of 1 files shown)


antoligy:
Sorry to bother you again @lsmith77, would you mind looking at this one?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -74,7 +74,6 @@ public function process(BinaryInterface $binary)
         }
 
         $pb->add($input = tempnam(sys_get_temp_dir(), 'imagine_optipng'));
```
notrix: This line should be removed as $input is defined in line 62

antoligy: 👍

lsmith77:
ah .. there it is already

antoligy:
Yep, tests passing - Not sure how it slipped by in the rebase, I must have resolved one of the merges incorrectly the first.
I'm thinking it might (separately) be worth adding some tests to prevent this kind of regression.

Issue #728, opened 2016-04-19
Downscale and Upscale are the same


trsteel88:
Has anyone noticed that DownscaleFilter and UpscaleFilter are exactly the same (with the exception of the rounding)?

https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/Loader/DownscaleFilterLoader.php
https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/Loader/UpscaleFilterLoader.php

I think something has gone wrong here. Also, if you use DownscaleFilter and set something large (e.g 1200x1200) and the original images if 400x400, it actually increases the size to 1200x1200. Shouldn't it stay at 400x400 because it's a Download (not an upscale)?

michellesanver:
Looks weird, is this still an issue? Need to investigate :)

franmomu:
I made an example in https://github.com/franmomu/liip-imagine-bundle-examples and it looks like it works. With an image of 400x400 if the downscale filter is set to 1200x1200 it does nothing and also if the upscale filter is set to 300x300 it does nothing. I guess the `isImageProcessable` method is the one who determines if should scale or not.

PR #770, opened 2016-08-14
Fix #769

### LINKED ISSUE #769: BC break causing Length of either side cannot be 0 or negative error
I believe this has been a common error from time to time, **Length of either side cannot be 0 or negative, current size is 0x0** * in vendor\imagine\imagine\lib\Imagine\Image\Box.php at line 42*, this error is appearing with the new release 1.6.0 and not when I revert back to 1.5.3

deviprsd21:
This permanently and efficiently fixes the problem

### DIFF
--- a/Imagine/Filter/Loader/DownscaleFilterLoader.php
+++ b/Imagine/Filter/Loader/DownscaleFilterLoader.php
@@ -29,9 +29,9 @@ class DownscaleFilterLoader implements LoaderInterface
         if ($origWidth > $width || $origHeight > $height) {
             $widthRatio = $width / $origWidth;
             $heightRatio = $height / $origHeight;
-            
+
             // faster check than is_null
-            if($width === null || $height === null) { 
+            if ($width === null || $height === null) { 
                 $ratio = max($widthRatio, $heightRatio);
             } else {
                 $ratio = min($widthRatio, $heightRatio);
--- a/Imagine/Filter/Loader/DownscaleFilterLoader.php
+++ b/Imagine/Filter/Loader/DownscaleFilterLoader.php
@@ -31,7 +31,7 @@ class DownscaleFilterLoader implements LoaderInterface
             $heightRatio = $height / $origHeight;
 
             // faster check than is_null
-            if ($width === null || $height === null) { 
+            if ($width === null || $height === null) {
                 $ratio = max($widthRatio, $heightRatio);
             } else {
                 $ratio = min($widthRatio, $heightRatio);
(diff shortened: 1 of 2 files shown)


Imagine/Filter/Loader/DownscaleFilterLoader.php
```diff
@@ -30,7 +30,12 @@ public function load(ImageInterface $image, array $options = array())
             $widthRatio = $width / $origWidth;
             $heightRatio = $height / $origHeight;
 
-            $ratio = min($widthRatio, $heightRatio);
+            // faster check than is_null
+            if ($width === null || $height === null) {
+                $ratio = max($widthRatio, $heightRatio);
+            } else {
+                $ratio = min($widthRatio, $heightRatio);
+            }
```
antoligy: could this be `$ratio = ($width === null || $height === null) ? max($widthRatio, $heightRatio) : min($widthRatio, $heightRatio);` ?

deviprsd21: Of course, either way that is just a style thing, I don't think it will be a performance issue, so I'm okay with anything.

Imagine/Filter/Loader/DownscaleFilterLoader.php
```diff
@@ -30,7 +30,16 @@ public function load(ImageInterface $image, array $options = array())
...
+
+            if ($ratio > 1) {
+                return $image;
+            }
```
deviprsd21: Could be controversial thing, discuss on this :8ball:

antoligy: hah, now that's my kinda fix! ;)

antoligy:
The diff for this PR is a little large, would it make sense to separate the fixes for #728 into a separate PR?

Otherwise, this makes a lot of sense - Thoughts @lsmith77 ?

deviprsd21:
@antoligy Wouldn't be such a bad idea! Just waiting for @lsmith77 to check it.

Imagine/Filter/Loader/DownscaleFilterLoader.php
lsmith77: in Symfony we generally use ``null === $foo``

Imagine/Filter/Loader/DownscaleFilterLoader.php
lsmith77: this should be an option, defaulting to off

deviprsd21: I was thinking more like, creating another filter "scale" which will do either, the filters should do only what they should, I don't know maybe it's just clear that way.

lsmith77: fine for me too

deviprsd21:
Then I'll close this and work on the filter, with another feature for these filters

PR #773, opened 2016-08-17
Scale filter and Downscale and Upscale as derivatives, with a new feature

deviprsd21:
Like I suggested in #770 Creating a new filter **scale** and deriving upscale and downscale from it will be much clear. Other than that, there are new options like **by** or **to** depending upon filter where a relative ratio or absolute ratio respectively can be set to up or down scale the image. For example, 

```yaml
filters:
    downscale: { max: [50, 70] } #or { by: 0.6 } -> Downscales to 40%
    upscale: { min: [200,null] } #or { by: 0.7 } -> Upscales to 170%
    scale: { dim: [600,750] } #or { to: 1.56 } -> Upscales to 156% | { to: 0.66 } -> Downscales to 66%
```
Though relative_resize has a scale method, the scale filter is more flexible than the latter.

### DIFF
--- a/Imagine/Filter/Loader/DownscaleFilterLoader.php
+++ b/Imagine/Filter/Loader/DownscaleFilterLoader.php
@@ -2,41 +2,22 @@
 
 namespace Liip\ImagineBundle\Imagine\Filter\Loader;
 
-use Imagine\Filter\Basic\Resize;
-use Imagine\Image\ImageInterface;
-use Imagine\Image\Box;
-
 /**
- * downscale filter.
+ * Downscale filter.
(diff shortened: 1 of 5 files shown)


lsmith77:
can you update the docs as well?

antoligy:
Looks way nicer!

deviprsd21:
I'll write some tests as well

deviprsd21:
After giving tests a good look, I don't think new tests are required. If upscale and downscale work, the scale should be assumed to work.

lsmith77:
thx

PR #775, opened 2016-08-25
Amend path resolution handlers and outside root check conditional in FileSystemLoader

### LINKED ISSUE #763: Finding if path is outside the root path check is wrong
LiipImagineBundle/Binary/Loader/FileSystemLoader.php lines 48-49 are just wrong.
First of all if I have ../ anywhere in the path this will throw an exception (how i actually discovered the problem) so if I have root/dir1/dir2/../dir2_1/A ... you get the point. 
Furthermore if somebody writes the path with the other slash '\' it will not pass the check even if it really goes outside of the defined root. As you can see checking if a path is a outside or a root folder is not a problem that can be solved with 1 line of code obviously.

robfrawley:
Fixes #763 by allowing "../" in both root and asset paths. Loading assets is still "chrooted" within the root path using a simple (and quick!) strpos check.

Test included!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 composer.phar
 vendor/
 Tests/Functional/app/web/media/cache
+.idea/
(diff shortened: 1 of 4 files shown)


robfrawley:
So... Any thoughts from @deciders?

lsmith77:
looks good to me ..

PR #777, opened 2016-08-29
Add visibility argument to flysystem resolver

cedricziel:
Before this, there was no way to modify the visibility in a central place,
which was fine for local resolvers, or with predefined acls.

This changeset introduces a new `visibility` argument on the FlysystemResolver
that will be taken into account everytime a cache object is written to a flysystem
filesystem.

The default is public, so that cached objects are accessible from the outside,
which should be sensible.


antoligy:
:thumbsup:

Are there any other config options which can be exposed this way? (Took a look and couldn't see anything obvious!)

cedricziel:
Most options are really adapter specific and those are configured on the underlying adapter instances. So far I think only visibility is really a shared option.

I thought about a mechanism for custom arguments, but that is something for another PR. We are already using the pull-request branch in production and visibility was really the only thing I had an urgent need for.

Thank you for reviewing, I consider it done, unless you you want me to correct something :)

PR #764, opened 2016-08-03
Implemented ConfigurablePostProcessorInterface in OptiPngPostProcessor

jehaby:

PR #778, opened 2016-08-30
Tempdir for postprocessors

jehaby:
Hi again, guys! 

My company use this bundle in production (thank you very much for this, buy the way). 
We had a problem with too much disk writes because of use OptiPng and Jpegoptim post processors write temporary files on hard drive. I suggest to use for this purpose `/run/shm` folder by default, which should be present on all modern linux distros. 

If you're not ok with that, I can return previous default directory. 

PS: I thought about implement it using stdin and stdout like in others PostProcessors, but Optipng doesn't support this feature, and Jpegoptim supports it only for version 1.4+ (while default version in Ubuntu 14.04 is 1.3).

PS2: also I fixed bugfix, accidentally introduced by me a in [previous PR](https://github.com/liip/LiipImagineBundle/pull/764/files#diff-22cf136767e2f2ef8d453e5673d09397R55) :)

### DIFF
--- a/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
@@ -35,19 +35,33 @@ class JpegOptimPostProcessor implements PostProcessorInterface, ConfigurablePost
     protected $progressive;
 
     /**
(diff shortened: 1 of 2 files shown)


jehaby:
Unable to reopen after deletion of branch.

PR #787, opened 2016-09-06
List deprecation - closes #731

### LINKED ISSUE #731: null values for size in php7
After upgrading from php 5.5 to php 7, my thumbnails stopped working. This is related to this [incompatible change](http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.list.empty) made in 7.
My previous configuration for a thumbnail was like `thumbnail: { size: [84, ], mode: outbound }` and I had to change it to  `{ size: [84, null], mode: outbound }` (notice the explicit `null`).
I think you should do at least one of the following:
* do not allow empty values for `size`
* add a control to [code](https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Filter/Loader/ThumbnailFilterLoader.php#L28)
* add a warning in docs

antoligy:
- [x] Migrating away from `list()`, due to its unpredictable nature between different versions of PHP.
- [x] Increasing code coverage of FilterLoaders with regression tests.
- [x] Enabling test suite for PHP 7.1.

### DIFF
--- a/Imagine/Filter/Loader/ScaleFilterLoader.php
+++ b/Imagine/Filter/Loader/ScaleFilterLoader.php
@@ -24,7 +24,7 @@ class ScaleFilterLoader implements LoaderInterface
     protected $ratioKey;
 
     /**
-     * @var boolean
+     * @var bool
      */
     protected $absoluteRatio;
 
@@ -51,8 +51,8 @@ class ScaleFilterLoader implements LoaderInterface
         if (isset($options[$this->ratioKey])) {
             $ratio = $this->absoluteRatio ? $options[$this->ratioKey] : $this->calcAbsoluteRatio($options[$this->ratioKey]);
         } elseif (isset($options[$this->dimensionKey])) {
-            $size   = $options[$this->dimensionKey];
-            $width  = isset($size[0]) ? $size[0] : null;
+            $size = $options[$this->dimensionKey];
+            $width = isset($size[0]) ? $size[0] : null;
             $height = isset($size[1]) ? $size[1] : null;
 
             $widthRatio = $width / $origWidth;
--- a/Imagine/Filter/Loader/ScaleFilterLoader.php
+++ b/Imagine/Filter/Loader/ScaleFilterLoader.php
@@ -13,9 +13,24 @@ use Imagine\Image\Box;
  */
 class ScaleFilterLoader implements LoaderInterface
 {
-    public function __construct($dimentionKey = 'dim', $ratioKey = 'to', $absoluteRatio = true)
+    /**
+     * @var string
+     */
+    protected $dimensionKey;
+
+    /**
+     * @var string
+     */
+    protected $ratioKey;
+
+    /**
+     * @var boolean
+     */
+    protected $absoluteRatio;
+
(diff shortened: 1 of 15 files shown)


Tests/Imagine/Filter/Loader/ThumbnailFilterLoaderTest.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+      return [
+        [100, 100],
+        [50, 50],
+        [1, 30],
+        [0, 30],
+        [30, 0],
+        [null, 16],
+        [37, null],
+        [null, null]
```
antoligy: This was never supported.

Tests/AbstractTest.php
```diff
@@ -54,51 +54,51 @@ protected function createFilterConfiguration()
 
     protected function getMockCacheManager()
     {
-        return $this->getMock('Liip\ImagineBundle\Imagine\Cache\CacheManager', array(), array(), '', false);
+        return $this->getMockBuilder('Liip\ImagineBundle\Imagine\Cache\CacheManager')->getMock();
```
robfrawley: So happy someone is taking the time to update the test suite to remove PHPUnit 5.x deprecations. I have a similar branch that removes all `getMock` methods in favor of `getMockBuilder`. I'll wait until this is merged to submit my PR for this.

antoligy:
I'll defer writing the appropriate documentation until after #789 is merged, there should not be much of it.  As a follow-up to this I may submit further PRs focusing purely on test coverage.

antoligy:
What do you think @lsmith77 / @robfrawley ?

robfrawley:
@antoligy Looks great to me. New tests introduced increase total test coverage from 61.6% to 65.4% --- and increases total test coverage of `Loaders` from 22.71% to 47.26%.

:+1:

antoligy:
Yeah thanks that feels pretty good!  I might have another pass at writing some tests to further increase Loader coverage as there's definitely a few easy wins in there, but that I can do irrespective of this PR.

The goal here was to remove `list()`, or to at least replace it in cases where we might likely run into a "breaking" change between 5 and 7.

robfrawley:
Adding cross-reference to issue #731 (issues in titles don't create cross-reference links).

Imagine/Filter/Loader/ScaleFilterLoader.php
```diff
@@ -13,9 +13,24 @@
...
+    protected $ratioKey;
+
+    /**
+     * @var bool
+     */
+    protected $absoluteRatio;
+
+    public function __construct($dimensionKey = 'dim', $ratioKey = 'to', $absoluteRatio = true)
     {
-        $this->dimentionKey = $dimentionKey;
```
robfrawley: @antoligy Wow, can't believe we had filters creating dynamic class properties...

antoligy: Oh, it still does, I plan on restructuring this at a later date - The ScaleFilter is inherited by the DownscaleFilter and UpscaleFilter both of which use slightly different parameters.

antoligy:
Paging dr @lsmith77 😄

PR #790, opened 2016-09-07
Add LICENSE.md

robfrawley:
Add the MIT License to the project root directory with a copyright line reflective of those who have contributed.


antoligy:
Should this also reference the original Avalanche bundle?

robfrawley:
Good question... I don't have a firm answer there. The purpose of this pull request is to move the license file ([currently burried under Resource/meta/LICENSE](https://github.com/liip/LiipImagineBundle/blob/1.0/Resources/meta/LICENSE)), update dates, move to a markdown varient, and add a link to the GitHub contributors page.

But it all started because it annoyed me that it took 10 minutes of hunting around to find this bundle's license initially. ;-)

That being said, even though the current license doesn't include a reference to Avalanche, that could have very likely been an oversight at the time. I'm more than happy to add a reference to Avalanche. Thoughts?

robfrawley:
After a quick trip down Git history, I can confirm this code base absolutely shares commit ancestors with Avalanche from early to late 2011, including a few instances of files of files still existing verbatim from that period in history.

Therefore, I've gone ahead and added the original copyright author that appears in the Avalanche codebase from the shared 2011 commits, providing us with the following [updated license file](https://github.com/robfrawley/LiipImagineBundle/blob/ad7bb953fa31ea5b3db592bb8d4d3cc3da408743/LICENSE.md).

antoligy:
Huge thanks for looking into that!

lsmith77:
the license file is in the current location due to Symfony standards .. then again they might have evolved since then .. 

just checked .. yeah it seems like the new standard is to put it into the root as "LICENSE":
http://symfony.com/doc/current/bundles/best_practices.html

robfrawley:
@lsmith77 Do you want to drop the markdown from the LICENSE? I'll update this PR if that is desired.

lsmith77:
checking via twitter https://twitter.com/lsmith/status/774138202567258112

robfrawley:
@lsmith77 Looks like Twitter has spoken. Updated.

PR #792, opened 2016-09-09
Added php_cs.dist / Updated .styleci.yml / Fixed and updated .travis.yml

robfrawley:
# Description

Adds `php-cs-fixer` and `php-cs-fixer-styleci-bridge` as development dependencies, enabling automatic configuration of `bin/php-cs-fixer` using the `.styleci.yml` file.

Now, running `php-cs-fixer` is easy, which will hopefully provide incentive for regular use of [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer):

```
bin/php-cs-fixer fix ./
```

## Additional Details

The `.styleci.yml` file was updated to enable a few additional, sensible options on top of the `symfony` preset. For example, as PHP 5.3 is supported, the `long_array_syntax` rule was added. **This caught a bug in the current production code base, [seen here](https://github.com/src-run/liip-imagine-bundle-sandbox/commit/a27d221a12c7e09bf7eb10f89dd3392b8d591738#diff-741c7f1cd78fc4c24ff27bd7659a6fd7L135)**.

The other, additional rules add consistency I believe worthwhile, such as:
- ordering file imports
- ordering phpdocs

Reference [their fixer docs](https://styleci.readme.io/docs/fixers) for details about what all the new rules do and provide feedback if you disagree with any of the newly introduced rules.

## Original Commit Message

- Added "friendsofphp/php-cs-fixer:~2.0" development dep.
- Added "sllh/php-cs-fixer-styleci-bridge:~2.1" development dep to enable php-cs-fixer configuration directly from .styleci.yml config file.
- Updated "styleci.yml" file rule additions (on top of symfony preset):
  - combine_consecutive_unsets
  - linebreak_after_opening_tag
  - no_short_echo_tag
  - ordered_imports
  - phpdoc_order
  - php_unit_construct
  - php_unit_dedicate_assert
  - long_array_syntax
- Added "sort-packages" to "config" key in "composer.json"
- Updated "suggest" key descriptions in "composer.json" for clarity.

PR #793, opened 2016-09-09
Bugfix for Failing Tests Introduced in #777

robfrawley:
This is a bugfix for failing tests introduced in #777. Sadly, this wasn't caught because the travis config conditional to include `league/flysystem` contained an incorrect conditional that would always resolve to false ([due to this line](https://github.com/liip/LiipImagineBundle/blob/1.0/.travis.yml#L35)) which caused [all flyststem tests to be skipped](https://travis-ci.org/liip/LiipImagineBundle/jobs/158697869#L311).

So, *the tests for this pull requests don't actually test the fix* because I inadvertently already fixed the travis file in #792.

So: this special branch includes both this pull request and #792: https://travis-ci.org/src-run/liip-imagine-bundle-sandbox/builds/158721955


antoligy:
Looks good 😄

robfrawley:
@lsmith77 This is a pretty important bug fix, as the test suite fails without it (it's only been green on Travis b/c it wasn't testing flyststem due to broken travis file fixed in #792)

PR #797, opened 2016-09-16
Fix travis build definition for conditional dependencies

cedricziel:
This change adjusts the before_install section to correctly install the optional dependencies.

Before this commit, the if statements silently failed (you can see that in the older build-logs), and the corresponding tests were skipped.

The patch ensures that the dependencies for `ext-mongo`, the `odm` and `flysystem` are installed so that the tests aren't skipped anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,12 +29,12 @@ matrix:
       env: COMPOSER_FLAGS="--prefer-lowest"
 
 before_install:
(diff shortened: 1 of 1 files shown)


robfrawley:
Thanks for the PRs! Looks like the issue you are targeting was already addressed in #792.

cedricziel:
Ahh, snap! :)

Thank you for the hint.

PR #799, opened 2016-09-19
Bugfix: Remove Short Array Syntax and Fix (Minor) Recent Merge Issues

robfrawley:
Remove accidental use of short array syntax in my PR that created .php_cs.dist

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -22,7 +22,7 @@ For the full copyright and license information, please view the LICENSE.md
 file that was distributed with this source code.
 EOF;
 
-$headerConfig = [ 'header_comment' => [ 'header' => $headerString ] ];
+$headerConfig = array( 'header_comment' => array( 'header' => $headerString ) );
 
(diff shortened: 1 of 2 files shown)


robfrawley:
@antoligy Somehow, [your php 7.1 travis addition](https://github.com/liip/LiipImagineBundle/blame/master/.travis.yml#L11) passed its build on your PR and only *now* fails to inform us we need to update the conditional to exclude mongo on 7.1 too. ;-) IDK why...

robfrawley:
@antoligy Ah, figured it out. The second half of the mongo conditional in the old travis file was broken, which is the only reason mongo wasn't included for 7-based builds (not because, as one would expect, the conditional was evaluated).

With my fixed travis file merged #792 *acting in combination with* inclusion of `7.1` in #787 with no check, caused the interesting case of your PR being green, and then our merged code being red.

All is well though, should be fixed.

antoligy:
Nice catch, cheers for fixing it!

PR #791, opened 2016-09-07
Add PHP File-Level DocBlocks Globally

robfrawley:
Add the following doc-block to all PHP files in this project:

```
/*
 * This file is part of the `liip/LiipImagineBundle` project.
 *
 * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
 *
 * For the full copyright and license information, please view the LICENSE.md
 * file that was distributed with this source code.
 */
```

This has been my file-level doc-block style of choice for quite some time (some may notice it is directly modeled after [Symfony's](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php#L3)).

That being said, I am more than open to alterations if the consensus suggests one.

PR #800, opened 2016-09-19
Update Source to Newly Merged Style Rule Additions

robfrawley:
This replaces #791 (which was done prior to the new "style rules" and "php-cs-fixer bridge" additions using a hand crafted regex of mine - I feel more comfortable having a real code tokenizer/lexer do it instead, heh).

**So, this is simple a PHP CS Fixer run using our new default rule set.**

While the change set is substantial in size, its actually quite simple. All the changes relate to one of the following three rules:
- Ensure all files have the proper file-level doc block.
- Ensure all class/method/etc doc blocks use consistent ordering of tags.
- Ensure all imports are ordered.

The only, singular exception to my above assertion is a [short-array to long-array syntax fix](https://github.com/liip/LiipImagineBundle/compare/master...src-run:bugfix-phpcsfixer-run?expand=1#diff-741c7f1cd78fc4c24ff27bd7659a6fd7L135) caught the fixer.

Hopefully people run this more often now! Remember, it's as easy as this now:

```
bin/php-cs-fixer fix
```

### DIFF
--- a/Imagine/Cache/Resolver/FlysystemResolver.php
+++ b/Imagine/Cache/Resolver/FlysystemResolver.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
 namespace Liip\ImagineBundle\Imagine\Cache\Resolver;
 
 use League\Flysystem\AdapterInterface;
@@ -107,9 +116,9 @@ class FlysystemResolver implements ResolverInterface
      * @param string $path   The path where the original file is expected to be
      * @param string $filter The name of the imagine filter in effect
      *
-     * @return string The absolute URL of the cached image
-     *
      * @throws NotResolvableException
+     *
+     * @return string The absolute URL of the cached image
      */
     public function resolve($path, $filter)
     {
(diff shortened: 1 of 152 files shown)


robfrawley:
Build failure unrelated to this PR and fixed in #799 (see [comment](https://github.com/liip/LiipImagineBundle/pull/799#issuecomment-247962004)). I'll rebase this after #799 is merged so the tests pass.

robfrawley:
@lsmith77 Re-based on master (with #799 merged) fixes tests; this should be good to go now.

Imagine/Cache/Resolver/FlysystemResolver.php
```diff
@@ -132,7 +141,7 @@ public function store(BinaryInterface $binary, $path, $filter)
         $this->flysystem->put(
             $this->getFilePath($path, $filter),
             $binary->getContent(),
-            ['visibility' => $this->visibility]
```
lsmith77: looks like this caught another use of short array syntax which means this isn't tested

robfrawley: Yeah, the the travis file always excluded `flysystem` from PHP 5.3 runs. I've been using the assumption FlySystem requires a higher PHP version, but if that's not the case we should obviously be testing it! I'll take a look at flysystems composer when I get a second.

robfrawley: @lsmith77 Yup, FlySystem requires PHP > 5.4, so that's why it is excluded from 5.3 builds on travis and why the above wasn't being tested.

PR #803, opened 2016-09-21
Add data_loader config to doc

davidfuhr:
If not set the default loader is used which is the local adapter.


robfrawley:
Thanks @davidfuhr

Issue #782, opened 2016-09-04
Breaking Changes / Road-map v2


robfrawley:
# Road-map 2.0

*Note: this will serve as an alternate, more ambitious (by which I mean, no regard for BC breaks whatsoever with an eye toward the future and disregard for the past), alternative to #686.*

- [ ] **Code cleanup** (consider ways to remove duplication and otherwise refactor code for the better)
- [ ] **Symfony 2.8+** (with *no usage of deprecated features, full 3.0+ compatibility*)
- [ ] **PHP 5.6+** (or maybe 7.0+)
- [ ] **LoaderInterface Rename** (rename LoaderInterfaces under `Liip\ImagineBundle\Imagine\Filter\Loader` and `Liip\ImagineBundle\Binary\Loader` to remove confusion)
- [ ] **FilterLoader Rename** ("Loader" is confusing in this context, perhaps "FilterProcessor" or "FilterRunner" or similar?)
- [ ] **Binary/FileBinary Standardization** (Remove ability to return or provide as argument both models in a bunch of code, standardize on one wherever possible.)
- [ ] **Move Controller Business Logic into Service** (Allow for use of service without Request)
- [ ] **Merge ConfigurablePostProcessorInterface and PostProcessorInterface**
- [ ] **See about supporting usage cases like #733**
- [ ] **Conversion between formats (ex: PNG -> JPG)**

More to be decided... (post additions in the comments; ideas welcome!)

### @antoligy 

- Currently Symfony's minimum version requirement is 5.5.9
- [Imagine supports 5.3](https://github.com/avalanche123/Imagine/blob/develop/composer.json#L23); Perhaps such an upgrade should be a collaboration effort?
- We gain [many PHP7 benefits due to] [potential new IMagick transforms and features](https://pecl.php.net/package-changelog.php?package=imagick).
- [Not the same for GMagick sadly](https://pecl.php.net/package-changelog.php?package=gmagick).
- [Plenty of bugfixes to GD](https://secure.php.net/ChangeLog-7.php).

## Original Message

@lsmith77 Any interest in creating a 2.x branch where we can start shaping this bundle's next major release? It's time to drop PHP versions that are no longer supported, as well as Symfony versions that have reached EOL. There are some structural changes that could benefit the code base as well.

I'd love to help!

lsmith77:
I am certainly open to exploring a 2.x.

[...]

PR #804, opened 2016-09-27
Refactoring logic to service

jehaby:
I tried to move all logic from controller to service, as was suggested [here](https://github.com/liip/LiipImagineBundle/issues/782)

The service has no dependencies on HTTP components. But controller probably still has too many dependencies, especially on exceptions. And the code doesn't look very clean. 

So I wanted to have some feedback and advice on how to make it better.


jehaby:
Oh, I wanted to make a pull request to sandbox repo ... 
It's too WIP for this one.

PR #805, opened 2016-09-27
Update basic-usage.rst

you-ser:
Error using $this at last line of code, looks like it must be $imagineCacheManager


robfrawley:
Thanks @you-ser - good catch!

PR #820, opened 2016-10-31
Updating 2.0 with corrections from 1.0

antoligy:


robfrawley:
@antoligy @lsmith77 Operations like this are much more appropriate as a CLI manual re-base than a PR/merge in the future, I would propose. Such avoids odd looking commit history like the duplication of commits and other oddities that may result (for example, see the commit duplicates on "Oct 30, 2016" on the 2.0 branch (that don't exist on 1.0) b/c of this PR: https://github.com/liip/LiipImagineBundle/commits/2.0

lsmith77:
I personally do not feel passionate about keeping the history clean at this level. it really has never really bitten me. but if you prefer I am fine to keep an eye on this in the future.

PR #825, opened 2016-11-25
Wrap relative path with asset() Twig function

bocharsky-bw:
I think it's a good idea to adjust relative path with `asset()` Twig function, isn't it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ be to pipe the path of your image to the provided `imagine_filter` Twig
 filter.
 
 ```twig
-<img src="{{ '/relative/path/to/image.jpg' | imagine_filter('my_thumb') }}" />
+<img src="{{ asset('/relative/path/to/image.jpg') | imagine_filter('my_thumb') }}" />
 ```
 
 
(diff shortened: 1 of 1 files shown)


robfrawley:
@bocharsky-bw So, I know the **Asset Component** was introduced in Symfony 2.7, but did this function exist prior to the component being compartmentalized into its own release?

Most importantly, was this function available in a default Symfony 2.3 installation? I honestly don't remember if it was, and I do know we want to retain explicit 2.3 support on this branch...

That said, if it did exist, let me know, otherwise, I'll merge it into the 2.x branch instead, as we are breaking BC on that branch, and I do like the overall sentiment of this change.

lsmith77:
http://symfony.com/doc/2.3/reference/twig_reference.html#asset

robfrawley:
@lsmith77 Also, I didn't intentionally turn on the Coveralls verbose comment reporting, and I'm not quite sure I like it, either. Thoughts? I might go ahead and disable that for this project (it's already added to the repo checks, we don't need strings of comments from the coveralls bot on top of that, do we?)

lsmith77:
@robfrawley imho its sufficient if such coverage infos can easily be seen via a link in the checks.

bocharsky-bw:
Now I'm not sure about this changes. What if assets will be hosted on a CDN? Then the `asset()` function will build an absolute URL for them I think. Looks like it works for relative paths only.

lsmith77:
the function lets you configure different CDN base urls .. but I guess for most users it makes sense to use the ``asset`` filter ..

antoligy:
:tada:

PR #827, opened 2016-12-01
Let the client pick the scheme

greg0ire:
I can't figure out why my templates still output `http` schemes when my page is in https, I have configured trusted proxies, and even tried harcoding the request context scheme. The request context scheme is http when arriving in the the web path resolver.

I propose this alternative, that fixes my bug with less code : let's use '//' urls everywhere, and let the client figure out which scheme should be used.

### DIFF
--- a/Imagine/Cache/Resolver/WebPathResolver.php
+++ b/Imagine/Cache/Resolver/WebPathResolver.php
@@ -158,8 +158,7 @@ class WebPathResolver implements ResolverInterface
         }
         $baseUrl = rtrim($baseUrl, '/\\');
 
-        return sprintf('%s://%s%s%s',
-            $this->requestContext->getScheme(),
+        return sprintf('//%s%s%s',
             $this->requestContext->getHost(),
             $port,
             $baseUrl
--- a/Tests/Functional/Command/ResolveCacheTest.php
+++ b/Tests/Functional/Command/ResolveCacheTest.php
@@ -58,7 +58,7 @@ class ResolveCacheTest extends WebTestCase
 
         $this->assertFileExists($this->cacheRoot.'/thumbnail_web_path/images/cats.jpeg');
         $this->assertFileNotExists($this->cacheRoot.'/thumbnail_default/images/cats.jpeg');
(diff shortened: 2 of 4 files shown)


cedricziel:
I'm fine with the change itself, but is it possible that you're behind a proxy that doesn't set the correct headers like AppEngine?

greg0ire:
I'm behind the proxy, and I configured the trusted proxies, which seems to work (I have far less urls that are http when they should be https), but for some reason, not completely, unless I apply this change, which is something good IMO b/c it's one less thing to worry on your side.

greg0ire:
Regarding my problem, I found that there is a sub-request, and that is where the problem occurs, so I shall find a workaround soon.

greg0ire:
Adding `127.0.0.1` to the trusted proxies solves my problem…

antoligy:
Hi @greg0ire, thanks for the contribution!

I don't think that this change is actually required as Symfony will derive the scheme from the `X_FORWARDED_PROTO` header which is usually sent as a part of a proxied request.  Enabling your reverse proxy as a trusted proxy means that Symfony respects this header, and in turn will prefix URLs with the `https` scheme correctly.  ([See here](https://github.com/symfony/http-foundation/blob/master/Request.php#L1163-L1186)).

Considering this, the use-cases where the scheme might differ from the main request are quite rare, as when connecting via SSL, the plaintext request to fetch images is blocked by browsers by default.  The inverse situation is very rare, and I don't know if we should be supporting this use-case.

However perhaps it is worth adding this case to the documentation.

Hope this helps!
Alex.

antoligy:
Actually, this is a bigger change as some browsers don't handle schemeless URLs correctly.
e.g. IE7/8 apparently will download files referred to this way twice.

Going to close this, and will raise a separate issue about adding this case to the FAQs - Thanks anyway @greg0ire!

greg0ire:
@antoligy : as you can see I fixed my problem, but I thought this might help you support current use-cases with less code. ~~However I understand if you don't want to merge this by fear this might break something in unforeseen ways.~~ Looks like it does break something, good job dodging this.

PR #831, opened 2016-12-04
Skip tests with dependencies that are not installed by default

rpkamp:
I just checked out the repository for the first time, ran phpunit, and got a bunch of fails, one of which was a fatal PHP error that stopped the tests because symfony templating was not installed (`PHP Fatal error:  Class 'Symfony\Component\Templating\Helper\Helper' not found`).
By skipping the tests that require components that are not installed by default at least all the tests run out of the box.
Alternatively the missing dependencies could be added to the dev dependencies but I did not want to be presumptuous and just add them.

### DIFF
--- a/Tests/Functional/WebTestCase.php
+++ b/Tests/Functional/WebTestCase.php
@@ -15,6 +15,17 @@ use Symfony\Bundle\FrameworkBundle\Test\WebTestCase as BaseWebTestCase;
 
 abstract class WebTestCase extends BaseWebTestCase
 {
+    public function setUp()
+    {
+        if (!class_exists('Symfony\Component\Asset\Package')) {
+            $this->markTestSkipped('The symfony/asset PHP library is not available.');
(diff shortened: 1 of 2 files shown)


xabbuh:
This is probably related to the fact that the bundle's dependency on FrameworkBundle pulled in a lot of packages in the past. Now that with the release of Symfony 3.2 the FrameworkBundle doesn't pull in the whole world anymore, it's probably the best idea to update the required packages for this package.

rpkamp:
Good point. I created an alternative PR to require the components explicitly: https://github.com/liip/LiipImagineBundle/pull/832

antoligy:
This feels like something we should probably fix in the 2.0 branch, and it's probably for the best that the tests fail in other versions

rpkamp:
No longer needed now that https://github.com/liip/LiipImagineBundle/pull/832 is merged

PR #834, opened 2016-12-06
Updating twig call to utilise asset(), to match README.md (closes #830)

### LINKED ISSUE #830: Update RST documentation to use `asset()` Twig function
The `README.md` code examples now make use of the Twig `asset()` function, per #825. We should bring the relevant examples in the RST documentation in-line with this convention. While the prior mentioned PR affected only four `README.md` lines, I know that their are a few additional examples in the RST documentation, all of which should be consistent, IMHO.

antoligy:
I could only find two applicable Twig calls, there is another under the GridFS data-loader page however I believe this uses Mongo document IDs and so wouldn't work with `assetic()`.

### DIFF
--- a/Resources/doc/basic-usage.rst
+++ b/Resources/doc/basic-usage.rst
@@ -103,7 +103,7 @@ your newly defined ``my_thumb`` filter set immediately within a template.
 
     .. code-block:: html+twig
 
-        <img src="{{ '/relative/path/to/image.jpg' | imagine_filter('my_thumb') }}" />
+        <img src="{{ asset('/relative/path/to/image.jpg') | imagine_filter('my_thumb') }}" />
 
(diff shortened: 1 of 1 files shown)


antoligy:
Not sure why the builds are failing here, it seems to be a failure during the git clone ...

robfrawley:
@antoligy They failed b/c you renamed your branch from `use-asset-for-twig` to `update-basic-usage-assetic`

robfrawley:
@antoligy You should make branches for PRs on your own fork and no directly on this repo, IMHO. Otherwise, you end up creating a new version on packagist names `dev-update-basic-usage-assetic`

antoligy:
Ah good call, I'll close this and reopen under another branch.  Cheers!

PR #836, opened 2016-12-11
Updating twig call to utilise asset(), to match README.md (closes #830)

### LINKED ISSUE #830: Update RST documentation to use `asset()` Twig function

antoligy:
See #834, only on my fork and with without me breaking the test suite.

### DIFF
--- a/Resources/doc/basic-usage.rst
+++ b/Resources/doc/basic-usage.rst
@@ -103,7 +103,7 @@ your newly defined ``my_thumb`` filter set immediately within a template.
 
     .. code-block:: html+twig
 
-        <img src="{{ '/relative/path/to/image.jpg' | imagine_filter('my_thumb') }}" />
+        <img src="{{ asset('/relative/path/to/image.jpg') | imagine_filter('my_thumb') }}" />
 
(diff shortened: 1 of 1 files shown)


robfrawley:
Hmm... why is Travis trying to clone branch `feature/use-asset-for-twig`?

antoligy:
Not sure, I reckon it's cached a copy of the repo with my initial incorrect branch and it's trying to somehow resolve that commit to that branch - All the more reason to avoid pushing to branches on here!

robfrawley:
@antoligy Cherry picked your commit for 2.x branch: https://github.com/liip/LiipImagineBundle/commits/2.0

antoligy:
Sounds good-  Is it worth establishing a standard for how we push 1.0 changes to 2.0?  I am happy to cherry pick but obviously this becomes more of a problem with more commits.

Issue #501, opened 2014-09-19
Exception: PHP exif extension is required to use the ExifMetadataReader


JasonHanley:
When upgrading to the lastest dev-master, I receive the following when simply trying to read an image:

Uncaught PHP Exception Imagine\Exception\NotSupportedException: "PHP exif extension is required to use the ExifMetadataReader" at /home/rumie/test-content/vendor/imagine/imagine/lib/Imagine/Image/Metadata/ExifMetadataReader.php line 25 {"exception":"[object] (Imagine\Exception\NotSupportedException: PHP exif extension is required to use the ExifMetadataReader

Probably better to simply not read the exif data rather than blowing up.

I suspect this used to either be optional or go through a different path.

lukaszrak:
Still not fixed.

x3r0z:
I fixed that problem change the value of the parameter liip_imagine.meta_data.reader.class in imagine.xml by default the value is Imagine\Image\Metadata\ExifMetadataReader and I change to Imagine\Image\Metadata\DefaultMetadataReader, if that parameter can change in the bundle config to that way not modified the bundle.

stefann-sasori:
@x3r0z editing imagine.xml solved the problem for me thank you.
If it can help, the absolute path of this file is : 
vendor\liip\imagine-bundle\Liip\ImagineBundle\Resources\config\imagine.xml

Miliooo:
I added this to my parameters: (config.yml)
liip_imagine.meta_data.reader.class: "Imagine\Image\Metadata\DefaultMetadataReader"

WebDevCodr:
@Miliooo Thanks for your solution, works great!

MAXakaWIZARD:
@lsmith77 this is not an issue with imagine lib, the bundle configures it with `ExifMetadataReader` which requires `ext-exif`. I think we need to change it to `DefaultMetadataReader` or configure it dynamically depending on `ext-exif` availability.

stickyhands:
@Miliooo @WebDevCodr I have an unrecognized option error here with that fix

cedricziel:
@stickyhands I just opened #841 which should fix the issue with a compiler pass, so you don't have to fiddle around with parameters here.

btw: @Miliooo surely meant setting it through a parameter and not directly in config.yml.

PR #841, opened 2017-01-03
Set DefaultMetadataReader when ext-exif is not present

cedricziel:
This fixes the issue where an exif reader is used, but ext-exif
is not loaded.

Closes: #501

### DIFF
--- a/DependencyInjection/Compiler/MetadataReaderCompilerPass.php
+++ b/DependencyInjection/Compiler/MetadataReaderCompilerPass.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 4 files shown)


robfrawley:
We need to allow for people to manually overwrite this parameter in their own configuration, no?

cedricziel:
Yes, sure. I think comparing the set parameter with the bundle default parameter should be sufficient. Fix upcoming.

robfrawley:
My thinking exactly. :+1:

cedricziel:
That should be it. Now the parameter is only re-set, if it's equal to the default value

antoligy:
Awesome!

Should we offer up some kind of warning or log message when the parameter is reset?   It's useful behaviour, but this is definitely something which might cause confusion if a user isn't aware of what's going on.

cedricziel:
Agreed. I changed the CompilerPass to emit a log message to the compiler log.

cedricziel:
I added a basic test. Let me know what you think. :)

antoligy:
I'm pretty happy with this!

lsmith77:
@robfrawley you have permissions to merge right?

lsmith77:
ah @cedricziel too :)

cedricziel:
Yep, I've just wanted to wait for reviews - thx for participating and the valuable feedback!

robfrawley:
@lsmith77 Sure do have merge rights! It's just been a bit since you've been around and many merges have happened with the "new, active maintainers crew"; my review request was just a good way to have you check in and make sure you were comfortable with everything so far. Thanks @cedricziel for the PR!

PR #847, opened 2017-01-11
Fix strange wording in readme

svenluijten:
To do something "pragmatically" is to do it from a practical point of view. I'd say this entire bundle is made to solve uploading "pragmatically". I think the word you were looking for in this section was "programmatically".


cedricziel:
Good catch, thank you!

antoligy:
Thanks for the contribution!  Merged and cherry-picked into 2.0.

svenluijten:
👌 Sweet!

robfrawley:
Good find!

Issue #846, opened 2017-01-10
"Cache directory creation problem" and "source image could not be found" with v1.7


davidromani:
Hello.

After upgrade this bundle from 1.6 to 1.7 my project doesn't works well. 

In my local environment everything works well. My "web/media/cache" dir exists with all permissions. But when I deploy my app to a production server with Capifony, it seems that this bundle can't create "cache" directory.

Is there some change related with directory creation in this update? In my deployment server I only have "web/media" dir with all permissions, the "web/media/cache" is not deployed...

antoligy:
I don't believe that there's been any changes to the creation (or lack thereof) of the `web/media/cache` directory.  I'll take a look to double-check...

(For reference - https://github.com/liip/LiipImagineBundle/compare/1.6.0...1.7.0)

adaniloff:
Hello there,

+1 with davidromani.

We got a problem with an update in the `Binary/Loader/FileSystemLoader.php` file.
I got the "Source image invalid "%s" as it is outside of the defined root path" error message since 1.7 upgrade.

Tried to downgrade to 1.6.* and it worked fine again.

It seems to be an environment problem (maybe because of symlinks), we're actually trying to figure it out; because it works perfectly on our local env but does not on production.

raphaChoquet:
Hello,

The error is due to symbolic links created by 'assets:install --symlink'.
I made a pull request to fix it :  
https://github.com/liip/LiipImagineBundle/pull/848

picks44:
Thanks @raphaChoquet! Hope there'll be a new release with this fix soon, I had to downgrade to 1.6.

antoligy:
Unfortunately this PR has broken the testing suite and so I'm not comfortable with merging and releasing it just yet.  However I'll prioritise it, so as soon as we can be confident it won't break then we will create a patch level release.

raphaChoquet:
Yes, tests has broken because program doesn't replace paths containing '../' in $path.
I think we need to replace realpath function by other logic to keep symbolic link.

robfrawley:
@raphaChoquet We've cover this in multiple issues prior: you need to set a proper `data_root` value in your bundle configuration to allow for the **real path** once the symbolic link is resolved.

antoligy:
Thanks for the clarification @robfrawley, I remember that being a particular pain-point in earlier issues.  I wasn't actually aware of the need to configure data_root either, so I wonder if it's worth us documenting that to avoid more issues of this nature?

[...]

PR #848, opened 2017-01-11
Fix absolute path

raphaChoquet:
There was conflict with 'assets:install --symlink', which cause the NotLoadableException. It searches in web but the realpath was returned the path in src/.../. That path was outside the defined root path (/web by default).

### DIFF
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -59,7 +59,7 @@ class FileSystemLoader implements LoaderInterface
      */
     public function find($path)
     {
-        if (!($absolutePath = realpath($this->rootPath.DIRECTORY_SEPARATOR.$path))) {
+        if (!($absolutePath = realpath($this->rootPath).DIRECTORY_SEPARATOR.$path)) {
             throw new NotLoadableException(sprintf('Source image not resolvable "%s"', $path));
         }
 


antoligy:
Thanks for contributing!

antoligy:
I'm fine with merging this in principle and releasing a patch level, especially as it should fix #846, but this has broken the tests introduced in #775 and I would also prefer to hear from @robfrowley first.

robfrawley:
I'm :-1: on this. Changing
```php
if (!($absolutePath = realpath($this->rootPath.DIRECTORY_SEPARATOR.$path))) {
```
to
```php
if (!($absolutePath = realpath($this->rootPath).DIRECTORY_SEPARATOR.$path)) {
```
Removes the ability to ensure the image path is within the root path. Again, you simply have to set a data root in the bundle configuration if you require files outside the default. Also, you might as well change the line to 
```php
if (!($absolutePath = $this->rootPath.DIRECTORY_SEPARATOR.$path)) {
```
and remove the call to `realpath` altogether (though this is a horrible idea), as the root path is already resolved at this point.

robfrawley:
Also, can you clarify what you intended to describe @raphaChoquet? The description you gave doesn't make sense:

> It searches in web but the realpath was returned the path in src/.../. That path was outside the defined root path (/web by default).

The `src/../` is what is *should return* as `web` lives within this path, once resolved.

raphaChoquet:
I wanted to say i have used the command assets:install with option -symlinks. Then  the realpath resolve 'bundles/app/img/my-image.png' by 'src/AppBundle/Resources/public/img/my-image.png'. This cause exception 'Source image invalid "%s" as it is outside of the defined root path'. I suppose we need to define a function to determinate if $path is in the root_path without destroying the symlinks. Perhaps you considerate symlink as a security fail ? 

Sorry for my english, i'm not have a good level ^^'

robfrawley:
@raphaChoquet Don't worry about your English; it's absolutely fine! ;-) I'll spend some time with this over the weekend to see if we can't allow for an array of data roots, **or** define an alternate realpath function that handles this more gracefully. I get where you are running into problems, but the issue with allowing symbolic links blindly is the ability for unintentionally following routes to other, possibly secure, paths on the entire filesystem. @antoligy Any ideas or guidance on which of the above possible solutions you think may be better to implement?

robfrawley:
@antoligy @davidromani @adaniloff @raphaChoquet @picks44 Alternatively to all other previously mentioned paths forward, perhaps we should create a `realpath` resolution toggle via a bundle configuration option? This would allow a user to disable it and ignore the security ramifications of blindly following symbolic links. It should be **off** by default, IMHO, that way people need to explicitly enable it (and understand the ramifications of doing so) before the security check is skipped. Something like:

```yml
liip_imagine:
  loaders:
    default:
      filesystem:
        resolve_sym_links: false
```

robfrawley:
See https://github.com/liip/LiipImagineBundle/issues/846#issuecomment-272563029 for a possible solution.

robfrawley:
Closing as this "fix" breaks unit tests. Please feel free to re-open a new PR if you can provide an implementation that doesn't break the requirements of this component. I suggest you see if #851 fixes your issue.

antoligy:
👍

PR #851, opened 2017-01-13
Allow multiple root paths for FileSystemLoader

### LINKED ISSUE #846: "Cache directory creation problem" and "source image could not be found" with v1.7

robfrawley:
Allow multiple data roots for file system loader as a possible approach to resolve #846. Includes `beforeNormalization` step in configuration tree-builder to allow prior string values for `data_root` bundle configuration to be seamlessly converted into an array. For example, both of the following are now valid configurations:
```yml
liip_imagine:
  loaders:
    default:
      filesystem:
        data_root: "/your/data/root"
```
```yml
liip_imagine:
  loaders:
    default:
      filesystem:
        data_root: 
          - "/your/data/root/a"
          - "/your/data/root/b"
```
This also adds a "feature" in that we can look in multiple locations using the same file system loader now, which I suspect could be beneficial for some use-cases. Most importantly, though, this implementation doesn't break any prior unit tests (but it *does* add a few unit tests to cover the new cases introduced).

### DIFF
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -32,26 +32,32 @@ class FileSystemLoader implements LoaderInterface
     /**
      * @var string
      */
-    protected $rootPath;
+    protected $dataRoots;
 
     /**
      * @param MimeTypeGuesserInterface  $mimeTypeGuesser
      * @param ExtensionGuesserInterface $extensionGuesser
-     * @param string                    $rootPath
+     * @param string[]                  $dataRoots
      */
     public function __construct(
         MimeTypeGuesserInterface $mimeTypeGuesser,
         ExtensionGuesserInterface $extensionGuesser,
-        $rootPath
+        $dataRoots
     ) {
         $this->mimeTypeGuesser = $mimeTypeGuesser;
         $this->extensionGuesser = $extensionGuesser;
 
-        if (empty($rootPath) || !($realRootPath = realpath($rootPath))) {
-            throw new InvalidArgumentException(sprintf('Root image path not resolvable "%s"', $rootPath));
-        }
+        $this->dataRoots = array_map(function ($root) {
+            if (!empty($root) && false !== $realRoot = realpath($root)) {
+                return $realRoot;
+            }
 
-        $this->rootPath = $realRootPath;
+            throw new InvalidArgumentException(sprintf('Root image path not resolvable "%s"', $root));
+        }, (array) $dataRoots);
+
+        if (count($this->dataRoots) === 0) {
+            throw new InvalidArgumentException('One or more data root paths must be specified.');
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


robfrawley:
Fixed PR to support PHP `5.3` (`$this` calling context isn't implemented with Closures in `5.3`).

robfrawley:
@antoligy @cedricziel How do you guys feel about this feature as both a ligament feature and a possible fix for those attempting to use multiple symbolic links outside a single path.

robfrawley:
Amended PR to add a test case that ensures `data_root` configuration value as a string is normalized to an array.

antoligy:
Nice, with that latest addition this should be fully backwards compatible with the latest release! 🙂 

I'm very happy with this approach, not least of all as it addresses the "real" issue, but also as it adds more thorough tests.
I still think that it is worth amending the documentation to better detail this behaviour, however that's something which we can address later.  For the meantime, no objections from me, but I will take a more in-depth look first before hitting approve.

robfrawley:
@antoligy I too like the format `/foo/bar:/bar/foo`, but no, I do not believe such is well-defined elsewhere. I do agree we should standardize on something. Perhaps we should investigate and see where similar usage might be appropriate. Also, docs PR will be incoming shortly that details this behavior, as well as provides some concrete information for commonly encountered errors.

robfrawley:
So, I'm pretty positive this doesn't result in any BC breaks: the configuration value for `data_root` continues to support a **scalar**, as well as the new type of **array of scalars**; the `FileSystemLoader` continues to support passing either a **scalar** or **array of scalars** for the data root. The only BC break is for those extending this class and relying on the protected `rootPath` member, which has been renamed to `rootPaths` (plural). Thoughts?

robfrawley:
So, @antoligy: we want to merge this?

antoligy:
Sorry for the delay, and yes I think we should be good to merge this!  Can we can warn about the potential BC breakage in the changelog? (I doubt that this bit of the API is actually used though, and I agree that in 2.0 we should reduce the surface area exposed by the API - As we're looking to decouple business logic from controllers the API definition would be greatly improved anyway!)

robfrawley:
Agreed (re decoupling and reduction of surface area). Ok; I'll merge this and release `1.7.1` later today, which will include an updated changelog. Are there any other pressing issues I'm not aware of that need to be part of this patch release?

antoligy:
Nothing as far as I'm aware

antoligy:
🎉

robfrawley:
Still need to cherry pick for master and 2.0 branches, but [1.7.1](https://github.com/liip/LiipImagineBundle/releases/tag/1.7.1) release is now live! ;-)

robfrawley:
Comment about the BC break for those extending `FileSystemLoader` is in the `UPGRADE.md` file in the respective line about the changes to that class: https://github.com/liip/LiipImagineBundle/blob/1.7.1/UPGRADE.md#17x

PR #853, opened 2017-01-15
[WIP] Composer Update Notice Hook

robfrawley:
This is a composer update/install hook that reads in the [UPGRADE.md](https://github.com/robfrawley/LiipImagineBundle/blob/feature-composer-upgrade-notice/UPGRADE.md) file, searches for a header matching the version Composer is installing, and, if such a version exists in the file, prints out the corresponding text for the user to acknowledge. See https://github.com/liip/LiipImagineBundle/issues/846#issuecomment-272584682 for the comments where the discussion for this component began.

Here is an example of the output

liip-imagine-composer-update-notice

Lastly, it is important to note that I cut off support for PHP 5.3; users using this release will simply not see any notices (it doesn't error out or anything, it is just turned off for them). This helps facilitate a sane, well-laid-out code base (PHP 5.3 doesn't properly implement closure scope).


robfrawley:
Nevermind; only root scripts are called.

Issue #874, opened 2017-02-09
[BUG] 1.7.2 One FileSystemLocator for several FileSystemLoader


sebastianblum:
| Q | A
| --- | ---
| Bug Report? | yes
| Feature Request? |no
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | 1.7.2

Hello,
with release 1.7.2 our implementation is broken.

We use several FileSystemLoaders to hide the directories we store our images and to shorten the image urls. We have these config:

```iml
liip_imagine:
  resolvers:
    default:
      web_path:
        web_root: "%kernel.root_dir%/../web"
        cache_prefix: media

  loaders:
    assets:
      filesystem:
        data_root: "%kernel.root_dir%/../web/assets/"
    default:
      filesystem:
        data_root: "%kernel.root_dir%/../web"
    guide:
      filesystem:
        data_root: "%kernel.root_dir%/../web/assets/guide"
    news:
      filesystem:
        data_root: "%kernel.root_dir%/../web/assets/news"
    press:
      filesystem:
        data_root: "%kernel.root_dir%/../web/assets/press"
```

With version 1.7.2, you introduced one FileSystemLoader service for all FileSystemLoader.

https://github.com/liip/LiipImagineBundle/blob/1.0/Binary/Locator/FileSystemLocator.php#L28
will be executed in our case 5 times with
 "%kernel.root_dir%/../web/assets/",
"%kernel.root_dir%/../web",
 "%kernel.root_dir%/../web/assets/guide",
"%kernel.root_dir%/../web/assets/news",
"%kernel.root_dir%/../web/assets/press"

and only the last configuration ("%kernel.root_dir%/../web/assets/press") is stores in $this->roots attribute.

In my opinion, we need one FileSystemLocator each FileSystemLoader.

robfrawley:
FYI: You can skip all of this and simply pass an array now; no need to use different loaders:
```yml
liip_imagine:
  loaders:
    all:
      filesystem:
        data_root: 
          - "%kernel.root_dir%/../web/assets/"
          - "%kernel.root_dir%/../web"
          - "%kernel.root_dir%/../web/assets/guide"
          - "%kernel.root_dir%/../web/assets/news"
          - "%kernel.root_dir%/../web/assets/press"
```
That said, this is definetly a bug that requires attention.

sebastianblum:
hello @robfrawley  thank you very much for your answer.

[...]

PR #875, opened 2017-02-12
[Loader] [Locator] FileSystemLocator service must not be shared

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #874
| License | MIT

This PR adds a compiler pass that gathers the `Locator` services and makes them __not shared__, causing the Symfony container to instantiate new objects each time they are requested (instead of the default behavior of re-using the first instance created).

This solves a bug introduced in `1.7.2` that causes those who use multiple `FileSystemLoader` definitions to only have their last path registered. Functional tests have been added to ensure a similar regression does not occur in the future.

Additionally, when the `*Locator` classes were implemented, exceptions originating from Symfony's `OptionsResolver` were ignored; this PR adds a try/catch block to the `FileSystemLocator::setOptions()` method and throws an exception originating from the `Liip` namespace so users don't have to worry about catching external dependencies' exceptions.

---

__Note:__ This PR should _not_ be merged into the `2.0` branch. The special handling seen in this PR is only required because the `1.0` branch supports both Symfony `<2.8` _and_ `>=2.8`. These versions require different mechanisms for achieving non-shared services (with `<2.8` requiring "prototype" scoping and `>=2.8` allowing the service to be simply defined as "not shared"). The `2.0` branch can achieve the same intent of this PR by simpy adding `shared="false"` to the locator service definitions.

```xml
<service shared="false" id="liip_imagine.binary.locator.filesystem" class="%liip_imagine.binary.locator.filesystem.class%" public="false"></service>
<service shared="false" id="liip_imagine.binary.locator.filesystem_insecure" class="%liip_imagine.binary.locator.filesystem_insecure.class%" public="false"></service>
```

### DIFF
--- a/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
+++ b/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
@@ -11,8 +11,10 @@
 
 namespace Liip\ImagineBundle\DependencyInjection\Factory\Loader;
 
+use Liip\ImagineBundle\Utility\Framework\SymfonyFramework;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Reference;
 
 class FileSystemLoaderFactory extends AbstractLoaderFactory
@@ -24,7 +26,7 @@ class FileSystemLoaderFactory extends AbstractLoaderFactory
     {
         $definition = $this->getChildLoaderDefinition();
         $definition->replaceArgument(2, $config['data_root']);
-        $definition->replaceArgument(3, new Reference(sprintf('liip_imagine.binary.locator.%s', $config['locator'])));
+        $definition->replaceArgument(3, $this->createLocatorReference($config['locator']));
 
         return $this->setTaggedLoaderDefinition($loaderName, $definition, $container);
     }
@@ -60,5 +62,19 @@ class FileSystemLoaderFactory extends AbstractLoaderFactory
                     ->end()
                 ->end()
             ->end();
+    }
+
+    /**
+     * @param string $reference
+     *
+     * @return Reference
+     */
+    private function createLocatorReference($reference)
+    {
+        return new Reference(
+            sprintf('liip_imagine.binary.locator.%s', $reference),
+            ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
+            SymfonyFramework::hasDefinitionSharedToggle()
+        );
     }
 }
--- a/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
+++ b/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
@@ -73,7 +73,7 @@ class FileSystemLoaderFactory extends AbstractLoaderFactory
     {
         $name = sprintf('liip_imagine.binary.locator.%s', $reference);
 
-        if (SymfonyFramework::hasDefinitionSharedToggle()) {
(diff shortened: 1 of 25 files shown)


DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
```diff
@@ -61,4 +63,18 @@ public function addConfiguration(ArrayNodeDefinition $builder)
...
+     * @param string $reference
+     *
+     * @return Reference
+     */
+    private function createLocatorReference($reference)
+    {
+        return new Reference(
+            sprintf('liip_imagine.binary.locator.%s', $reference),
+            ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
+            SymfonyFramework::hasDefinitionSharedToggle()
```
sebastianblum: The third argument (strict) was removed in symfony 3.0.
Should we use an switch like

```php
if (symfony 2.x) {
    return new Reference(
        sprintf('liip_imagine.binary.locator.%s', $reference),
            ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE,
            false
    );

return new Reference(sprintf('liip_imagine.binary.locator.%s', $config['locator']);
}
```

robfrawley: The reason I left it like this is because that will always be true for Symfony `>=2.8`, and therefore it doesn't matter that it's ignored. We only care about it when it is `false` on Symfony `<2.8`. Otherwise, the extra parameter can be, and is, safely ignored (the PHP interpreter allows and ignores any extra parameters). Do you really think it's necessary to add a switch?

sebastianblum: Good morning @robfrawley, not an easy question. You are right, the code is fine.
I would prefer a feature switch at the moment and remove it in future (maybe 2.0) when at least symfony 2.8 will be supported.

robfrawley: Ok; ammended the code to the following:
```php
$name = sprintf('liip_imagine.binary.locator.%s', $reference);

if (SymfonyFramework::hasDefinitionSharedToggle()) {
    return new Reference($name);
}

return new Reference($name, ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE, false);
```

I didn't use a version toggle, as I want to use the exact mechanism that requires a non-strict reference, which [can be found here](https://github.com/liip/LiipImagineBundle/pull/875/files#diff-761cf6b460d040d983ac2342a13ff0c5R35).

noniagriconomie:
Hi @robfrawley thank you for this PR that actually fix our problem too

Do you think it is going to be merged soon?

Many thanks :)

robfrawley:
@noniagriconomie It's excellent to hear that this PR resolves any issues you were encountering in your project! With regard to the rollout of this bugfix, I plan to merge this PR into `master` by late tomorrow (EST, UTC -5). I intend to give the code a final review and perform any additional cleanup starting in the afternoon, after which you can immediately require the mainline release by using the `1.0.x-dev` version string in your project's `composer.json` file.

As for the next stable version to include this fix (`1.7.3`), I am shooting for a public release over the coming weekend. There a two additional PRs (which I will be posting shortly) that I would like to include in the upcoming release; those are the main reason for the short delay. _At the latest_ you will be able to benefit from this fix (using a stable version requirement like `~1.7`) starting on _Monday 2017-02-27_.

@antoligy Thanks for allotting the time to provide feedback; it always feels wrong to merge PRs without some form of peer-review. :-) As always, much appreciated.

noniagriconomie:
fine, waiting the publication of v1.7.3 then :)

Issue #881, opened 2017-02-22
Public resources are denied by the default FileSystemLoader


bobvandevijver:
| Q | A
| --- | ---
| Bug Report? | yes
| Feature Request? | yes
| BC Break Report? | yes
| Imagine Bundle Version | 1.7.2

With the release of the 1.7.2, the FileSystemLoader no longer accepts images which are not placed in the "save" data roots. However, this breaks the logic of the public resources in your src directory, if they are symlinked in the web directory. 

This asset symlinking is a commonly used option with Symfony, and therefor this should not be considered unsafe. I know it is possible to add the dataroot of these `src` resources to the configuration of this bundle, but in my opinion, that should not be necessary.

In conclusion: images installed in the `web/bundles` folder which are symlinked to the `src` bundles should not be considered unsafe.

robfrawley:
@bobvandevijver You can either configure your data roots explicitly (yes, this does require you specify all resolved symbolic link paths) or you can switch to the old resolver algorithm.

The default moving forward is to consider symbolic links unsafe. While the usage you are describing is safe, how would you expect we determine that *all* symbolic links are safe? We can't: they could point anywhere on the file system and they bypass the intent of the `data_root` value, which is to constrain your files to the configured director[y|ies].

As for automatically detecting "safe" paths, there is no "one" way to setup a Symfony application, and there is no dependency on a `src` directory, simply a common convention. I don't find it unnecessarily complicated to define the paths *you* consider safe, versus making arbitrary decisions for the user.

```yml
# provide an array of paths to use as data roots
liip_imagine:
  loaders:
    default:
      filesystem:
        data_root:
          - "%kernel.root_dir%/../web"
          - "%kernel.root_dir%/../src/AppBundle/Resources/public"
```
```yml
# use the current locator algorithm (default)
liip_imagine:
  loaders:
    default:
      filesystem:
        locator: filesystem
```
```yml
# use the old locator algorithm (pre 1.7.x)
liip_imagine:
  loaders:
    default:
      filesystem:
        locator: filesystem_insecure
```

bobvandevijver:
I do not want to consider all symbolic links safe, only the symbolic links that point to bundles which have a `Resources/public`-folder, which is, considering the Symfony standard, safe for web usage.

[...]

PR #883, opened 2017-02-23
[Loader] Add bundle resources to safe path when requested

bobvandevijver:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | 
| Fixed tickets | #881
| License | MIT
| Doc PR | Not yet, will be made after approval

This change adds the functionality to add bundle public resource paths automatically when the `bundle_resources` option is set. See #881 for the discussion on this.

### DIFF
--- a/Binary/Locator/FileSystemInsecureLocator.php
+++ b/Binary/Locator/FileSystemInsecureLocator.php
@@ -21,10 +21,12 @@ class FileSystemInsecureLocator extends FileSystemLocator
      */
     protected function generateAbsolutePath($root, $path)
     {
-        if (false !== strpos($path, '..'.DIRECTORY_SEPARATOR)) {
+        if (false !== strpos($path, '..'.DIRECTORY_SEPARATOR) ||
+            false !== strpos($path, DIRECTORY_SEPARATOR.'..') ||
+            false === file_exists($absolute = $root.DIRECTORY_SEPARATOR.$path)) {
             return false;
         }
 
-        return file_exists($absolute = $root.DIRECTORY_SEPARATOR.$path) ? $absolute : false;
+        return $absolute;
     }
 }
--- a/Binary/Locator/FileSystemLocator.php
+++ b/Binary/Locator/FileSystemLocator.php
@@ -49,14 +49,51 @@ class FileSystemLocator implements LocatorInterface
      */
     public function locate($path)
     {
-        foreach ($this->roots as $root) {
-            if (false !== $absolute = $this->generateAbsolutePath($root, $path)) {
-                return $this->sanitizeAbsolutePath($absolute);
-            }
+        if (false !== $absolute = $this->locateUsingRootPlaceholder($path)) {
+            return $this->sanitizeAbsolutePath($absolute);
+        }
+
+        if (false !== $absolute = $this->locateUsingRootPathsSearch($path)) {
+            return $this->sanitizeAbsolutePath($absolute);
         }
 
         throw new NotLoadableException(sprintf('Source image not resolvable "%s" in root path(s) "%s"',
             $path, implode(':', $this->roots)));
+    }
+
+    /**
+     * @param string $path
+     *
+     * @return bool|string
+     */
+    private function locateUsingRootPathsSearch($path)
+    {
+        foreach ($this->roots as $root) {
+            if (false !== $absolute = $this->generateAbsolutePath($root, $path)) {
+                return $absolute;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param string $path
+     *
+     * @return bool|string
+     */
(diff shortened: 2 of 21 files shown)


robfrawley:
This looks good. I'm going to clean up a few things and add some additional functionality (allow the user to whitelist and/or blacklist bundles), but otherwise I'm not against adding the *option* to enable this. We definetly need tests and docs, though I assume you were waiting on the "okay" before providing them.

Also, I thought you said you were going to implement a means to differentiate between bundles using a bundle prefix or something? With this setup, all paths are loadable, so given the following files

```
./src/FooBundle/Resources/public/foobar.jpg
./src/BarBundle/Resources/public/foobar.jpg
```

The user could request the file `foobar.jpg` outright, and since both `Resources/public` paths are registered, the first will be loaded. Now, I know generally people will load them using the symlinked path of `bundles/[foo|bar]/foobar.jpg` but it would be nice to implement some logic to allow the user to simply request `@FooBundle:foobar.jpg` or `@BarBundle:foobar.jpg`, no? What do you think about adding something like that to the mix?

Doing so shouldn't be especially difficult, as we can register the data paths array as an associative array of `bundle-name => bundle-public-path`, and then using the index to find the correct root when the path matches this format. What do you think?

robfrawley:
FYI, the `1.7.3` release is scheduled for this coming Monday, February 27 2017, so be sure to get the tests and docs implemented prior if you want this included in that release. Otherwise, the next release will likely be anywhere between 3 and 6 weeks from then.

bobvandevijver:
I've added some doc and updated the test to complete correctly. I also duplicated the failing test with the `bundle_resources` parameter set to `true`, but I can not check anything on the output besides having the original data root, as there are no `Resources/public` folders available during the test. Maybe you have a good suggestion for the extra test?

Concerning the differentiation on bundle prefix: I believe I did not say that I would implement it, although it should indeed not be to hard. However, I believe it should be a feature for another PR, and I even doubt that kind of logic should be incorporated in this bundle.

If I need to do anything for this to be merged, please let me know 👍

robfrawley:
@bobvandevijver Can you test the updated code in live projects and report back that it operates as expected? The following changes have been implemented since you initial commit:
- Squashed your commits into a single one
- Updated to work properly on all the versions of Symfony we currently support
- Updated to use non-reflection method of bundle name/path resolution, if available
- Updated to provide blacklisting/whitelisting support for bundle auto-registration
- Updated to support specific root path placeholders via syntax `@root-name:the/path/to/file.ext`
- Updated to have functional tests and unit tests with 100% coverage
- Updated docs to describe newer configuration syntax

bobvandevijver:
@robfrawley Just checked it in my project, works like a charm :+1: I like the whitelist/blacklist feature by the way! Also, didn't know about the `kernel.bundles_metadata`, don't know why I did not find that one when I was looking for the information (and yes, I have it available)...

Looks ready to merge to me 👍

robfrawley:
@bobvandevijver The `kernel.bundles_metadata` parameter seems hardly documented at all, short of [one issue](https://github.com/symfony/symfony/issues/21646) in the Symfony repo and the source code for the [`HttpKernel\Kernel` class](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Kernel.php#L543), I so not believe there is any official documentation whatsoever ... don't feel bad about not finding it: I stumbled upon it out of sheer luck. ;-) Glad to hear this PR still works as expected for you! It will be part of the `1.7.3` release, scheduled for tagging later today.

robfrawley:
@bobvandevijver By the way, thanks for pushing this and taking the time to explain your idea in detail during my stubborn conversation with you in #881. This turned out to be a solid feature, despite my initial pessimism.

robfrawley:
Feature merged and available immediately via our `dev-master` branch alias `1.0.x-dev` - will be included in the upcoming `1.7.3` release, which should be tagged within the next six hours or so.

bobvandevijver:
@robfrawley No problem, I'm also known to be stubborn :-) And it all worked out for the best 👍

tobias-93:
@robfrawley Could you tag the release please?

robfrawley:
@tobias-93 Yeah, it'll be tagged today. I was trying to get a few additional PRs merged, but due to the [Amazon S3 failures](https://techcrunch.com/2017/02/28/amazon-aws-s3-outage-is-breaking-things-for-a-lot-of-websites-and-apps/) that directly affected Travis CI (and many other websites), jobs were queued forever and failing for no reason. Things have cleared up now and all PRs are merged for this release.

robfrawley:
@tobias-93 It's out.

Issue #892, opened 2017-03-02
 Changes in Aws Definition


plozmun:
| Q | A
| --- | ---
| Bug Report? | yes
| Feature Request? | no
| BC Break Report? | yes
| RFC? | yes
| Imagine Bundle Version | 1.7.3

When update at this version i got error:

`Cannot dump definition because of invalid class name ('\\Aws\\S3\\S3Client') `
Can be related to this [Changes in Aws Definition](https://github.com/liip/LiipImagineBundle/compare/1.7.2...1.7.3#diff-5584d98c341355797ea40fde49e441b1L28) ?

Chrysweel:
+1

cedricziel:
Please comment on #893 whether this fixes it for you. Thank you!

PR #893, opened 2017-03-02
Revert adding leading slash to S3 class names

cedricziel:
Closes: #892

| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | /no
| Deprecations? | no
| Tests pass? | yes/no/maybe
| Fixed tickets | #892 
| License | MIT

The service definition for the S3 resolver was changed slightly and instantiation broke for others. This change reverts it.

### DIFF
--- a/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
+++ b/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
@@ -25,12 +25,12 @@ class AwsS3ResolverFactory implements ResolverFactoryInterface
     public function create(ContainerBuilder $container, $resolverName, array $config)
     {
         $awsS3ClientId = 'liip_imagine.cache.resolver.'.$resolverName.'.client';
-        $awsS3ClientDefinition = new Definition('\Aws\S3\S3Client');
+        $awsS3ClientDefinition = new Definition('Aws\S3\S3Client');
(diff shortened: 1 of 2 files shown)


cedricziel:
Ping @Chrysweel , @plozmun can you test the bugfix?

rvanlaarhoven:
Thanks, this seems to resolve it

Chrysweel:
Yes it resolve the problem.

But my question is .. why a commit added the / in definition class previously ?

cedricziel:
Well, that's how bugs get introduced :)

Those kinds of refactorings will be much more stable when we raise the php version requirement and thus can use class constants.

robfrawley:
Sorry guys, those got caught in a regex fixing FQCN in tests (which fixes some IDEs).

robfrawley:
@cedricziel Is there anything we can do test-wise to ensure this regression is never introduced in the future?

cedricziel:
@robfrawley hmm. I don't think it's worth the hassle. - we should much rather move on and rely on PHP for those in the future :)

Don't worry about it.

cedricziel:
Ahh, @robfrawley - we need a bugfix release. Can you create one?

robfrawley:
Yeah, on it.

cedricziel:
Awesome! :tada:

robfrawley:
Live!

PR #908, opened 2017-04-13
[WIP] Remove deprecated and legacy code, migrate some syntax, cleanup

robfrawley:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

This PR removes all obviously marked deprecations and legacy code/tests from the `1.x` branch for the `2.x` branch, removing backward compatibility layers for older Symfony versions, and moves from legacy syntaxes to more modern solutions, such as from the long to short array syntax.

## Requirements

- Symfony `>= 2.8`
- PHP `>= 7.0`

## Todo Still

- [x] Change to short array syntax
- [x] Remove obviously marked deprecations
- [x] Remove obviously marked legacy code/text
- [x] Remove unobvious deprecations
- [x] Remove unobvious legacy code/texts
- [x] Update `UPGRADE.md` with all changes
- [ ] Update any relevant documentation

## Should we Go Further

~~Do we want to bump one more version for PHP and require PHP `7.0`, which would allow for us to really clean up the code with function types explicitly mapped everywhere. IDK if it's worth it, but it's a thought.~~

*Thanks to some comments, previous discussions, and my own thoughts, ATM this PR does go a step further to requiring PHP `7.x`.*

The only consideration we need to make, with a move to `7.x`, is moving HHVM to an "allowed failure" for the time being. This is required due to HHVM's incorrect handling of language-level type checking, as well as globally-enabled function type checking, (in stark contrast to how PHP handled this), which breaks our code and some dependencies when running HHVM in PHP 7 mode. See the following issues for reference:
- https://github.com/composer/composer/issues/5978
- https://github.com/composer/composer/issues/6230
- https://github.com/composer/composer/issues/6134
- https://github.com/facebook/hhvm/issues/7198
- [and more](https://github.com/facebook/hhvm/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20php7)... 

I've worked around this a bit by explicitly disabling the "type checking" feature of HHVM via the INI setting, but other errors still persist in PHP7-mode, so IDK.


robfrawley:
Ooh yeah; Travis test time down from 13-15 minutes to 4 minutes!

antoligy:
There's a lot to this diff so it'll take me a while to go through in detail, but most immediately looking through how much more ... _readable_ the testing configuration now is ... _thank you so much!_

rpkamp:
Seeing how PHP 5.6 is officially end of life, I don't see a reason to support it anymore in 2.0.

Also Symfony 4 will also require PHP 7.0 (https://medium.com/@fabpot/fabien-potencier-4574622d6a7e) and Monolog 2.x will also require PHP 7.0 (https://github.com/Seldaek/monolog/blob/master/composer.json)

DependencyInjection/LiipImagineExtension.php
```diff
@@ -25,12 +25,12 @@ class LiipImagineExtension extends Extension
     /**
      * @var ResolverFactoryInterface[]
      */
-    protected $resolversFactories = array();
+    private $resolversFactories = array();
```
rpkamp: `array()` could be replaced with `[]`

DependencyInjection/LiipImagineExtension.php
```diff
@@ -25,12 +25,12 @@ class LiipImagineExtension extends Extension
...
 
     /**
      * @var LoaderFactoryInterface[]
      */
-    protected $loadersFactories = array();
+    private $loadersFactories = array();
```
rpkamp: `array()` could be replaced with `[]`

Templating/FilterExtension.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->helper = $helper;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFilters()
+    {
+        return [
+            new \Twig_SimpleFilter('imagine_filter', array($this->helper, 'filter')),
```
rpkamp: can be replaced with `[$this->helper, 'filter']`

Templating/FilterExtension.php
```diff
@@ -0,0 +1,48 @@
...
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
```
rpkamp: `Twig_Extension#getName` is deprecated and should be removed.

See https://github.com/twigphp/Twig/blob/2.x/CHANGELOG#L153

robfrawley:
@rpkamp Thanks for the quick review. I also tend to agree with your assessment as it applies to support (or dropping it) for PHP `5.6`.

robfrawley:
@antoligy @cedricziel Any change you can carve out some time to review this mega-PR? Thanks guys. ;-)

robfrawley:
Rebased to fix merge conflicts introduced by #914 (that PR removed all class name parameters from the service definition file). 

@antoligy @cedricziel Aside from updating the `UPGRADE.md` and `CHANGELOG.md` files, this is ready to go. Things to consider before finalizing this PR:

- __Are we comfortable making the `2.x` branch PHP `7.x` only?__ Benefits include a vastly simplified Travis test definition file, as well as allowing us to introduce function type hints (which, IMHO, help with usage, debugging, and provide a great firewall between an improper use of a method), as well as some other language constructs. Moreover, as was already pointed out, the next major Symfony version will require PHP `7.x` and I foresee releasing `2.x` around the same timeframe (Novemberish). Of course, the `1.x` branch will remain as is for those who need to use PHP `5.6`, and I intend to continue to accept PRs and issues against it for some time, so it's not like people won't be able to use this bundle with older PHP versions...thoughts?

- __The Mongo functionality was removed due to it no longer working__ (the same extension isn't going to be made available for PHP `7.x`); do we want to rewrite it instead of removing it (for the new extension)?

Aside from getting your opinion on the above items, I'd really appreciate it if you guys could take some time to review some of the changes introduced by the PR. I have other PRs that will start architecturally changing this bundle on the `2.x` branch, but I don't want to add those until we've got this completed to avoid merge conflicts.

Binary/Loader/FileSystemLoader.php
```diff
@@ -38,30 +38,19 @@ class FileSystemLoader implements LoaderInterface
...
         $this->extensionGuesser = $extensionGuesser;
-
-        if (count($dataRoots) === 0) {
-            throw new InvalidArgumentException('One or more data root paths must be specified.');
-        }
-
-        if (func_num_args() >= 4 && false === ($this->locator = func_get_arg(3)) instanceof LocatorInterface) {
-            throw new \InvalidArgumentException(sprintf('Method %s() expects a LocatorInterface for the forth argument.', __METHOD__));
-        } elseif (func_num_args() < 4) {
-            @trigger_error(sprintf('Method %s() will have a forth `LocatorInterface $locator` argument in version 2.0. Not defining it is deprecated since version 1.7.2', __METHOD__), E_USER_DEPRECATED);
```
robfrawley: *Note to self:* This deprecation notice needs to be amended to properly articulate the changes introduced in the `2.x`.

Koc:
+1 for PHP 7.x or even 7.1 for nullable return types

Binary/Loader/GridFSLoader.php
```diff
@@ -1,54 +0,0 @@
...
- * For the full copyright and license information, please view the LICENSE.md
- * file that was distributed with this source code.
- */
-
-namespace Liip\ImagineBundle\Binary\Loader;
-
-use Doctrine\ODM\MongoDB\DocumentManager;
-use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
-
-class GridFSLoader implements LoaderInterface
```
robfrawley: *Note:* Removed due to the deprecated PHP `5.6` [`mongo`](https://pecl.php.net/package/mongo) extension that will __not__ be ported to PHP `7.x`. Do we want to re-write this for the new [`mongodb`](https://pecl.php.net/package/mongodb) extension instead of removing it ([documentation reference](http://php.net/manual/en/set.mongodb.php))?

Binary/Loader/FileSystemLoader.php
```diff
@@ -38,30 +38,19 @@ class FileSystemLoader implements LoaderInterface
...
-            $this->locator = new FileSystemLocator();
-        }
-
-        $this->locator->setOptions(array('roots' => (array) $dataRoots));
+        $this->locator = $locator;
+        $this->locator->setOptions(['roots' => $rootPaths]);
```
rpkamp: Why pass the `$rootPaths` to this class anyway since it's only handing them off to the passed in `LocatorInterface`? Why not set the option on the `LocatorInterface` _before_ passing it to this class?

robfrawley: That is a valid question and something I've considered changing, but the current configuration handling of these items is tied to the `*Loader` implementations and not the `*Locator` classes; the `FileLocator` and `InsecureFileSystem` "locator" variants are really implementation details ATM.

I'm not really positive how we should handle this moving forward and am very open to its discussion, but I do think that it is more appropriate to handle that in a PR of its own, no?

robfrawley: IDK: maybe it should be fully handled in this PR; what do you think?

DependencyInjection/Factory/Resolver/WebPathResolverFactory.php
```diff
@@ -13,24 +13,22 @@
...
         $resolverDefinition->replaceArgument(3, $config['cache_prefix']);
-        $resolverDefinition->addTag('liip_imagine.cache.resolver', array(
+        $resolverDefinition->addTag('liip_imagine.cache.resolver', [
             'resolver' => $resolverName,
-        ));
-        $resolverId = 'liip_imagine.cache.resolver.'.$resolverName;
+        ]);
 
-        $container->setDefinition($resolverId, $resolverDefinition);
+        $container->setDefinition($resolverId = 'liip_imagine.cache.resolver.'.$resolverName, $resolverDefinition);
```
rpkamp: I would not make this change, it makes the code harder to read.

DependencyInjection/LiipImagineExtension.php
```diff
@@ -57,7 +57,7 @@ public function getConfiguration(array $config, ContainerBuilder $container)
     }
 
     /**
-     * @see Symfony\Component\DependencyInjection\Extension.ExtensionInterface::load()
+     * @see \Symfony\Component\DependencyInjection\Extension.ExtensionInterface::load()
```
rpkamp: Why add the leading backslash here?

DependencyInjection/LiipImagineExtension.php
```diff
@@ -92,59 +86,31 @@ public function load(array $configs, ContainerBuilder $container)
...
      */
-    protected function loadResolvers(array $config, ContainerBuilder $container)
+    private function loadResolvers(array $config, ContainerBuilder $container)
     {
         foreach ($config as $resolverName => $resolverConfig) {
-            $factoryName = key($resolverConfig);
-            $factory = $this->resolversFactories[$factoryName];
-
-            $factory->create($container, $resolverName, $resolverConfig[$factoryName]);
+            $this->resolversFactories[$factoryName = key($resolverConfig)]->create($container, $resolverName, $resolverConfig[$factoryName]);
```
rpkamp: assigning and using a variable in the same line makes it harder to read.

DependencyInjection/LiipImagineExtension.php
```diff
@@ -92,59 +86,31 @@ public function load(array $configs, ContainerBuilder $container)
...
-            $definition = $container->getDefinition($service);
-            if (method_exists($definition, 'setFactory')) {
-                // to be inlined in imagine.xml when dependency on Symfony DependencyInjection is bumped to 2.6
-                $definition->setFactory($factory);
-            } else {
-                // to be removed when dependency on Symfony DependencyInjection is bumped to 2.6
-                $definition->setFactoryClass($factory[0]);
-                $definition->setFactoryMethod($factory[1]);
-            }
+            $this->loadersFactories[$factoryName = key($loaderConfig)]->create($container, $loaderName, $loaderConfig[$factoryName]);
```
rpkamp: assigning and using a variable in the same line makes it harder to read.

DependencyInjection/LiipImagineExtension.php
robfrawley: This is being migrated everywhere (slowly, they are everywhere) as many IDEs try to resolve these as within the *current namespace* when they are not declared as a root FQCN, instead of how they should be resolved, as true FQCN.

DependencyInjection/Factory/Resolver/WebPathResolverFactory.php
robfrawley: We do this often in the codebase now and can be found all over the place. While I disagree it makes it harder to read, I recognize many people feel this way. We should likely adopt a `CONTRIBUTING.md` file that defines some coding standards that the community agrees on as a whole.

Resources/config/imagine.xml
```diff
@@ -84,12 +84,12 @@
 
         <!--  Templating helpers and extensions -->
 
-        <service id="liip_imagine.twig.extension" class="Liip\ImagineBundle\Templating\ImagineExtension" public="false">
+        <service id="liip_imagine.templating.filter_extension" class="Liip\ImagineBundle\Templating\FilterExtension" public="false">
             <tag name="twig.extension" />
-            <argument type="service" id="liip_imagine.cache.manager" />
+            <argument type="service" id="liip_imagine.templating.filter_helper" />
```
rpkamp: Why couple the twig helper to the templating helper? I realise the mostly have the same code, but making one dependent on the other feels off. If there is an abstraction here I would say that both the twig extension and the templating helper should depend on a shared service.

DependencyInjection/Factory/Resolver/WebPathResolverFactory.php
robfrawley: Updated

DependencyInjection/LiipImagineExtension.php
robfrawley: Amended code to negate this issue

DependencyInjection/LiipImagineExtension.php
robfrawley: Amended code to negate this issue

Resources/config/imagine.xml
robfrawley: The intention was to remove code duplication; now that we've bumped the PHP requirements, though, traits serve the same purpose, so this has been rectified in that was, allowing both the helper and extension to again consume the `CacheManager`.

robfrawley:
@rpkamp Thanks for the review!

Binary/Loader/FileSystemLoader.php
rpkamp: Since you're making a signature change here I would solve the problem fully in this PR, instead of in another PR which breaks the signature a second time. People aren't supposed to be using 2.x yet I guess, but you never know :smiley:

DependencyInjection/Factory/ChildDefinitionTrait.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\DependencyInjection\DefinitionDecorator;
+
+trait ChildDefinitionTrait
+{
+    /**
+     * @return ChildDefinition|DefinitionDecorator
+     */
+    private function getChildDefinition($parent)
+    {
+        return class_exists('\Symfony\Component\DependencyInjection\ChildDefinition') ?
```
Koc: you can use `ChildDefinition::class` here

rpkamp: Actually you can't, because if the class doesn't exist (in case of older Symfony versions) that can't be resolved.

Koc: it is safe for non-existent classes

Koc: http://blog.kevingomez.fr/til/2016/01/15/php-class-keyword/

robfrawley: Updated all FQCN strings to the special `::class` compile-time constant: https://github.com/liip/LiipImagineBundle/pull/908/commits/ee9c83473de7c4ab57e25fc6b661eb5eb86a0171. Thanks for the recommendation.

robfrawley:
All the changes introduced by the pull request, as well as #914, have been added to the `UPGRADE.md` file in commit https://github.com/liip/LiipImagineBundle/pull/908/commits/f04ffa38b306679a4d38ad93afd9c72b396209c8. @antoligy @cedricziel Any comments before this is merged?

DependencyInjection/LiipImagineExtension.php
antoligy: Excellent point, and there's more on how php docblocks FQCNs are supported here: https://phpdoc.org/docs/latest/guides/types.html#class-name

antoligy:
On the topic of PHP 7, the primary argument to keeping our requirements lower were to match up with Symfony 2.8.

Since we've handled deprecations in 2.8 in the 1.0 branch anyway, now that we're raising to PHP 7 I'm honestly wondering if we ought to bump to Symfony 3.0?

Utility/Framework/SymfonyFramework.php
```diff
@@ -18,24 +18,6 @@ class SymfonyFramework
     /**
      * @return bool
      */
-    public static function hasDefinitionSharing()
```
antoligy: Not that anybody should have bound to this (or has had the time), but, should we deprecate this instead of flat-out removing?

Either way I think we ought to reference flag removal in the UPGRADE doc.

Tests/Imagine/Cache/SignerTest.php
```diff
@@ -21,9 +22,9 @@ class SignerTest extends AbstractTest
 {
     public function testImplementsSignerInterface()
```
antoligy: This comment is not intended for action/review...

Now that I look at it, this is a very strange thing to test for!  We typehint on SignerInterface, don't we?

antoligy: (And all the others like it)

Binary/Loader/GridFSLoader.php
antoligy: I think so, but I'm not sure that this should hold back this PR.  Let's add to the roadmap for 2.0?

antoligy:
Thanks for taking the trouble to go through and assemble this PR @robfrawley, this is a hefty chunk of the 2.0 roadmap and I cannot emphasise enough how useful all this is.

I've gone through with a more finely toothed comb, and all seems well, save for a few trivial things mostly housekeeping related...

This to me raises a few points to consider, potentially separately to this PR (because I would very much like to see it merged!):

- Do we know, or have a way of telling, who uses the adapters that we provide?  I am thinking primarily of GridFS and wondering if the replacement/alternative (should one even be required, my bias is that I do not yet use it!) could be potentially handled by a separate library.
- Similarly, do we know if anybody uses this library running in a Hack environment?  _Have we just broken it with this suite of changes? (looks unlikely, given your reports on the testing setup and the linked issues)_
- Smaller, and more frequent PRs for the future?  I understand that a lot of this was similar-ish and had to be within the same PR, but anything we can do to reduce the size of individual diffs is a win!

There's a trend of metrics here, unfortunately I haven't been able to find any useful information in Packagist...  Possibly something to raise in a further issue, because 2.0 gives us the chance to trial feature apoptosis.

Tests/Imagine/Cache/SignerTest.php
robfrawley: I always thought these tests were odd as well; :+1: remove them in another PR.

Utility/Framework/SymfonyFramework.php
robfrawley: Looks like I missed this in the `UPGRADE.md` file; it should absolutely be mentioned. As for removal vs depreciation, I believe we should take advantage of not having a `2.x` release yet and remove it.

Without a `2.x` release, those running this development branch should know the API isn't stable. The moment we have an initial `2.0.0` release, this changes and we absolutely need to deprecate, but as of now, we should take the ability to remove outright and do so over depreciation, IMHO.

antoligy: Definitely agreed about the 2.x API being unstable so it's ok to break atm until we are settled, we'd also be doing this by killing off the GridFS driver. 🙂 
Happy to brush this one up then!

robfrawley:
@antoligy I'm comfortable with Symfony `3.x` as our `2.x` base requirement.

robfrawley:
@antoligy 

> Do we know, or have a way of telling, who uses the adapters that we provide? I am thinking primarily of GridFS and wondering if the replacement/alternative (should one even be required, my bias is that I do not yet use it!) could be potentially handled by a separate library.

We don't and have no way of knowing. I don't use it either. I assume we'll have a PR shortly to add an adapter for the new [`mongodb`](https://pecl.php.net/package/mongodb) extension, but as I don't use it I'm not particularly qualified to create the adapter myself.

> Similarly, do we know if anybody uses this library running in a Hack environment? Have we just broken it with this suite of changes? (looks unlikely, given your reports on the testing setup and the linked issues)

Again, I don't know how we'd get access to any statistics. That said, the HHVM (in PHP 7 mode) issues extend outside this bundle to Composer and Symfony components, as well, so I do not believe anyone _can_ relying on this bundle under those circumstances. Once HHVM fixes these core language issues, I'll be sure to enable support in Travis again, though it seems like many projects are leaving HHVM support now that PHP 7 was released and offers similar performance improvements. Sadly, __the current state of HHVM makes it impossible to support PHP 7 and HHVM at the same time.__

> Smaller, and more frequent PRs for the future? I understand that a lot of this was similar-ish and had to be within the same PR, but anything we can do to reduce the size of individual diffs is a win!

Agreed; this blew up quick in size due to how many changes affected many files, but generally I agree with you here.

> Since we've handled deprecations in 2.8 in the 1.0 branch anyway, now that we're raising to PHP 7 I'm honestly wondering if we ought to bump to Symfony 3.0?

I'm comfortable with Symfony 3.x as our 2.x base requirement.

robfrawley:
I plan to merge this pull request at the end of today, pending any objections from @antoligy, @cedricziel, @rpkamp or @Koc. Last chance to make additional changes!

Binary/Loader/GridFSLoader.php
rpkamp: Flysystem also has support for GridFS (https://flysystem.thephpleague.com/adapter/gridfs/), so people who rely on GridFS could switch to that as well if need be.

rpkamp:
What about https://github.com/liip/LiipImagineBundle/pull/908/files#diff-af283426a4b2643990ccbcbae0d7555fR53? Will you fix it in this PR or should it be fixed later? Either way is fine by me :smiley:

robfrawley:
@rpkamp Again, there is no easy way to move the root path assignment to the `Locator` instances instead of the `Loader` instances; it involves likely changing the configuration layout to accommodate it. 

For example, we can't just assign the roots to a parameter during the `LocatorFactory` configuration

```php
$container->setParameter('liip_imagine.data_roots', $this->resolveDataRoots($config['data_root'], $config['bundle_resources'], $container));
```

And change the `Locator` service definitions to

```yaml
<service id="liip_imagine.binary.locator.filesystem" class="Liip\ImagineBundle\Binary\Locator\FileSystemLocator" public="false" shared="false">
            <tag name="liip_imagine.binary.locator" shared="false" />
            <call method="setOptions">
                <argument type="collection">
                    <argument key="roots">%liip_imagine.data_roots%</argument>
                </argument>
            </call>
        </service>

        <service id="liip_imagine.binary.locator.filesystem_insecure" class="Liip\ImagineBundle\Binary\Locator\FileSystemInsecureLocator" public="false" shared="false">
            <tag name="liip_imagine.binary.locator" shared="false" />
            <call method="setOptions">
                <argument type="collection">
                    <argument key="roots">%liip_imagine.data_roots%</argument>
                </argument>
            </call>
        </service>
```

Because doing so breaks the ability to have multiple, differently configured `Loader` instances. I'm not positive how to solve this issue, and it will only make this huge pull request even larger as solving the issue will absolutely span many files.

Can you take a look and submit a pull request; the three tries I made to solving the issue did not result in a functional fix. :-(

Binary/Loader/FileSystemLoader.php
robfrawley: I have not found a straightforward solution to this, and based on my testing the solution will likely be quite large (only making this pull request even larger), so I don't forsee this issue being resolved by this pull request. See https://github.com/liip/LiipImagineBundle/pull/908#issuecomment-299718198 for additional details.

As for breaking changes (multiple times), I'm not too concerned about that; the whole point of having a development branch is to be able to _actively develop on it_. Those running the `1.x` branch should expect no backward incompatible changes, but that doesn't apply here. I run the `2.0` branch on a few personal projects, and while it's annoying to update it when breaking changes occur, that is a contract I agreed to when I decided to run an unreleased version. We're going to need to introduce a plethora of additional breaking changes to complete many of the TODO items for `2.0`, so let's focus on making the release amazing and not on making it easier for those running unreleased software. Right?

rpkamp:
I will have a look once this PR is fixed, no problem

robfrawley:
Another thing to note is that the `Locators` were meant to be an internal abstraction to allow for some edge cases, and are not intended to be used directly by anyone. In the case of the `Loaders`, in all instances, they are passed the values required to handle the job. Does it really make sense to change this responsibility on the `WebLoader` and pass values to the `Locator` instead?

Issue #806, opened 2016-09-27
Refactor controller business logic to service(s)


robfrawley:
With the `1.x` codebase, the controller defines a large subset of routing and conditional logic between the different objects required to successfully generate and cache a transformed image, and as the current controller interface requires a request object, such must be mocked when called outside a context where a request object is natively available.

This is inconvenient, and can be improved.

This `2.x` feature is to tear the vast majority of business logic out from the controller and define a collection of (or one?) services that implement the required logic to "glue" together the different layers needed. Most importantly, this new functionality should be request-agnostic.

robfrawley:
Current work on this feature is being contributed by @jehaby in my [sandbox repository clone](https://github.com/src-run/liip-imagine-bundle-sandbox/pull/2). A formal PR will originate from the sandbox repo once an implementation is finalized, for inclusion in this repo's `2.x` branch.

robfrawley:
@jehaby Just moved this to the ["in peer review"](https://github.com/liip/LiipImagineBundle/projects/1) stage on the [project tracker](https://github.com/liip/LiipImagineBundle/projects/1). I will be taking a look over the next few days to provide feedback on this initial stab at implementing this feature!

dbu:
the controller seems cleaned up now, i think this has been done.

PR #922, opened 2017-05-05
Introduce FilterService as abstraction for creating images

rpkamp:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | MIT
| Doc PR |

This PR splits all the image manipulation from the ImagineController and ResolveCacheCommand classes to a new FilterService class, which was on the roadmap for 2.x:

> Move Controller Business Logic into Service (Allow for use of service without Request)

The advantages of this are:
- Controller is only concerned with HTTP infrastructure
- Command is only concerned with CLI I/O
- Controller and command are easier to read, more highlevel
- Functionality is easier to use and extend in custom controllers/command from third parties
- FilterManager#applyFilter is only called in one spot throughout the entire codebase, making it easier to allow for multiple filter implementations later on

This PR is missing tests for the FilterService class, but I intend to add those once everyone agrees on a general direction of this PR. I did test the functionality in the browser (by using it in a separate test project) and that works as expected.

**Discussion**
In the FilterService the method `createFilteredImage` is closely tied to `getUrlOfFilteredImage`, but the first is to create the image and the latter is to retrieve the URL of the created image. Same thing goes for `createFilteredImageWithRuntimeFilters` and `getUrlOfFilteredImageWithRuntimeFilters`.
I split these out because of the CQS principle, but have now introduced a temporal coupling between the two. A more pure CQS approach would be to generate a URL first and then pass it to `createFilteredImage` as the desired URL for the image. However, it would be a mayor overhaul at this point to accomplish that, if indeed it is at all possible.

An alternative could be to let `createFilteredImage` return the URL of the resulting image (whether is was obtained from cache or not) in kind of find-or-create pattern way. This breaks CQS, but might be easier to understand and is less verbose in usage because one would only need one call instead of two to create a filtered image and obtain its URL.


Service/FilterService.php
```diff
@@ -0,0 +1,164 @@
...
+        return $this->cacheManager->resolve($runtimePath, $filter, $resolver);
+    }
+
+    /**
+     * @param string $path
+     * @param string $filter
+     * @param array|null $runtimeFilters
+     * @return \Liip\ImagineBundle\Binary\BinaryInterface
+     * @throws \Liip\ImagineBundle\Exception\Imagine\Filter\NonExistingFilterException
+     */
```
robfrawley: All doc blocks need their CS fixes such that they are ordered `param`, `throws`, `return`, each type is separated by a two newlines, and have their `param` variables aligned. We also use short class names, even if they aren't used in the current file a and require an additional `use` statement.
```php
/**
 * @param string     $path
 * @param string     $filter
 * @param array|null $runtimeFilters
 *
 * @throws \Liip\ImagineBundle\Exception\Imagine\Filter\NonExistingFilterException
 *
 * @return \Liip\ImagineBundle\Binary\BinaryInterface
 */
```

rpkamp: I'm confused regarding using short class names, as you said the complete opposite in this comment: https://github.com/liip/LiipImagineBundle/pull/908/files/f9cb8243475a1e492cb77ac5e639d306381fe9a8#diff-3edb79e074a2b20890b43396b316b5d3L60

rpkamp: Or is the short class name only used for `@param` but not for `@throws` and `@return`?

robfrawley: My bad, regarding the referenced comment, no I thought that was a `@covers` which is where we've been adding leading backslashes. That, too, should be fixed, I just skimmed over what it was too fast.

robfrawley: We have a `php-cs-fixer` `require-dev` requirement as well as the associated style settings; I would just run that, which will take care of everything but the FQCNs.

robfrawley: Erm, looks like our fixer rules are out of date and most of them are skipped, so that isn't a sure-fire bet either...

rpkamp: Is this what you mean? https://github.com/liip/LiipImagineBundle/pull/922/commits/4bce009f73c55659143d1c731709aed78b435e19

robfrawley: Short of the FQCNs that still exist, yes.

robfrawley: These instances *should* be the only examples of FQCNs: https://github.com/liip/LiipImagineBundle/blob/1.0/Tests/Binary/Loader/FileSystemLoaderTest.php#L22

rpkamp: Ah yes, that makes sense. Thanks.

robfrawley:
The RST documentation (which is what is pushed to [Symfony's online documentation website](http://symfony.com/doc/current/bundles/LiipImagineBundle/index.html)) requires changes and additional pertinent to this change, as well.

rpkamp:
I've merged this PR with master so I could use short array syntax again :tada:

Koc:
Lets wait until #926 will be merged into 2.0

Koc:
Also closes #806

rpkamp:
Not sure what's going on with https://github.com/liip/LiipImagineBundle/pull/926 since it's closed but not merged?

robfrawley:
The `1.x` branch will be merged with `2.0` once we release `1.8.0` (see #928).

Koc:
You need apply your service also here https://github.com/liip/LiipImagineBundle/blob/2.0/Async/ResolveCacheProcessor.php

robfrawley:
All branches up to date.

rpkamp:
I've updated the `ResolveCacheProcessor` to make use of the new `FilterService`.

rpkamp:
Does anyone have any thoughts on the CQS vs Temporal Coupling discussion in my initial PR message? 

The more I think about it, the more I think I'd rather have a CQS violation than temporal coupling, so instead of one method to query what the URL would be and a command to create the image, create one find-or-create method that creates the image if it doesn't exist and then return the URL, or if the image already existed just return the URL.
The reason I like it better is because it easier to understand and harder to get wrong for people starting out with this bundle.

robfrawley:
@rpkamp 

> Let createFilteredImage return the URL of the resulting image (whether is was obtained from cache or not) in kind of find-or-create pattern way. This breaks CQS, but might be easier to understand and is less verbose.

Yes, I absolutely prefer this approach.

rpkamp:
@robfrawley 

> Yes, I absolutely prefer this approach.

Done. This approach makes more sense for this project than the previous approach :smiley:

rpkamp:
Could I get a review on this one please?

Koc:
Nice PR, can anybody review?

makasim:
LGTM

rpkamp:
This PR has been open for a long time (since May of this year).
Anything I can do to advance the merging of this PR?

makasim:
@rpkamp  Thank you for the contribution.

Issue #924, opened 2017-05-06
Symfony Flex Recipe


robfrawley:
| Q | A
| --- | ---
| Bug Report? | no
| Feature Request? | yes
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | 1.7.x

I've created an initial stab at a [Symfony Flex](https://github.com/symfony/flex) contributor [recipe](https://github.com/symfony/recipes#symfony-recipes), which I'll be submitting. Take a look at [my fork of the contributor recipe repo](https://github.com/robfrawley/recipes-contrib) and the added files to support this bundle, and let me know if we should add any additional items to it.

---

See https://github.com/symfony/recipes-contrib/pull/14 for the pull request submitted against the contributed recipe repository.

antoligy:
Great idea, and the recipe you've submitted looks like a good starting point too (We don't want to overcomplicate/prematurely overconfigure if this is a potential entrypoint for new users!).  I've posted a comment about the issue raised by the PR checker.

LGTM!

Koc:
@robfrawley why copy-paste and not just include bundle routing file inside `routing/imagine.yaml`?

robfrawley:
@Koc The original intention of using YAML in the Flex recipe was to lower the entry barrier for someone looking to, perhaps, edit the route paths; XML is generally more difficult for newcomers to follow or edit. Do we want to prioritize ease of use in the recipe or duplication removal?

robfrawley:
How does everyone feel about converting our XML to YAML so it can be easily used by the recipe in a user-friendly format, as well as used in the Flex recipe through the `manifest.json` file (instead of having two copies of it).

This question only applies to the `2.x` branch as we obviously cannot remove the XML in `1.x` and break everyone's prior usage. @Koc @antoligy Thoughts?

robfrawley:
@Koc Perhaps we can combine a few of your ideas to resolve this?

Let's change `imagine.xml` to import our new `imagine.yml` file on the `1.x` branch, and then remove the `imagine.xml` file on the `2.x` branch, giving us:

```xml
<!-- /Resources/config/routing.xml -->

<?xml version="1.0" encoding="UTF-8" ?>
<routes xmlns="http://symfony.com/schema/routing"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://symfony.com/schema/routing
        http://symfony.com/schema/routing/routing-1.0.xsd">

[...]

PR #925, opened 2017-05-06
Deprecate XML routing file in favor of YAML

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | #924 
| License | MIT
| Doc PR | 

Deprecating the XML routing file in favor of YAML enables providing a more user-friendly format for the Symfony Flex recipe, and YAML is, IMHO, generally more user-friendly for this bundle as a stand-alone package, as well.

The old `routing.xml` file remains, to provide backward compatibility for users, but instead of containing routes it simply imports the new `routing.yaml` file. The XML file is deprecation in the `UPGRADE.md` file and can be safely removed from the `2.x` branch entirely.

The installation RST documentation has been updated to use the YAML routing varient, as well.

### DIFF
--- a/Resources/config/routing.xml
+++ b/Resources/config/routing.xml
@@ -5,15 +5,6 @@
         xsi:schemaLocation="http://symfony.com/schema/routing
         http://symfony.com/schema/routing/routing-1.0.xsd">
 
-    <route id="liip_imagine_filter_runtime" path="/media/cache/resolve/{filter}/rc/{hash}/{path}" methods="GET">
-        <default key="_controller">%liip_imagine.controller.filter_runtime_action%</default>
-        <requirement key="filter">[A-z0-9_\-]*</requirement>
-        <requirement key="path">.+</requirement>
-    </route>
+    <import resource="@LiipImagineBundle/Resources/config/routing.yaml" />
 
-    <route id="liip_imagine_filter" path="/media/cache/resolve/{filter}/{path}" methods="GET">
-        <default key="_controller">%liip_imagine.controller.filter_action%</default>
-        <requirement key="filter">[A-z0-9_\-]*</requirement>
-        <requirement key="path">.+</requirement>
-    </route>
 </routes>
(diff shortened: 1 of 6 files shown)


Resources/config/routing.yaml
```diff
@@ -0,0 +1,23 @@
+---
+
+liip_imagine_filter_runtime:
+    path: /media/cache/resolve/{filter}/rc/{hash}/{path}
+    defaults:
+        _controller: '%liip_imagine.controller.filter_runtime_action%'
+    methods:
+        - GET
+    requirements:
+        filter: '[A-z0-9_\-]*'
```
rpkamp: If a dash is at the last position of a character match you don't need to escape it, so you could use `[A-z0-9_-]*`.

Resources/config/routing.xml
```diff
@@ -5,15 +5,6 @@
...
-    </route>
+    <import resource="@LiipImagineBundle/Resources/config/routing.yaml" />
 
-    <route id="liip_imagine_filter" path="/media/cache/resolve/{filter}/{path}" methods="GET">
-        <default key="_controller">%liip_imagine.controller.filter_action%</default>
-        <requirement key="filter">[A-z0-9_\-]*</requirement>
-        <requirement key="path">.+</requirement>
-    </route>
-</routes>
+</routes>
```
Pierstoval: Missing new line here :)

Resources/config/routing.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
Pierstoval: Why?

robfrawley:
@rpkamp @Pierstoval Comments addressed; thanks!

UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade
 
+## 1.7.5
```
Pierstoval: In terms of semver, deprecations should come in minor versions, not patch versions, is it a problem if you need to push these changes to 1.8.0 instead of 1.7.5?

robfrawley: Nope

Pierstoval: Ok, well if it's a bugfix it's totally fine 😆

PR #931, opened 2017-05-12
Add $quantTable parameter to MozJpeg post processor

franmomu:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| License | MIT

I wanted not to use this parameter, this makes possible to change or unset the quant-table parameter that right now is hardcoded.

### DIFF
--- a/Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
@@ -32,18 +32,24 @@ class MozJpegPostProcessor implements PostProcessorInterface, ConfigurablePostPr
     /** @var null|int Quality factor */
     protected $quality;
 
+    /** @var null|int quantization table */
+    protected $quantTable;
+
     /**
      * Constructor.
      *
      * @param string   $mozjpegBin Path to the mozjpeg cjpeg binary
      * @param int|null $quality    Quality factor
+     * @param int|null $quantTable    Quantization table
      */
     public function __construct(
         $mozjpegBin = '/opt/mozjpeg/bin/cjpeg',
-        $quality = null
+        $quality = null,
+        $quantTable = 2
     ) {
         $this->mozjpegBin = $mozjpegBin;
         $this->setQuality($quality);
+        $this->setQuantTable($quantTable);
     }
 
     /**
@@ -54,6 +60,18 @@ class MozJpegPostProcessor implements PostProcessorInterface, ConfigurablePostPr
     public function setQuality($quality)
     {
         $this->quality = $quality;
+
+        return $this;
+    }
+
+    /**
+     * @param int $quantTable
+     *
+     * @return MozJpegPostProcessor
+     */
+    public function setQuantTable($quantTable)
+    {
+        $this->quantTable = $quantTable;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
```diff
@@ -88,8 +106,12 @@ public function processWithConfiguration(BinaryInterface $binary, array $options
         $pb = new ProcessBuilder(array($this->mozjpegBin));
 
         // Places emphasis on DC
-        $pb->add('-quant-table');
-        $pb->add(2);
+        $quantTable = array_key_exists('quant_table', $options) ? $options['quant_table'] : $this->quantTable;
+        if (null !== $quantTable) {
```
antoligy: I suppose that this should never be null, if we're reliant on the value of 2 above?  If we were to arrive in that situation, this would be a BC break as we'd no longer be specifying it.

Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
```diff
@@ -32,18 +32,24 @@ class MozJpegPostProcessor implements PostProcessorInterface, ConfigurablePostPr
...
+    protected $quantTable;
+
     /**
      * Constructor.
      *
      * @param string   $mozjpegBin Path to the mozjpeg cjpeg binary
      * @param int|null $quality    Quality factor
+     * @param int|null $quantTable    Quantization table
      */
     public function __construct(
```
antoligy: Wondering if there are any other ways we can go about this, it'd be nice to specify a global default and that's definitely functionality we should look to surface, but given the sheer number of options that we can pass cjpeg it feels like adding a them all to the constructor is going to get messy quickly.

I guess this is related to #909.  Should we go ahead with this and defer a nicer approach to 2.0, or should we look to solve this now?
cc @lsmith77 @robfrawley

Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
franmomu: It should be `null` in the constructor, I set to `2` because right now is set to 2, so there is no BC break. Setting `quant-table` to `null` in filters would remove the use of this parameter (which is what I wanted to do).

Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
lsmith77: I guess the time is now for this “nicer” approach .. I guess we should use the OptionResolver component in place of ever extending constructor arguments?

lsmith77: as such we I should add, as we are about to soon freeze 1.x for features I am ok adding “the last round” of constructor parameters (same for #909) .. and then immediately refactor to the OptionResolver for 2.x

wdyt @antoligy / @robfrawley ?

antoligy: I agree, the OptionsResolver is a great component and makes perfect sense everywhere we keep extending the constructor.

franmomu:
Closed in favor of https://github.com/liip/LiipImagineBundle/pull/993

PR #930, opened 2017-05-11
[Data Loader] [Data Locator] [Dependency Injection] Pass root paths to FileSystemLocator during construction

rpkamp:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | discussion in https://github.com/liip/LiipImagineBundle/pull/908#issuecomment-299718198
| License | MIT

The point of this PR is to require a fully instantiated Locator in FileSystemLoader.
Before, you would need to pass a LocatorInterface and rootpaths to the FilesystemLoader, and the FilesystemLoader would then pass the rootpaths to the LocatorInterface, making the FilesystemLoader aware of the inner workings of LocatorInterface which it should not be.

By requiring a LocatorInterface with rootpaths set, this knowledge about inner workings of the LocatorInterface is removed.

This also removed the need for setOptions on the LocatorInterface, which did not really belong the main purpose of the interface (locating files) anyway.

It also removed the dependency on OptionsResolver from FileSystemLocator since the rootpaths can now be passed directly to the constructor, making the rootpaths a hard dependency instead of an optional one - which was incorrect.

**Update by @robfrawley:** Ran `php-cs-fixer` using new style rules and added types to locator interface/class method signatures.

PR #932, opened 2017-05-13
Pass roots to FileSystemLocator during construction (1.0 branch)

rpkamp:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | See https://github.com/liip/LiipImagineBundle/pull/930
| License | MIT

### DIFF
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -13,7 +13,6 @@ namespace Liip\ImagineBundle\Binary\Loader;
 
 use Liip\ImagineBundle\Binary\Locator\FileSystemLocator;
 use Liip\ImagineBundle\Binary\Locator\LocatorInterface;
-use Liip\ImagineBundle\Exception\InvalidArgumentException;
 use Liip\ImagineBundle\Model\FileBinary;
 use Symfony\Component\HttpFoundation\File\MimeType\ExtensionGuesserInterface;
 use Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesserInterface;
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -38,30 +38,41 @@ class FileSystemLoader implements LoaderInterface
     /**
      * @param MimeTypeGuesserInterface  $mimeGuesser
      * @param ExtensionGuesserInterface $extensionGuesser
-     * @param string[]                  $dataRoots
-     * @param LocatorInterface          $locator
+     * @param LocatorInterface|string[] $locatorOrDataRoots
      */
     public function __construct(
         MimeTypeGuesserInterface $mimeGuesser,
         ExtensionGuesserInterface $extensionGuesser,
-        $dataRoots
-        /* LocatorInterface $locator */
+        $locatorOrDataRoots
     ) {
         $this->mimeTypeGuesser = $mimeGuesser;
         $this->extensionGuesser = $extensionGuesser;
 
-        if (count($dataRoots) === 0) {
-            throw new InvalidArgumentException('One or more data root paths must be specified.');
-        }
+        if (is_array($locatorOrDataRoots)) { // BC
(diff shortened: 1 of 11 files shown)


Binary/Loader/FileSystemLoader.php
```diff
@@ -38,30 +38,41 @@ class FileSystemLoader implements LoaderInterface
...
+        $locatorOrDataRoots
     ) {
         $this->mimeTypeGuesser = $mimeGuesser;
         $this->extensionGuesser = $extensionGuesser;
 
-        if (count($dataRoots) === 0) {
-            throw new InvalidArgumentException('One or more data root paths must be specified.');
-        }
+        if (is_array($locatorOrDataRoots)) { // BC
+            if (func_num_args() === 4 && func_get_arg(3) instanceof LocatorInterface) {
```
rpkamp: I've been wondering about this one. To me this seems like a BC break, meaning this cannot be in a 1.x version, but according to semver should bump the version to 2.0.
On the other hand this is an internal class and people aren't supposed to create it themselves, so maybe it's okay.

robfrawley: Here is my take:
- The constructor never actually had a fourth parameter in the signature (it was commented out), so those extending this class won't have their code break.
- [According to semvar](http://semver.org/#how-should-i-handle-deprecating-functionality) we just need to increment the minor version to introduce a deprecation.

As long as you haven't broken any of the following signatures, we haven't broken any code:
- `MimeTypeGuesserInterface $mimeGuesser, ExtensionGuesserInterface $extensionGuesser, $dataRoots` (where $dataRoots can be a single string root or an array of roots)
- `MimeTypeGuesserInterface $mimeGuesser, ExtensionGuesserInterface $extensionGuesser, $dataRoots, LocatorInterface $locator` (where $dataRoots can be a single string root or an array of roots)
- `MimeTypeGuesserInterface $mimeGuesser, ExtensionGuesserInterface $extensionGuesser, LocatorInterface $locator`

Ensure your changes to the tests didn't remove any of the aforementioned signature checks, as well.

robfrawley: Do you agree with my assertions @antoligy?

rpkamp: > `MimeTypeGuesserInterface $mimeGuesser, ExtensionGuesserInterface $extensionGuesser, $dataRoots, LocatorInterface $locator (where $dataRoots can be a single string root or an array of roots)`

exactly that use case is broken in these proposed changes, because you can't pass data roots to the LocatorInterface anymore, since it's a constructor argument and cannot (and should not, imho) be changed at runtime.

Maybe it's better to just add deprecation notices to 1.0 but don't actually change anything, and then use 2.0 to actually perform the change?

robfrawley: Sorry, I didn't even realize you'd removed `setOptions` from the `Loader`; that *needs* to stay with a deprecation notice. Then, in the `Locator` constructor, you should pass the array to `setOptions`, as before. As far as the other changes, of defaulting to 2.0 usage, that's fine, as long as we support the old use cases.

rpkamp: > As far as the other changes, of defaulting to 2.0 usage, that's fine, as long as we support the old use cases.

Do you mean the `setOptions` must stay? If so this PR is a step backward because it only adds complexity. I was hoping for removing `setOptions` altogether in 2.0 since it's basically a bundle-internal class and 2.0 already indicates there will be BC breaks.

rpkamp: We could then change this PR to only generate E_DEPRECATED warnings in 1.0, without actually changing anything.

robfrawley: No, I was saying to keep `setOptions` for `1.x` with a deprecation added; it can be removed for `2.x`, of course.

rpkamp: In that case I'll close this PR and create a new one with deprecation warnings only :smiley:

Issue #903, opened 2017-03-22
Multiple data_path and source image still can't be resolved


pzaj2:
| Q | A
| --- | ---
| Bug Report? | yes
| Feature Request? | no
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | 1.7.2

I'm having a hard time using this bundle since 1.7 update. I've been able, following other issues' answers to resolve issue with symlinks, at least that's what I thought. I'm currently experiencing quite opposite issue.

I'm using Deployer PHP to deploy.

this is my loaders part of config:

```yml
liip_imagine:
    loaders:
        default:
            filesystem:
                data_root: "%liip_imagine_data_path%"
```

and my parameter is set that way:

```yml
liip_imagine_data_path:
        - '%kernel.root_dir%/../../../shared/web/'
        - '%kernel.root_dir%/../web/'
```
It seems these paths are resolved correctly, because the exception being logged is as follows:
```php
"Source image could not be found" at /var/www/releases/28/vendor/liip/imagine-bundle/Controller/ImagineController.php line 101 {
	"exception":"(...): 
		Source image could not be found at /var/www/releases/28/vendor//liip/imagine-bundle/Controller/ImagineController.php:101, 
			(...)NotLoadableException(code: 0): 
				Source image not resolvable \"images/application-octet-stream.png?v=1490184091182\" in root path(s) \"/var/www/shared/web:/var/www/releases/28/web\" at /var/www/releases/28/vendor/liip/imagine-bundle/Binary/Locator/FileSystemLocator.php:52)"} []
```

The problem is that there's a file
"/var/www/releases/28/web/images/application-octet-stream.png" and because of that I believe it should be resolvable, meaning it's either something I don't understand here or there's a bug somewhere.

robfrawley:
The image you are requesting, per the exception message, is `images/application-octet-stream.png?v=1490184091182` whereas the file you want to load is `images/application-octet-stream.png`.

pzaj2:
@robfrawley I understand and I did test it, you are right about that, I suppose my best shot would be to remove the twig all to asset function in order to strip the version parameter. Or perhaps there's some sort of a "hack", or parameter, to ignore certain parts of request URL?

robfrawley:
Based on a cursory examination of the code, I do not see any special handling or configuration options that would allow it to ignore the query string section of a path. A pull request enabling such would be welcome.

rpkamp:
[...]

PR #933, opened 2017-05-14
[Templating] Handle uri query strings passed to templating extension and helper

rpkamp:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/903
| License | MIT

Gracefully handle any query strings in paths passed to the templating filters.
See added tests for behaviour.

### DIFF
--- a/Templating/Helper/ImagineHelper.php
+++ b/Templating/Helper/ImagineHelper.php
@@ -30,7 +30,10 @@ class ImagineHelper extends Helper
     }
 
     /**
-     * Gets the browser path for the image and filter to apply.
+     * Gets the browser path for the image and filter to apply. If your path inadvertently contains a query string
+     * - which might happen if you use asset versioning - the query string will be stripped from the path, the
+     * URL will be resolved using the path without query string, and the stripped query string will be appended to
+     * the resulting URL.
      *
      * @param string $path
      * @param string $filter
@@ -40,7 +43,13 @@ class ImagineHelper extends Helper
      */
     public function filter($path, $filter, array $runtimeConfig = array())
     {
-        return $this->cacheManager->getBrowserPath($path, $filter, $runtimeConfig);
+        $pathParts = explode('?', $path, 2);
+        $url = $this->cacheManager->getBrowserPath($pathParts[0], $filter, $runtimeConfig);
+        if (empty($pathParts[1])) {
+            return $url;
+        }
+
+        return $url.(strpos($url, '?') ? '&' : '?').$pathParts[1];
     }
 
     /**
--- a/Templating/ImagineExtension.php
+++ b/Templating/ImagineExtension.php
@@ -41,18 +41,27 @@ class ImagineExtension extends \Twig_Extension
     }
 
     /**
-     * Gets the browser path for the image and filter to apply.
+     * Gets the browser path for the image and filter to apply. If your path inadvertently contains a query string
+     * - which might happen if you use asset versioning - the query string will be stripped from the path, the
+     * URL will be resolved using the path without query string, and the stripped query string will be appended to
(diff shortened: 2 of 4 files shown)


Templating/Helper/ImagineHelper.php
```diff
@@ -40,7 +43,13 @@ public function __construct(CacheManager $cacheManager)
      */
     public function filter($path, $filter, array $runtimeConfig = array())
     {
-        return $this->cacheManager->getBrowserPath($path, $filter, $runtimeConfig);
+        $pathParts = explode('?', $path, 2);
```
rpkamp: I'd love to hear a better name than `$pathParts` :roll_eyes:

I tried `list($path, $queryString)`, which looks a lot better, but doesn't work when the original path doesn't contain a question mark (undefined index 1)

robfrawley:
We've discussed this in the past and one other idea brought up was to expose a configuration option that would enable this behavior. Also, isn't it possible that such handling should exist outside the templating component as well?

rpkamp:
I consciously restricted the behaviour to the templating component because of the nature of templates, where it is easy to mix up paths and URLs. Other components should in my opinion only deal with paths and not concern themselves with query strings.

In terms of hexagonal architecture I think of the templating component as an infrastructure thing where user input is received that might be messy, whereas after that we enter the application layer, where input should be normalised to adhere to the application spec. So that is why I've made the templating component forgiving of query strings and fix them when sending them to the application and adding them back to the application response, but the rest of the bundle is still blissfully unaware of query strings.

felipyamorim:
+1

Templating/Helper/ImagineHelper.php
felipyamorim: you can try this, to keep list function use.

`       
 list($path, $queryString) = array_pad(explode('?', $path, 2), 2, null);
`

rpkamp: It works, but is quite hard to read IMO.

felipyamorim: I agree, $pathParts sounds betters.

Waiting this merge.

robfrawley:
So, while I prefer _this_ behavior over our current handling, I consider this a BC break. Many filesystems are happy to allow a filename such as "my-cool-image?with=an&odd=name.jpg" and this change would break such a usage.

As such, we need to default to the old behavior and add the ability to enable this behavior via a configuration toggle for the `1.x` branch. This configuration option can then have its default changed on the `2.x` branch to enable this handling by default.

I already have a commit that adds the configuration required to do this, and I'll push it to this PR shortly, as I'd like to include this in the next release. Just wanted to give everyone a heads up as to any progress.

rpkamp:
This PR has been open for a long time (since May of this year).
Anything I can do to advance the merging of this PR?

makasim:
> As such, we need to default to the old behavior and add the ability to enable this behavior via a configuration toggle for the 1.x branch. This configuration option can then have its default changed on the 2.x branch to enable this handling by default.

@rpkamp you could as suggested by @robfrawley or simply change the base branch to 2.0

rpkamp:
I've been thinking about this some more and have come to the conclusion that this PR solves the wrong problem. The problem is that the filter is applied to the output of `asset`, which is very inefficient as it would first copy the image from where-ever it is to the web root and then modify the copied image in the webroot.
So instead of allowing query strings in the render extensions we should discourage people from loading images using `asset`, but set up a correct loader for the path where there images are instead.

As such, I'm closing this PR.

PR #934, opened 2017-05-14
Use new Twig notation

z38:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

~~After the removal of the Symfony Templating component, templates can no longer be referenced with `MyBundle:Foo:bar.html.twig`.~~ When Symfony is installed without the Templating component, bundle references like `MyBundle:Foo:bar.html.twig` are not resolved. This PR updates all occurrences of the old syntax to use the new native Twig prefix syntax. This allows the bundle to be used even when the Templating component is not present.

### DIFF
--- a/DependencyInjection/LiipImagineExtension.php
+++ b/DependencyInjection/LiipImagineExtension.php
@@ -94,7 +94,7 @@ class LiipImagineExtension extends Extension
 
         $container->setParameter('twig.form.resources', array_merge(
             $container->hasParameter('twig.form.resources') ? $container->getParameter('twig.form.resources') : [],
-            ['LiipImagineBundle:Form:form_div_layout.html.twig']
+            ['@LiipImagine/Form/form_div_layout.html.twig']
         ));
     }
 


robfrawley:
This is contingent on symfony/symfony#21036 actually being merged, correct?

z38:
You're right, it seems like component is not actually deprecated yet. I came across this issue when installing this bundle in conjunction with the (unreleased) Symfony 3.3. Based on [a compiler pass](https://github.com/symfony/symfony/blob/03914e9e9bdf11f13d38eb52e4cde1569c51caef/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/ExtensionPass.php#L84-L92) in TwigBundle, the loader which supports bundle references is only loaded when the Templating component is available.

I will rephrase the PR description to state that this is just a more compatible syntax (as it will work whether the component is installed or not).

trsteel88:
This would be a good addition. Namespaced template paths are faster as per
the documentation "As an added bonus, the namespaced syntax is faster"

http://symfony.com/doc/current/templating/namespaced_paths.html

z38:
With Symfony 3.3 released, I poked around a bit more. This patch is needed whenever the Templating component is not loaded by the Framework Bundle (disabled per default, but probably enabled in most cases because Symfony Standard Edition enables it).

z38:
@robfrawley Just saw that the pull-request is still being reviewed, do you need me to fix anything?

cedricziel:
Good to go, Thx!

@robfrawley , @antoligy - wdyt?

robfrawley:
Thanks @z38!

PR #939, opened 2017-05-21
Use simplified Symfony version comparison operation and CS fixes

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

This changes the internal behavior of `SymfonyFramework::kernelVersionCompare()` by using the `Kernel::VERSION_ID` constant instead of the separated version numbers (major, minor, patch). The resulting API behavior is identical. To see only these changes, reference commit [39d820](https://github.com/robfrawley/LiipImagineBundle/commit/46c7db19ea639b7ea952dd8a1e8bbd4339d820c1).

The second commit contains automated CS fixes from a PHP CS Fixer run.

Also of interest, I added the internal docblock to the `SymfonyFramework` class, which I suggest we more regularly adopt for code we don't intend to be used by consuming projects.

### DIFF
--- a/Utility/Framework/SymfonyFramework.php
+++ b/Utility/Framework/SymfonyFramework.php
@@ -13,6 +13,9 @@ namespace Liip\ImagineBundle\Utility\Framework;
 
 use Symfony\Component\HttpKernel\Kernel;
 
+/**
+ * @internal
+ */
 class SymfonyFramework
 {
     /**
@@ -75,19 +78,6 @@ class SymfonyFramework
      */
     private static function kernelVersionCompare($operator, $major, $minor = null, $patch = null)
     {
-        $vernum = $major;
-        $kernel = Kernel::MAJOR_VERSION;
-
-        if ($minor) {
-            $vernum .= '.'.$minor;
-            $kernel .= '.'.Kernel::MINOR_VERSION;
-
-            if ($patch) {
-                $vernum .= '.'.$patch;
-                $kernel .= '.'.Kernel::RELEASE_VERSION;
-            }
-        }
-
-        return version_compare($kernel, $vernum, $operator);
+        return version_compare(Kernel::VERSION_ID, sprintf("%d%'.02d%'.02d", $major, $minor ?: 0, $patch ?: 0), $operator);
     }
 }
--- a/Async/CacheResolved.php
+++ b/Async/CacheResolved.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Liip\ImagineBundle\Async;
 
 use Enqueue\Util\JSON;
@@ -14,9 +15,9 @@ class CacheResolved implements \JsonSerializable
      * @var \string[]
      */
     private $uris;
-    
+
     /**
-     * @param string $path
+     * @param string        $path
      * @param string[]|null $uris
      */
     public function __construct($path, array $uris)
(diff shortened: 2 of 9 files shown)


Utility/Framework/SymfonyFramework.php
```diff
@@ -75,19 +78,6 @@ public static function isKernelLessThan($major, $minor = null, $patch = null)
...
-            $kernel .= '.'.Kernel::MINOR_VERSION;
-
-            if ($patch) {
-                $vernum .= '.'.$patch;
-                $kernel .= '.'.Kernel::RELEASE_VERSION;
-            }
-        }
-
-        return version_compare($kernel, $vernum, $operator);
+        return version_compare(Kernel::VERSION_ID, sprintf("%d%'.02d%'.02d", $major, $minor ?: 0, $patch ?: 0), $operator);
```
rpkamp: Why are you padding the minor and patch components? The previous code didn't do this and I don't see a reason why it would be needed.

robfrawley: Because we're now comparing to the `Kernel::VERSION_ID` constant, which looks something like `int(20801)` for version `2.8.1`.

robfrawley: By the way, I found that the `:?` is actually not required, as `null` is cast to `0` automatically, but I found this made the usage more clear, and I don't like relying on PHP's internal type casting.

rpkamp: Ah, my mistake, I thought you were comparing against VERSION, not VERSION_ID. 
That being said I can't find any documentation on the Version constant, so I'm not sure how it's constructed but from what I've seen thusfar numeric comparison should suffice.

robfrawley: As for simple numeric comparison, that isn't possible due to the nature of this method signature: it allows passing in the comparison operator, hence the usage of `version_compare`.

The `VERSION` constant isn't a bad choice, but it is sometimes followed by a string value like `-dev` and `-beta` and while I know `version_compare` can deal with those, it just doesn't seem as clear to me. From my 100% unscientific polling of other projects, usage of `VERSION_ID` seems to be the most often used way to version detect, which was another reason I'd chosen it originally. 

This change actually originated from a bug in `symfony/symfony#master` where the kernel's `VERSION_MAJOR` was accidentally set to `0` causing our tests to fail: I settled on `VERION_ID` after looking at what others used.

Do you feel that using `VERSION` instead of `VERSION_ID` would add any value?

rpkamp: Using VERSION instead of VERSION_ID is more in line with what the code previously did, but other projects use VERSION_ID and that works for them so there is some empirical evidence that it works I'd say go for it.

Of course, feature detection (`class_exists`, `method_exists` et al) is always better than kernel version comparison, but that's beside the point of this PR.

robfrawley: Yeah, feature detection is of course preferable. Thankfully, this class is no longer required (for either version or feature detection) on the `2.x` branch. You may also feel better (as I do) that version detection is only used once or twice in the production codebase; the majority of calls occur in the test suite.

:-)

rpkamp: Lol, that _does_ make me feel better :smiley:

Issue #938, opened 2017-05-20
Retina


kironet:
| Q | A
| --- | ---
| Bug Report? | no
| Feature Request? | no
| BC Break Report? | no
| RFC? | yes
| Imagine Bundle Version | last one

Hey,

is it possible to somehow intergrate retina support? 
If retina show double size.

Or how to generate 2 images.
first one car.jpg
second one car@2x.jpg

Thanks

robfrawley:
At the moment, you can just define two separate filter sets for normal and retina displays.

kironet:
@robfrawley thank. Sorry, but how to generate two images with different name
car.jpg nad car@2x.jpg? 

Right now i'm doing it like it is shown in documentation:

`<img class="replace-2x" src="{{ asset('/path/to/car.jpg') | imagine_filter('thumb') }}" />`

Thanks

Issue #123, opened 2012-11-19
Retina support


bartlomiej84:
I was wondering what would be the easiest way to add support for Retina display. What functionality I should add ? I don't want to modify bundle, I only want to replace some managers, or maybe a controller, but I have no idea where is the best place to start.

I would like this bundle to make 2 cached images with different resolutions (second twice bigger as first), in the same cached directory and off course with suffix to filename like `@2x`, so it can be supported by Retina.js. 

Any help would be appreciate. 
Thanks.

kironet:
What about 2017? Is retina supported?

lsmith77:
again from my point of view the approach to generate 2 images at once is wrong. whatever you do on the frontend (ie. css/js) simply has to reference the images with the relevant configuration to resize accordingly. so the topic is imho beyond the scope of this Bundle.

see also `srcset` attribute

havvg:
I neither see any issue for this bundle regarding retina support, @kironet. This issue has been closed over 4 years ago and the bundle is capable of rendering images since the beginning. It's only up to you, to support your retina requirements; see my previous comment for one possible solution.

From the perspective of an image, there is no "retina support". "Retina" merely "uses" a different resolution of an image being shown. If you upload an image with `4000x4000px` with `72ppi` resolution, then that's the "normal" one. If you want a retina version of it, you would require a `8000x8000px` version of the very same image and render it with e.g. `144ppi`, which would result in a `4000x4000px` display size. If you want the same image to be available at `288ppi`, then you would be required to upload a `16000x16000px` at `72ppi`. 

As you can see, there are multiple options to resolve this. I would recommend you to upload the highest resolution you have and render the lower resolutions at the target image size. For example when your requirement is having a `4x`, `2x` and `1x` resolution version of the image while preserving the display size, you would require to upload the `4x` one. Let's say you upload a `100x100px` at `288ppi`. That's your `4x` version. Now you can setup two filters to yield the `2x` and `1x` resolution version of it. You would need to set the resolution to `144ppi`, which first yields an image of size `200x200px` and then you scale it down to `100x100px`. The second filterset just does the same but with `72ppi` respectively to yield the `1x` version.

PR #941, opened 2017-05-22
[Filter] Add resolution filter loader

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #123, #938
| License | MIT
| Doc PR | 

Many questions have been asked about "retina support" including #123 and #938. While I do not believe this bundle needs to enable any retina-specific functionality, as it already allows users to define any number of "filter sets" to generate any number of variations required, we should expose a filter to properly handle changing the image resolution, or "resampling".

This PR adds a `ResampleFilterLoader` that allows for creating different "filter sets" with different resulting image resolution (not to be confused with dimensions). Included is the ability to set `x` and `y` resolution integers, a corresponding `unit` (one of "inch" or "centimeter"), and an optional resampling `filter` (one of `\Imagine\Image\ImageInterface::FILTER_%s`).

This is a first-pass at adding resampling support, and as such I'm looking for comments about the manner in which this is implemented, specifically [ResampleFilterLoader.php:45](https://github.com/liip/LiipImagineBundle/pull/941/files#diff-3dd9a8ba39966e3e29eaa3ceccfe0a43R45). The `imagine-library` package this bundle leverages for image transformations _doesn't offer any means for resampling outside of **saving out the file and reading it back in**_, which isn't the most straightforward manner to handle this operation. 

**Any thought on this implementation or ideas for an alternate?**

This PR also introduces the following items intrinsically need for the filter loader implemented:
- `\Liip\ImagineBundle\Exception\Imagine\Filter\LoadFilterException`
  Added for exceptions thrown from within filters loader implementations
- `Utility/OptionsResolver/OptionsResolver:setDefined()`
  Added to allow for setting optional options that should not have a corresponding default value

Tests for all added functionality have now been implemented, as well. ~Additions to the documentation are now yet implemented.~

### DIFF
--- a/Imagine/Filter/Loader/ResampleFilterLoader.php
+++ b/Imagine/Filter/Loader/ResampleFilterLoader.php
@@ -1,0 +1,153 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Imagine\Filter\Loader;
+
+use Imagine\Image\ImageInterface;
+use Imagine\Image\ImagineInterface;
+use Liip\ImagineBundle\Exception\Imagine\Filter\LoadFilterException;
+use Liip\ImagineBundle\Utility\OptionsResolver\OptionsResolver;
+use Liip\ImagineBundle\Exception\InvalidArgumentException;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\Exception\ExceptionInterface;
+
+class ResampleFilterLoader implements LoaderInterface
+{
+    /**
+     * @var ImagineInterface
+     */
+    private $imagine;
+
+    /**
+     * @param ImagineInterface $imagine
+     */
+    public function __construct(ImagineInterface $imagine)
+    {
+        $this->imagine = $imagine;
+    }
+
+    /**
+     * @param ImageInterface $image
+     * @param array          $options
+     *
+     * @throws LoadFilterException
+     *
+     * @return ImageInterface
+     */
+    public function load(ImageInterface $image, array $options = array())
+    {
+        $options = $this->resolveOptions($options);
+        $tmpFile = $this->getTemporaryFile($options['temp_dir']);
+
+        try {
+            $image->save($tmpFile, $this->getImagineSaveOptions($options));
+            $image = $this->imagine->open($tmpFile);
+            $this->delTemporaryFile($tmpFile);
+        } catch (\Exception $exception) {
+            $this->delTemporaryFile($tmpFile);
+            throw new LoadFilterException('Unable to save/open file in resample filter loader.', null, $exception);
+        }
+
+        return $image;
+    }
+
+    /**
+     * @param string $path
+     *
+     * @throws \RuntimeException
+     *
+     * @return string
+     */
+    private function getTemporaryFile($path)
+    {
+        if (!is_dir($path) || false === $file = tempnam($path, 'liip-imagine-bundle')) {
+            throw new \RuntimeException(sprintf('Unable to create temporary file in "%s" base path.', $path));
+        }
+
+        return $file;
+    }
+
+    /**
+     * @param $file
+     *
+     * @throws \RuntimeException
+     */
+    private function delTemporaryFile($file)
+    {
+        if (file_exists($file)) {
+            unlink($file);
+        }
+    }
+
+    /**
+     * @param array $options
+     *
+     * @return array
+     */
+    private function getImagineSaveOptions(array $options)
+    {
+        $saveOptions = array(
+            'resolution-units' => $options['unit'],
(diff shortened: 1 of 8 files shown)


robfrawley:
Looks like Packagist endpoints are having some trouble right now (or another similar issue), resulting in Composer failing to build anything in our Travis tests. I've gone ahead and made @Seldaek aware in a similar issue at composer/composer#6342.

Imagine/Filter/Loader/ResampleFilterLoader.php
```diff
@@ -0,0 +1,160 @@
...
+        $resolver->setAllowedTypes('temp_dir', array('string'));
+        $resolver->setAllowedTypes('filter', array('string'));
+
+        $resolver->setAllowedValues('unit', array(
+            ImageInterface::RESOLUTION_PIXELSPERINCH,
+            ImageInterface::RESOLUTION_PIXELSPERCENTIMETER
+        ));
+
+        $resolver->setNormalizer('filter', function (Options $options, $value) {
+            foreach (array('\Imagine\Image\ImageInterface::FILTER_%s', '\Imagine\Image\ImageInterface::%s', '\%s', '%s') as $format) {
```
Koc: do we need leading backslashes here?

robfrawley: I added them as a very "micro-optimization" as it does remove some opcodes (otherwise it first searches in the defined namespace, followed by the global namespace, the latter of which we want to force for the first two formats, but now that you bring this up, we could likely change:
```php
array('\Imagine\Image\ImageInterface::FILTER_%s', '\Imagine\Image\ImageInterface::%s', '\%s', '%s')
```
to
```php
array('\Imagine\Image\ImageInterface::FILTER_%s', '\Imagine\Image\ImageInterface::%s', '%s')
```

Of course, we don't *need* them anywhere, but I think it adds clarity and acts as a small speedup.

robfrawley:
@antoligy @cedricziel Pending any objections, I'm going to merge this forinclusion in the `0.9.0` release.

robfrawley:
Last call for comments @antoligy @cedricziel

antoligy:
Seems like the obvious points have already been made, I'm very happy with this to be merged.

Is it possible to process the file with a buffered stream versus buffering it into a file?  (I'm actually unfamiliar with the underlying IMagick operation so it may be a limitation there).

Also, as a more general point, do we have a given convention for when operations are IMagick only, because if so I think we can possibly start adding filters a little more quickly (I've got a few between projects relying on IMagick only features, including another attempt at DPR adjustment).

robfrawley:
@antoligy Thanks for the quick review. Here is a verbose response. ;-)

> Is it possible to process the file with a buffered stream versus buffering it into a file? (I'm actually unfamiliar with the underlying IMagick operation so it may be a limitation there).

The resampling functionality exposed by the `imagine-library` package is not well documented, but from what I've been able to gather by looking at the brief explanation in their docs, as well as the source code itself, we must **explicitly save the file for the resolution to be applied**. I can't find any way around this.

>Also, as a more general point, do we have a given convention for when operations are IMagick only, because if so I think we can possibly start adding filters a little more quickly (I've got a few between projects relying on IMagick only features, including another attempt at DPR adjustment).

The way I implemented the functional resample test case was confusing. It's **not** the *filter implementation* that is `imagick`-only, [but **only** the *test itself*](https://github.com/liip/LiipImagineBundle/pull/941/commits/309690d4811213dcb4b86d2ae5705699d40afa50#diff-1da778d924a0f6dbd661a6233f26aa6cR26) that required `imagick`, as the `Imagine\Imagick\Imagine` class was used to read the resulting image's resolution for the test's assertion.

I've gone ahead and abstracted the test case a bit to support both `imagick` and `gmagick`. It isn't currently possible for the test to support `gd` as [its `imageresolution()` function](http://php.net/manual/en/function.imageresolution.php) wasn't added until PHP `7.2`.

**As to your question about driver-specific filters**, I do not have any particular issue with allowing such implementation, so long as the following conditions are met:
- *[Requirement]* The filter documentation is obvious and explicit about driver support.
- *[Requirement]* An instructive and detailed exception is thrown when an unsupported filter is used.
- *[Recommendation]* Pre-processing of the configuration to throw an immediate exception if unsupported filters are configured. This exception would be thrown regardless of whether the filter is ever actually called during the request. This would avoid surprises for the user by ensuring their configured filters are valid for their environment (for example, we don't want the exception to only crop up three months later when the user removed their cached images; we want it to be thrown the moment it enters their config).
- *[Recommendation]* Coinciding with the above recommendation, we could add a new interface for filters that are driver-aware (`DriverDependantFilterInterface` or something) that adds a new method to return an array of supported drivers, like `supportedDrivers(): string[]` or something similar.

---

The https://github.com/liip/LiipImagineBundle/pull/941/commits/cbbc9c4e73ad1eb8f1c6613250e61346d5aec890. commit enabled multiple driver support for the `resample` filter test (as explained above), and adds some additional file types and resolution sizes to the `resample` test via a new data provider.

gnutix:
@robfrawley I can't seem to find this code in the current version of this bundle, and yet I can't find any commit / remarks about it's deprecation or removal. Any idea what happened ?

PR #949, opened 2017-06-13
Use 307 temporary redirect instead of 301 permanently redirect

yplam:
| Q | A
| --- | ---
| Branch? | 1.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | maybe
| Deprecations? | no
| Tests pass? | 
| Fixed tickets | 
| License | MIT
| Doc PR | 

Since Symfony 3.2, RedirectResponse will remove 'cache-control' header from 301 redirect: https://github.com/symfony/symfony/blob/3.2/src/Symfony/Component/HttpFoundation/RedirectResponse.php .

The web server like nginx may cache the image request because it is 301 redirect to the same url (or maybe web browser like Chrome will do the same thing ), and that will causes a  "too many redirects" error.

### DIFF
--- a/Controller/ImagineController.php
+++ b/Controller/ImagineController.php
@@ -109,7 +109,7 @@ class ImagineController
                 );
             }
 
-            return new RedirectResponse($this->cacheManager->resolve($path, $filter, $resolver), 301);
+            return new RedirectResponse($this->cacheManager->resolve($path, $filter, $resolver), 307);
         } catch (NonExistingFilterException $e) {
             $message = sprintf('Could not locate filter "%s" for path "%s". Message was "%s"', $filter, $path, $e->getMessage());
 
@@ -177,7 +177,7 @@ class ImagineController
                 $resolver
             );
 
-            return new RedirectResponse($this->cacheManager->resolve($rcPath, $filter, $resolver), 301);
+            return new RedirectResponse($this->cacheManager->resolve($rcPath, $filter, $resolver), 307);
         } catch (NonExistingFilterException $e) {
             $message = sprintf('Could not locate filter "%s" for path "%s". Message was "%s"', $filter, $hash.'/'.$path, $e->getMessage());
 
--- a/Resources/doc/resolve-cache-images-in-background.rst
+++ b/Resources/doc/resolve-cache-images-in-background.rst
@@ -5,7 +5,7 @@ Overview
 --------
 
 By default the LiipImagineBundle processes the image on demand.
-It does in resolve controller and saves the result, does a 301 redirect to the processed static image.
+It does in resolve controller and saves the result, does a 307 redirect to the processed static image.
 The approach has its benefits.
 The most notable is simplicity.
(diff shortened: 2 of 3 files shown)


trsteel88:
It should be a 301 redirect for SEO reasons. The 301 tells Google and other search engines not to index the resolve url and instead follow it through and index the redirect.

trsteel88:
It should be a 301 redirect for SEO reasons. The 301 tells Google and other search engines not to index the resolve url and instead follow it through and index the redirect.

yplam:
Find out this is same as: https://github.com/liip/LiipImagineBundle/pull/439

Use 307 will not work friendly with aws.

So if I want to use WebPathResolver and want nginx to handle cached image directly, I need to override the filterAction?

trsteel88:
@yplam why don't you want nginx to cache the result? If the url is unique a cache should be a good thing.

If your images are always changing you should look at using a runtime config.

yplam:
My nginx config file:

```
location ~* /media/cache {
    	try_files $uri /app.php$is_args$args;
	expires 30d;
}
```

Resolver config:

```
liip_imagine:
    resolvers:
        default:
            web_path:
                cache_prefix: media/cache/resolve
```
Because filterAction return 301 redirect to the same url, and nginx(or chrome) cached it, that will make a "too many redirects" error.

Maybe the following code will works better (return 307 for the first visitor):

```
    {
        // decoding special characters and whitespaces from path obtained from url
        $path = urldecode($path);
        $resolver = $request->get('resolver');

        try {
            if (!$this->cacheManager->isStored($path, $filter, $resolver)) {
                try {
                    $binary = $this->dataManager->find($filter, $path);
                } catch (NotLoadableException $e) {
                    if ($defaultImageUrl = $this->dataManager->getDefaultImageUrl($filter)) {
                        return new RedirectResponse($defaultImageUrl);
                    }

                    throw new NotFoundHttpException('Source image could not be found', $e);
                }

                $this->cacheManager->store(
                    $this->filterManager->applyFilter($binary, $filter),
                    $path,
                    $filter,
                    $resolver
                );
                // return 307 for the first visitor
	        return new RedirectResponse($this->cacheManager->resolve($path, $filter, $resolver), 307);
            }

            return new RedirectResponse($this->cacheManager->resolve($path, $filter, $resolver), 301);
        } catch (NonExistingFilterException $e) {
            $message = sprintf('Could not locate filter "%s" for path "%s". Message was "%s"', $filter, $path, $e->getMessage());

            if (null !== $this->logger) {
                $this->logger->debug($message);
            }

            throw new NotFoundHttpException($message, $e);
        } catch (RuntimeException $e) {
            throw new \RuntimeException(sprintf('Unable to create image for path "%s" and filter "%s". Message was "%s"', $path, $filter, $e->getMessage()), 0, $e);
        }
    }

```

trsteel88:
Your cache prefix should not be the same url.

Try changing your cache prefix from ***media/cache/resolve*** to ***media/cache***

Issue #948, opened 2017-06-10
Best practice for a filter_set with multiple DataLoader


winzou:
| Q | A
| --- | ---
| Bug Report? | no
| Feature Request? | no
| BC Break Report? | no
| RFC? | yes
| Imagine Bundle Version | 1.7.1

Hello,

I have a usecase where I need to apply the same filters on the images coming from 2 different data loader:
* A Stream loader, Gaufrette, where user-uploaded images are
* A Filesystem loader, where some static images are

What is the best practice to apply the same filters on these images? I didn't find anything like `CascadeDataLoader` that could return the first existing image in multiple data loaders. Or is it better to define 2 different filter_sets and find a way to factorize their parameter (same thumbnail size, same post_processors, etc.) ?

Thanks for your help!

robfrawley:
Give #953 a try and let me know how it works for you.

winzou:
Thanks, I went for the chained loader. It works like a charm!

robfrawley:
@winzou Glad that the new `chain` loader worked out well for you!

PR #953, opened 2017-06-20
Add chain loader implementation

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #948
| License | MIT
| Doc PR | 

Chain data loader implementation. This allows one to define one "filter set" that searches multiple loader configurations (for example, this allows you to check both a Gaufrette stream loader and a file system loader simultaneously). 

Here is an example configuration for the loader:

```yaml
liip_imagine:

    loaders:

        foo:
            filesystem:
                # setup this loader here

        bar:
            flysystem:
                # setup this loader here

        baz:
            chain:
                # name the loaders to chain
                loaders: [foo, bar]

    filter_sets:

        my_custom_filter_set:
            data_loader: baz
            filters:
                # define your filters here
```

This implementation simply loops through the available loaders and selects the first valid result, so if the file exists in multiple loaders, the first defined in your configuration will be selected. For example, if `file.ext` exists in both the `foo` and `bar` loader, the one found via the `foo` loader will be returned because it is defined first for `baz.chain.loaders`.

### DIFF
--- a/Binary/Loader/ChainLoader.php
+++ b/Binary/Loader/ChainLoader.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Binary\Loader;
+
+use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
+
+class ChainLoader implements LoaderInterface
+{
+    /**
+     * @var LoaderInterface[]
+     */
+    private $loaders;
+
+    /**
+     * @param LoaderInterface[] $loaders
+     */
+    public function __construct(array $loaders)
+    {
+        $this->loaders = array_filter($loaders, function ($loader) {
+            return $loader instanceof LoaderInterface;
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function find($path)
+    {
+        foreach ($this->loaders as $loader) {
+            try {
+                return $loader->find($path);
+            } catch (\Exception $loaderException) {
+                // handle exception later
+            }
+        }
+
+        throw new NotLoadableException(vsprintf('Source image not resolvable "%s" using "%s" loaders.', array(
+            $path,
+            $this->getLoaderNamesString(),
+        )));
+    }
+
+    /**
+     * @return string
+     */
+    private function getLoaderNamesString()
+    {
+        $names = array();
+        foreach ($this->loaders as $n => $l) {
+            $names[] = sprintf('%s=[%s]', $n, get_class($l));
+        }
+
+        return implode(':', $names);
+    }
+}
--- a/Tests/Binary/Loader/ChainLoaderTest.php
+++ b/Tests/Binary/Loader/ChainLoaderTest.php
@@ -1,0 +1,138 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Tests\Binary\Loader;
+
+use Liip\ImagineBundle\Binary\Loader\ChainLoader;
+use Liip\ImagineBundle\Binary\Loader\FileSystemLoader;
+use Liip\ImagineBundle\Binary\Locator\FileSystemLocator;
+use Liip\ImagineBundle\Model\FileBinary;
+use Symfony\Component\HttpFoundation\File\MimeType\ExtensionGuesser;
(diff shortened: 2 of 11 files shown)


Tests/Functional/Binary/Loader/ChainLoaderTest.php
```diff
@@ -0,0 +1,42 @@
...
+        return $this->getService(sprintf('liip_imagine.binary.loader.%s', $name));
+    }
+
+    public function testFind()
+    {
+        static::createClient();
+
+        $loader = $this->getLoader('baz');
+
+        foreach (['images/cats.jpeg', 'images/cats2.jpeg', 'file.ext', 'bar-bundle-file.ext', 'foo-bundle-file.ext'] as $file) {
```
antoligy: Shorthand array syntax breaking PHP 5.3 builds

Binary/Loader/ChainLoader.php
```diff
@@ -0,0 +1,64 @@
...
+            $this->getLoaderNamesString()
+        ]));
+    }
+
+    /**
+     * @return string
+     */
+    private function getLoaderNamesString()
+    {
+        $names = [];
```
antoligy: Shorthand array syntax is breaking PHP 5.3 builds.

Tests/DependencyInjection/Factory/Loader/ChainLoaderFactoryTest.php
```diff
@@ -0,0 +1,102 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @covers \Liip\ImagineBundle\DependencyInjection\Factory\Loader\ChainLoaderFactory
+ */
+class ChainLoaderFactoryTest extends FactoryTestCase
+{
+    public function testImplementsLoaderFactoryInterface()
+    {
+        $rc = new \ReflectionClass('\Liip\ImagineBundle\DependencyInjection\Factory\Loader\ChainLoaderFactory');
```
antoligy: Might have asked this before but I wonder if these assertions on interface are still needed?

Tests/DependencyInjection/Factory/Loader/ChainLoaderFactoryTest.php
```diff
@@ -0,0 +1,102 @@
...
+            'chain' => array(
+                'loaders' => [
+                    'foo',
+                    'bar',
+                ],
+            ),
+        ));
+
+        $this->assertArrayHasKey('loaders', $config);
+        $this->assertCount(2, $config['loaders']);
```
antoligy: Worth asserting that the array itself is as we expect, versus the count?

Tests/Binary/Loader/ChainLoaderTest.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * @param FileBinary|mixed $return
+     * @param string|null      $message
+     */
+    private function assertValidLoaderFindReturn($return, $message = null)
+    {
+        $this->assertInstanceOf('\Liip\ImagineBundle\Model\FileBinary', $return, $message);
+        $this->assertStringStartsWith('text/', $return->getMimeType(), $message);
```
antoligy: Not a change, just curiosity--Is this assertion purely for the case where something is returned?

Binary/Loader/ChainLoader.php
antoligy: https://travis-ci.org/liip/LiipImagineBundle/jobs/244807011

Tests/Binary/Loader/ChainLoaderTest.php
robfrawley: Yup

robfrawley:
@antoligy Never tried to "dismiss" a review; wanted to see how that worked. ;-) Anyway, that was a crazy quick reviewing! How do you feel about the merits of this PR, as well as the implementation? Worth it? FYI: This still isn't ready for merge, as docs are missing.

Tests/DependencyInjection/Factory/Loader/ChainLoaderFactoryTest.php
robfrawley: IDK if they are "needed" but for consistency with the other loader tests, I feel like we should include it for now. In the meantime, I have changed it to remove the reflection-based method used for the much more sensible use of PHPUnit's `isInstanceOf()`.

antoligy:
Fair enough!  I didn't say anything about the merits of this PR earlier: I think it's a very useful addition, and similar projects of this type have similar logic.  For me it's particularly interesting in clustered environments where a system might have an asset available locally to it quicker than the same asset over a network.  Will be interesting to hear/see other uses that may emerge too!

Tests/Binary/Loader/ChainLoaderTest.php
antoligy: Very nice way of going about it, and a slightly better assertion IMO than just checking something is present.

robfrawley:
https://github.com/liip/LiipImagineBundle/pull/953#issuecomment-309846957:
> This still isn't ready for merge, as docs are still missing.

I'll open a new PR with the docs sometime tomorrow since that fell through the cracks.

cedricziel:
woops.

robfrawley:
@cedricziel It's all good; give #957 a review when you can.

PR #935, opened 2017-05-14
Allow v0.7 of Imagine Bundle

trsteel88:
https://github.com/avalanche123/Imagine/pull/547

| Q | A
| --- | ---
| Branch? | 1.0 and 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Tests pass? | yes
| License | MIT

PR #958, opened 2017-07-03
[Composer] Allow imagine-library version 0.7.0

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #935
| License | MIT
| Doc PR | 

This pull request adds the ability to utalize version `0.7.0` of the Imagine Library, and raises our upper limit constraint to `0.8.0`. By default, this will actually cause most people to upgrade from `0.6.0` to `0.7.0` of Imagine Library, but it allows those who explicitly constrain their project to Imagine Library `0.6.0` to remain on that version, as well. It provides a cleaner, alternate approach to #935 with passing tests. Fixes an important issue introduced by `3.4.3` of the `imagick` extension; see avalanche123/Imagine#547.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": "^5.3.9|^7.0",
-        "imagine/Imagine": "^0.6.3,<0.7",
+        "imagine/Imagine": "^0.6.3|^0.7.0,<0.8",
         "symfony/asset": "~2.3|~3.0",
         "symfony/filesystem": "~2.3|~3.0",
         "symfony/finder": "~2.3|~3.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
 matrix:
   include:
     - php: 5.3
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest" IMAGINE_VERSION=^0.6.3
       dist: precise
     - php: 5.6
       env: SYMFONY_VERSION=2.3.x-dev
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
     },
     "require": {
         "php": "^5.3.9|^7.0",
-        "imagine/Imagine": "^0.6.3,<0.7",
+        "imagine/Imagine": "^0.6.3|^0.7.0,<0.8",
```
enumag: The `,<0.8` is not necessary. `^0.7.0` means `>=0.7.0 <0.8.0.`

robfrawley: I realize; I believe the current syntax was used to provide clarity, as even before the change when it was `^0.6.3,<0.7`, the `<0.7` was unnecessary. It's important to note that the rule you mentioned *only* applies for fully enumerated versions, so it may be a good idea to retain the upper constraint so when someone comes along and changes it to, say `^0.7`, we don't open the floodgates.

There are a bunch of ways to write this, such as `^0.6.3|^0.7.0,<0.8`, `>=0.6.3,^0.6,<0.8`, `>=0.6.3,<0.8` or `~0.6.3|~0.7.0` (the latter of which I personally prefer), but the original intention seems to have been verbosity, which the current syntax provides, no?

If anything, I'd prefer `~0.6.3|~0.7.0`, honestly.

enumag: `^0.7` and `^0.7.0` are actually equivalent. But I understand your point. It is more clear for those who don't know how the the caret operator works.

robfrawley: > It's important to note that the rule you mentioned only applies for fully enumerated versions

I meant to append that with "above `1.x`" and I obviously used the wrong example of `0.7` where I intended to use `1.0`. Was typing quickly. ;-)

antoligy:
I guess to avoid any complaints on bumping most from .6 to .7, we potentially release under a minor version?

Either way 👍

trsteel88:
@robfrawley can you tag a version for this please?

robfrawley:
The next release (which will be tagged as `1.9.0`) is scheduled for this coming Monday (2017-07-17). The following PRs need to be properly reviewed and merged for that release: #933, #937, and #941.

You can reference the [1.9.0 milestone](https://github.com/liip/LiipImagineBundle/milestone/13) to track progress. With any luck, that date won't slip; with some luck, that date may be moved up. We'll see. :-)

trsteel88:
Any chance we can get this tagged sooner as a patch version @robfrawley? It's my understanding that bug fixes can/should be tagged straight away? e.g 1.8.1

Just to be a pain, we've just got this issue occurring on a number of sites where we have to manually edit the file on the server :(

robfrawley:
@trsteel88 See #983.

trsteel88:
Thank you. Appreciate the quick response.

cedricziel:
@trsteel88 As for fixes, you can use your own bugfixes by adding your fork as composer repository (https://getcomposer.org/doc/05-repositories.md#vcs) and require version `dev-$yourbranchname as 1.8.1`

robfrawley:
@trsteel88 The [`1.9.0` release](https://github.com/liip/LiipImagineBundle/releases/tag/1.9.0) is live! Enjoy.

trsteel88:
Thanks @robfrawley 

Good to know @cedricziel

PR #959, opened 2017-07-04
[enqueue] Upgrade to enqueue v5. Use its features

makasim:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes if you are migration from 1.0
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | docs are updated in this same PR

* Send a command (one-to-one) instead of an event as it was before.
* Send a reply to sender (optional)
* Make command exclusive. Means it is the only one processor on the given
queue. The message could be sent without any enqueue specific properties
and will still work.

### DIFF
--- a/Async/Commands.php
+++ b/Async/Commands.php
@@ -11,10 +11,10 @@
 
 namespace Liip\ImagineBundle\Async;
 
-class Topics
+class Commands
 {
     /**
      * @var string
      */
-    const CACHE_RESOLVED = 'liip_imagine_cache_resolved';
+    const RESOLVE_CACHE = 'liip_imagine_resolve_cache';
 }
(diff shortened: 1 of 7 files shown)


robfrawley:
@makasim Can you amend [UPGRADE.md](https://github.com/liip/LiipImagineBundle/blob/2.0/UPGRADE.md) to add a concise enumeration of user-facing changes introduced here (specifically BC breaks, showing the previous and new handling explicitly)?

makasim:
@robfrawley done

robfrawley:
I'd recommend the following UPGRADE notice instead of the current one (above the example):
> Enqueue's producer `send()` method has been renamed to `sendCommand()`. When interacting with the producer to resolve images in the background you must make the following changes to your code:

Otherwise, this PR looks good.

makasim:
Thanks for the review

PR #963, opened 2017-07-21
[Data Loader] Pass root paths to file system locator constructor (instead of loader)

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | #930
| License | MIT
| Doc PR | 

Continuation of #937 to pass roots directly to `LocatorInterface` instead of to `FileSystemLoader`, in preparation for breaking changes in `2.0`.

### DIFF
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -38,55 +38,47 @@ class FileSystemLoader implements LoaderInterface
     /**
      * @param MimeTypeGuesserInterface  $mimeGuesser
      * @param ExtensionGuesserInterface $extensionGuesser
-     * @param string[]                  $locatorOrDataRoots
-     * @param LocatorInterface          $locator
+     * @param string[]|LocatorInterface $locator
      */
-    public function __construct(
-        MimeTypeGuesserInterface $mimeGuesser,
-        ExtensionGuesserInterface $extensionGuesser,
-        $locatorOrDataRoots
-        /* LocatorInterface $locator */
-    ) {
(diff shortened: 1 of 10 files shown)


robfrawley:
I think this is ready to go. @cedricziel @antoligy Comments?

Binary/Loader/FileSystemLoader.php
```diff
@@ -38,30 +38,48 @@ class FileSystemLoader implements LoaderInterface
     /**
      * @param MimeTypeGuesserInterface  $mimeGuesser
      * @param ExtensionGuesserInterface $extensionGuesser
-     * @param string[]                  $dataRoots
-     * @param LocatorInterface          $locator
+     * @param string[]|LocatorInterface $locator
      */
```
antoligy: I would also place an `@deprecated` doc here.

robfrawley: Hmm; while I understand your intention, `@deprecated` is a doc-block scope declaration that would cause IDEs and other tools to believe the entire `__construct()` method had been deprecated, which is obviously not the case. Perhaps a comment explaining the depreciation would be more appropriate? We don't want people seeing "FileSystemLoader.php::~~__construct()~~" in their IDEs.

antoligy: Ahh, very good point and I didn't think of that.  A better idea is adding a comment, or maybe an `@see` referencing this PR or the docs added in this PR?

robfrawley: I think https://github.com/liip/LiipImagineBundle/pull/963/commits/23e79a7f41321010a2f8a38a00f49916ca64704a should do the trick. :-)

Issue #966, opened 2017-07-27
Force cache resolve ?


lucFM:
| Q | A
| --- | ---
| Bug Report? | no
| Feature Request? | yes
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | 1.8.0

Hi, is there a way to force cache resolve ? 
I would like to re generate all my thumbnails without remove them.

Thanks

robfrawley:
I actually needed exactly the same thing recently and had amended the console command to accomplish it. I turned my temporary code into a proper pull request in #967 --- seeh ow that works for you with the added option `--force` (which will force the resolved image to be regenerated, regardless of cache). If you want to do this in a batch fashion, you can do something like:

```bash
for f in "$(ls path/to/images/*.jpg)"; do
  bin/console liip:imagine:cache:resolve --filter=my_filter_name "${f}"
done
```

robfrawley:
An option to force resolution (regardless of the cache state) has been added to the console command (with the addition of the `--force` CLI option) as #967 for the `1.x` branch (merged) and #981 for the `2.x` branch (in review).

lucFM:
awesome !

PR #967, opened 2017-07-27
[Console] Add forced resolve to command and prittify and detail output

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #966
| License | MIT
| Doc PR | 

The PR changes the following aspects of the `liip:imagine:cache:resolve` command:
- Adds a `--force` option that forces re-resolution an image, regardless of cache.
- Expands the output to provide more detailed information on the actions executed.
- Allows resolution failures and continues with the other passed images and filters.
- Amends the `--help` output to make it easier to follow and clearer at a glance.
- A `--machine` option has been added that removed the title and summary from the output and only outputs machine-parsable lines.
- Deprecates the `--filters` option in favor of a `--filter` option (you only pass one filter per call to it, so it should not be plural).

Here is the new `--help` output:

liip-imagine-bundle-resolve-command-help

Here is the output with errors:

liip-imagine-bundle-cache-resolve-errors

Here is the new default behavior output (showing the image name, filter name, and resolved path):

liip-imagine-bundle-cache-resolve-cached-output

Here is the output when forcing resolution, ignoring the cache:

liip-imagine-cache-resolve-resolved-output

Here is the output with `--machine` enabled:

liip-imagine-bundle-cache-resolve-machine

And finally, here is the output showing `cached`, `resolved`, and `failed` items from one invocation:

liip-imagine-bundle-cache-resolve-complete

### DIFF
--- a/Command/RemoveCacheCommand.php
+++ b/Command/RemoveCacheCommand.php
@@ -26,12 +26,10 @@ class RemoveCacheCommand extends ContainerAwareCommand
             ->setName('liip:imagine:cache:remove')
             ->setDescription('Remove cache for given paths and set of filters.')
             ->addArgument('paths', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, 'Image paths')
-            ->addOption(
-                'filters',
-                'f',
-                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
-                'Filters list'
-            )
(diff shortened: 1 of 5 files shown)


robfrawley:
@cedricziel @antoligy I want to get a release tagged shortly; any objections to this PR? Time to review?

PR #943, opened 2017-05-29
Update ImageType to match symfony 3.*

lashus:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? |no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | MIT
| Doc PR |

Update getParent method to be compatbile with symfony 3.x

PR #975, opened 2017-08-17
Solve issues with file type, which wasn’t converted to the new standard

jorisros:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes/
| New feature? | no
| BC breaks? | no <only sf3 tested>
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #943
| License | MIT
| Doc PR |

### DIFF
--- a/Form/Type/ImageType.php
+++ b/Form/Type/ImageType.php
@@ -12,6 +12,7 @@
 namespace Liip\ImagineBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\FileType;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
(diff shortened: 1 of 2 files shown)


jorisros:
Thank you, i have created a issue for this. #976

cedricziel:
Thx @jorisros !

trsteel88:
According to composer.json php5.3 is supported. The class constant is only available from php5.5.

cedricziel:
The PR was made against the 2.0 branch, of the Bundle, which requires PHP 7. :)

Ref: https://github.com/liip/LiipImagineBundle/blob/2.0/composer.json#L21

robfrawley:
We actually discussed matching the Symfony 4.x requirement and having PHP 7.1 be the minimum requirement for 2.x, but I guess we forgot to make that change. Someone should submit a PR handling that...

PR #974, opened 2017-08-11
Added support for centerright and centerleft position

cmodijk:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | TODO If this gets accepted

Added support for `centerright` and `centerleft` position to the BackgroundFilterLoader

PR #979, opened 2017-08-28
Updated the docs for #974

cmodijk:
Updated the docs for #974


cmodijk:
@cedricziel Yes thanks!

Issue #857, opened 2017-01-24
Expose the ability to define custom process timeouts


alexwilson:
As of Symfony 2.4, we have the ability to define a custom timeout for processes created via the ProcessBuilder (currently only PostProcessors).

I am wondering if we should float this value to users, as I've seen in the eZ Community Slack that users are hitting the default of 60 seconds with higher levels of optimisation (in fact the default of 7!) when using optipng.

My thoughts are that either we allow a configuration variable exposing a global limit, or we add to the configuration of each individual post-processor (both general, and run-time configuration).  I lean towards the latter, as this offers more flexibility (i.e. filter `high_quality_notimeout` specifies a timeout of `null` which outright disables the ProcessBuilder timeout).

Does this make sense to implement?

robfrawley:
:+1: The general idea is great, but I would love to see a bit more flexibility by allowing both a global configuration and a per-post-processor overwrite.

robfrawley:
I'm working on a simple implementation that is compatible with the `1.x` branch (and by extension the `2.x` branch). It is actually relatively simple. More information to follow soon, as well as a PR. @antoligy I agree with your initial inclination that this is most useful when it can be configured individually for each post-processor. Currently, the implementation is as follows:

- No global timeout option
- Individual post-processor timeout options
  - When timeout option **is not** set, the `ProcessBuilder::setTimeout()` method is never called, resulting in the process using Symfony's default timeout value.
  - When timeout option **is** set, the `ProcessBuilder::setTimeout()` method is called to assign the timeout value (this includes integer values and `null` values to enable disabling the timeout completely).

Do the above declarations make sense? Anything about the behavior you would change or add/remove?
Are there any additional "features" that either @antoligy or @niponika can think of that would be of benefit, without greatly bloating the size of the changes? Currently, it is relatively small and simple.

For example, we could theoretically add a global timeout option, but the complexity of dealing with it (in combination with the possibility of other per post-processor timeouts that supersede it, doesn't seem worth it to me ATM.

Anyway, let me know. PR soon.

niponika:
For me this sounds like the best thing to do!

As in the current version, the post processor can't be used by any user who has heavy inputs, or low system config (like a cheap VPS).

There is something I have been thinking about, but that isn't really an emergency.
[...]

Issue #987, opened 2017-09-06
Process timeout


niponika:
| Q | A |
| --- | --- |
| Bug Report? | yes |
| Feature Request? | yes |
| BC Break Report? | no |
| RFC? | no |
| Imagine Bundle Version | 1.9.0 |

Hello,
First of all, it's the first time I ever do such a report, so please tell me if I forgot to include any element.
I encountered a problem when applying post processors to large files (OptiPNG & PNGquant specifically) . Symfony throws a ProcessTimedOutException whenever the process exceeds it's timeout value.

I saw that the process was generated using Symfony's ProcessBuilder, however no timeout value is specified, therefore the default value of 60 seconds is applied. However, this can fall a bit short for these post processors when they do heavy compression on large files.

Do you think it could be possible to implement some sort of timeout configuration in the filter's configuration, or simply apply longer default timeouts for the post processors ?

Best regards !

niponika:
Sorry, just saw that an issue was already opened for this. Please link it to #857

robfrawley:
Closing as duplicate, but *please do* continue the discussion in #857.

PR #993, opened 2017-09-13
[Post-Processors] Add timeout option, refactor processors, add tests for all, add/remove/deprecate options

robfrawley:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | #857
| License | MIT
| Doc PR | (docs included in PR)

This is a large pull request, and it is still a WIP, so I'm not going to go through everything just yet, but for the time being, here is an initial list of changes introduced:

- Enable configuring `timeout`, `prefix`, `env`, and `options` on the `ProcessBuilder` per-post-processor config entry.
- Major refactoring of post-processors to both cleanup their implementations and enable easier writing of tests for them:
  - Deprecated `ConfigurablePostProcessorInterface` and it's `processWithConfiguration()` method. Moving forward, usage of `PostProcessorInterface::process()` *without* a second parameter (containing an array of options) will throw a deprecation message advising that in `2.0` the second parameter is required. We shouldn't need two interfaces for post processors: this should allow us to migrate to one for `2.0`.
  - Pull a bunch of common functionality into a new `AbstractPostProcessor` that all our core-post-processors inherit from.
  - Deprecate all setter methods on post-processes: their state should be set via the constructor and their per-usage behavior overridden via the second argument of `process()`. These setters provided no value whatsoever and simply expose an additional public method of our API that we cannot change without a major version release. *(We should find more instances of useless methods that we can deprecate for removal or change of visibility to `protected/private` in `2.0`).*
  - Changes to some of the specific options available via the post-processors (all changes simply throw deprecation warnings and all prior options continue to work as expected until `2.0`).
  - Additions of a bunch of post-processor options that were previously unavailable (see changes to the RST documentation for reference).
- Due to the refactoring, new tests have been added for all post-processors, bringing their code coverage up from `0%` to `100%`.

There are other changes, but those are the broad strokes that should help open an initial discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ matrix:
   allow_failures:
     - php: 7.1
       env: SYMFONY_VERSION=dev-master
+    - php: hhvm
 
 before_install:
   - if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then echo "memory_limit = -1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi;
--- a/Exception/Imagine/Filter/PostProcessor/InvalidOptionException.php
+++ b/Exception/Imagine/Filter/PostProcessor/InvalidOptionException.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Exception\Imagine\Filter\PostProcessor;
+
+use Liip\ImagineBundle\Exception\ExceptionInterface;
+
+class InvalidOptionException extends \RuntimeException implements ExceptionInterface
+{
+    /**
+     * @param string $message
+     * @param array  $options
+     */
+    public function __construct($message, array $options = array())
+    {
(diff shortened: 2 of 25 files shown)


robfrawley:
@antoligy Glad to hear you're onboard with the goals of this PR! :+1: I finally got around to updating the code here to support PHP 5.3 (I always forget I can't rely on things like closures being bound to their current object scope, or short array syntax, etc).

Anyway, tests now pass and Coveralls shows this PR increases code coverage `6.9%` to an overall value of `82.4%` coverage globally for our codebase, with the post-processors now `100%` covered!

See: https://coveralls.io/jobs/29359232

A brief, initial review with any thoughts, comments, or concerns would be welcome from anyone with the time to do so.

lsmith77:
maybe this should go into 2.0 instead now?

also note that ProcessBuild is dropped in Symfony 4
https://github.com/liip/LiipImagineBundle/pull/1025

robfrawley:
Relates: #931

maximgubar:
@robfrawley is this PR still relevant ?

robfrawley:
@maximgubar Yes; I have a local copy that targets `2.0` that should be done soon. I'll update here once it's finished.

FabianSchmick:
Any updates on this? Because I got the same issues as mentioned here https://github.com/liip/LiipImagineBundle/issues/987 and https://github.com/liip/LiipImagineBundle/issues/857

franmomu:
This can be close because of with https://github.com/liip/LiipImagineBundle/pull/1219

dbu:
finished in #1219, thanks @robfrawley and @franmomu !

PR #1015, opened 2017-11-28
[Filter] Add "fixed-size" filter implementation

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #560
| License | MIT
| Doc PR | -

At the moment there is no way to create a fixed-size image.
This is a new filter that does this.

See https://github.com/liip/LiipImagineBundle/issues/560#issuecomment-275694967 for more details.

Usage
-------

```yml
liip_imagine:
    filter_sets:
        my_thumb:
            filters:
                fixed: { width: 120, height: 90 }
```

Issue #560, opened 2015-01-29
allow_upscale on thumbnail filter cant work due to imagine


scuben:
I tried to use the allow_upscale option on a thumbnail filter but the file did not upscale. So i debugged and found this in https://github.com/avalanche123/Imagine/blob/v0.6.1/lib/Imagine/Image/AbstractImage.php:

``` php
// if target width is larger than image width
// AND target height is longer than image height
if ($size->contains($imageSize)) {
    return $thumbnail;
}
```

I do not see any short workaround on that as far as this is code form imagine.

makasim:
I am not sure I get all the details, but as far as I can see the problem comes from imagine library hence it has to be fixed there.

fritzmg:
I also noticed, that `allow_upscale: true` does not actually work, because the Imagine Bundle does not actually allow images to be upscaled. In order for Imagine to always allow upscaling (or any scaling for that matter) the following change would have to be made (I use Gd as an example):

In [lib/Imagine/Gd/Image.php](https://github.com/avalanche123/Imagine/blob/v0.5.0/lib/Imagine/Gd/Image.php) change lines 326 - 358 from
```php
// if target width is larger than image width
// AND target height is longer than image height
if ($size->contains($imageSize)) {
    return $thumbnail;
}

if ($mode === ImageInterface::THUMBNAIL_INSET) {
    $ratio = min($ratios);
} else {
    $ratio = max($ratios);
}

if ($mode === ImageInterface::THUMBNAIL_OUTBOUND) {
    if (!$imageSize->contains($size)) {
        $size = new Box(
            min($imageSize->getWidth(), $size->getWidth()),
            min($imageSize->getHeight(), $size->getHeight())
        );
    } else {
        $imageSize = $thumbnail->getSize()->scale($ratio);
        $thumbnail->resize($imageSize);
    }
    $thumbnail->crop(new Point(
        max(0, round(($imageSize->getWidth() - $size->getWidth()) / 2)),
        max(0, round(($imageSize->getHeight() - $size->getHeight()) / 2))
    ), $size);
} else {
    if (!$imageSize->contains($size)) {
        $imageSize = $imageSize->scale($ratio);
        $thumbnail->resize($imageSize);
    } else {
        $imageSize = $thumbnail->getSize()->scale($ratio);
        $thumbnail->resize($imageSize);
    }
}
```
to
```php
if ($mode === ImageInterface::THUMBNAIL_INSET) {
    $ratio = min($ratios);
[...]

PR #1016, opened 2017-11-28
Upscale Thumbnail is not work

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | MIT
| Doc PR | -

`allow_upscale` is not work in Thumbnail filter #560

I propose to solve this problem by creating a separate filter #1015, because this filter is already [very complex](https://scrutinizer-ci.com/g/liip/LiipImagineBundle/code-structure/master/operation/Liip%5CImagineBundle%5CImagine%5CFilter%5CLoader%5CThumbnailFilterLoader%3A%3Aload).

### DIFF
--- a/Imagine/Filter/Loader/ThumbnailFilterLoader.php
+++ b/Imagine/Filter/Loader/ThumbnailFilterLoader.php
@@ -49,9 +49,7 @@ class ThumbnailFilterLoader implements LoaderInterface
             }
         }
 
-        if (($origWidth > $width || $origHeight > $height)
-            || (!empty($options['allow_upscale']) && ($origWidth !== $width || $origHeight !== $height))
-        ) {
+        if ($origWidth > $width || $origHeight > $height) {
             $filter = new Thumbnail(new Box($width, $height), $mode, $filter);
             $image = $filter->apply($image);
         }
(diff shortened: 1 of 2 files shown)


lsmith77:
@antoligy / @robfrawley how do we want to deal with this .. 1.0 or 2.0 ?

robfrawley:
Since it is a bug, we should likely fix it in both branches, IMHO.

lsmith77:
right, I had a bit of a concern that this might be the sort of bug fix, that is a BC break in that people might be relying on the buggy behavior.

robfrawley:
Ah, I see where you're coming from; I agree it could be considered a BC break and should therefore not be applied to `1.x`.

With that said though, after reacquainting myself with #560 I actually __don't__ believe we should do anything to fix the Thumbnail filter within this project; the issue actually lies in the `imagine-library` implementation. If we did want to resolve this, we should offer a PR to that library.

@lsmith77 Do you think we should work around `imagine-library`'s internal handling ourselves instead? It seems kind of "hacky" to intentionally alter the behavior of our transformation library, but I guess it may be a solution if this is deemed an important "fix".

I think #1015 on the `2.x` branch is the way to move forward.

lsmith77:
would be better to try at least to work with upstream to get the bug fixed

robfrawley:
@peter-gribanov Are you able to work on this with upstream https://github.com/avalanche123/Imagine?

dbu:
has the upstream bug been fixed?

peter-gribanov:
I think we can close this PR in favor #1015

dbu:
cool!

PR #1023, opened 2017-12-10
fix tests on Symfony 3.4/4.0

lsmith77:
drop support for all versions below 3.4

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

I think we went overboard with the functional tests.
At any rate as a result I made lots of services public, that don't need to be public.


robfrawley:
@lsmith77 Sorry; was away for a bit due to holidays and schedule, but I'll be active again if you need a review moving forward. All the best!

lsmith77:
no need to apologize .. but glad to hear you are back to help .. I have been struggling a bit to find time due to the time spend on the diversity initiative but was also hoping to finally setup the Bundle inside a Symfony 4 sandbox to see what issues remain.

PR #1031, opened 2018-01-03
remove unused php-cs-fixer from require-dev

dbu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | see travis-ci
| Fixed tickets | -
| License | MIT
| Doc PR | -

This dependency is not used since we use styleci instead (which is php-cs-fixer as a service). The cs fixer does not install with PHP 7.3, making the PHP nightly build error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,6 @@
         "doctrine/cache": "^1.1",
         "doctrine/orm": "^2.3",
         "enqueue/enqueue-bundle": "^0.7|^0.8",
-        "friendsofphp/php-cs-fixer": "^1.0|^2.0",
         "league/flysystem": "^1.0",
         "psr/log": "^1.0",
         "satooshi/php-coveralls": "^1.0",


dbu:
yay, makes nightly green!

robfrawley:
Why was this removed?

> This dependency is not used since we use styleci instead.

All contributors and/or collaborators rely on this tool to check our work *before* committing changes or (in my case) *requiring* it is run before I accept pull requests. StyleCI is a great check and allows us to express that our code style is sensible via a badge on the `README.md`, but no one uses its interface to apply patches or otherwise interact with the codebase (in my experience here, anyway). It becomes a nightmare to explain to everyone to install the tool (it's already a pain to get them to even *run it*, never mind requiring its installation). We should instead use `--ignore-platform-reqs` on the nightly build.

dbu:
i was not aware that you use it. i never bother and instead rely on styleci. the build had been broken for a while and this was the easiest way to fix it that i found. i'll quickly do a pr to add it back, ignoring the platform for nightly (which seems a good idea anyways).

dbu:
see #1045

PR #1035, opened 2018-01-08
[CI] Extend and enhance Travis build matrix

sebastianblum:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | MIT
| Doc PR | -

Try to extend the travis ci build matrix to fail on skipped tests


.travis.yml
```diff
@@ -21,22 +21,28 @@ matrix:
       env: SYMFONY_VERSION=4.0.*
     - php: 7.2
       env: SYMFONY_VERSION=3.4.*
+    - php: 7.2
+      env:
+        - SYMFONY_VERSION=4.0.*
+        - FLYSYSTEM_VERSION=1.0.*
+        - SIMPLE_PHPUNIT_PARAMETERS=" --stop-on-skipped"
```
dbu: i think this is not doing what you think it does. it stops execution when a skipped test is encountered, but not treat it as failure. see https://travis-ci.org/liip/LiipImagineBundle/jobs/326390446 

it looks like a [--fail-on-skipped](https://github.com/sebastianbergmann/phpunit/issues/2840) is considered as unimportant by sebastian, so we could either maybe do a complicated thing with a test listener to detect skipped tests, or use an environment variable that we define on the line that runs phpunit that we look at in those tests and if its present fail instead of skip.

.travis.yml
```diff
@@ -21,22 +21,28 @@ matrix:
       env: SYMFONY_VERSION=4.0.*
     - php: 7.2
       env: SYMFONY_VERSION=3.4.*
+    - php: 7.2
+      env:
+        - SYMFONY_VERSION=4.0.*
+        - FLYSYSTEM_VERSION=1.0.*
```
dbu: i like the approach we took here: https://github.com/php-http/HttplugBundle/blob/f034b168dde627de027c3d874410b10df2aa82af/.travis.yml#L56 - we define one dependencies variable and if any is set, have composer install those.

.travis.yml
```diff
@@ -21,22 +21,28 @@ matrix:
...
+  allow_failures:
     - env: SYMFONY_VERSION=dev-master
     - php: nightly
-    
+
 before_install:
   - bash ./.travis/exec-before.bash
 
 install:
   - travis_retry composer require "symfony/symfony:${SYMFONY_VERSION}" $COMPOSER_FLAGS
```
dbu: while we are looking at these things, we could use `dunglas/symfony-lock:^2` resp 3 and 4 to do this. the current code will hide missing dependencies on symfony components, as it installs all of symfony. the dunglas package conflicts with everything not in the right version instead, so the CI will show if there is a symfony component being used but not declared in require/require-dev.

sebastianblum:
@dbu yes, I hoped that --stop-on-skipped return an exit code greater than zero.

At the moment I try to add the dependencies for the skipped tests locally.

sebastianblum:
@dbu I fixed all tests. At the moment, Flysystem is included in composer.json so all are running.

I now try to remove the dev-dependency, what do you think?

sebastianblum:
I found a lot of wrong namespaces like here
https://github.com/liip/LiipImagineBundle/blob/2.0/Tests/Imagine/Filter/Loader/AutoRotateFilterLoaderTest.php

@dbu what do you think? can we run php-cs-fixer on the 2.0 branch?

sebastianblum:
@dbu at the moment, it is included in require-dev see https://github.com/liip/LiipImagineBundle/blob/2.0/composer.json

we have two possibilities:
- test case with composer update --no-dev and manually install simple-phpunit to see skipped tests
- leave it as it is

what would you prefer?

sebastianblum:
no @dbu, it is not ready at the moment. please give me one week more :)

sebastianblum:
@dbu now it looks better

composer plugin prestissimo:
aktiv: https://travis-ci.org/liip/LiipImagineBundle/builds/328729635
inaktiv: https://travis-ci.org/liip/LiipImagineBundle/builds/328712094
I can't see any advantage on travis. Maybe it is already included.
Should we remove it? what do you think?

build matrix:
see https://travis-ci.org/liip/LiipImagineBundle/builds/328729635

- 1. tests only the decencies in composer
- 2. tests with symfony/symfony:3.4.*
- 3. tests with symfony/symfony:4.0.*
- ...
- 8. tests without dev dependencies. Tests here should be skipped and not fail
- 9. only execute code coverage once and may fail if it takes too long


In my opinion it is ready for feedback.

robfrawley:
@sebastianblum We intentionally limited the scope of the build matrix for `2.x` after the difficulties we've had with the `1.x` branch, mostly regarding the total time.

This PR almost doubles the previous total time of `10` minutes to `17`. We need to tighten the scope of the build matrix in this PR and shoot for getting our times more in-line with what it was.

Remember, there are lots more Symfony versions to come and we like to test all supported minor releases, so the matrix will naturally become larger later; we will need headroom so that can happen without getting to the 1 hour 16 minute total time it takes for the `1.x` series to complete its CI build. :-)

sebastianblum:
@robfrawley good point. I removed two tests, they are not necessary in my opinion.

PHP 7.1 and PHP 7.2 without (symfony/symfony in a fixed version) are the most important tests.
Then the tests with fixed symfony/symfony version can run only in one php version.
The composer flags --prefer-lowest should also run in my opinion.
Also the check, that die require-dev are really optional.

PHP coverage runs now only once, also php 7.3 check.

Do you have more feedback?

robfrawley:
@sebastianblum Yeah; that looks better and brought us within two minutes of the current 2.x branch. As for any other feedback, I only really glanced at the build matrix and travis times; I'll take a broader look at the other changes later today and let you know if I notice anything. Thanks for the PR!

sebastianblum:
@robfrawley After this pr is merged, I will add another check for php-cs-fixer to the Build Matrix in pr #1040

I can add it to an existing test or create a new one. What would you suggest?

.travis.yml
```diff
@@ -9,34 +9,55 @@ cache:
...
+        - STABILITY=dev
     - php: nightly
-    
+
 before_install:
-  - bash ./.travis/exec-before.bash
+  - if [[ "$SYMFONY_VERSION" != "" ]]; then travis_retry composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update $COMPOSER_FLAGS; fi;
+  - if [[ "$DEPENDENCIES" != "" ]]; then travis_retry composer require ${DEPENDENCIES} --no-update $COMPOSER_FLAGS; fi;
+  - if [[ "$STABILITY" != "" ]]; then composer config minimum-stability $STABILITY; fi
+  - if [[ "$ENABLE_CODE_COVERAGE" != "true" && "$TRAVIS_EVENT_TYPE" != "cron" ]]; then rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini; fi;
```
robfrawley: Instead of `rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini` you should use the `phpenv` configuration plugin instead; see ["Preinstalled PHP extensions" (blog.travis-ci.com)](https://blog.travis-ci.com/2013-03-08-preinstalled-php-extensions).

```
phpenv config-rm xdebug.ini
```

robfrawley:
Thanks @sebastianblum!

PR #1039, opened 2018-01-20
change the default from web to public

lsmith77:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | depends :)
| New feature? | depends :)
| BC breaks? | yes
| Deprecations? | no
| Tests pass? | 
| Fixed tickets | 
| License | MIT
| Doc PR | included

default should now be `public` but wondering if it in fact should be `"%kernel.project_dir%/public"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 /composer.lock
 /composer.phar
 /vendor/
-/Tests/Functional/app/web/media/cache
+/Tests/Functional/app/public/media/cache
 /.idea/
 /var/
--- a/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
+++ b/DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
@@ -59,7 +59,7 @@ class FileSystemLoaderFactory extends AbstractLoaderFactory
                     ->end()
                     ->treatNullLike([])
                     ->treatFalseLike([])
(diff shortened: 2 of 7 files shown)


robfrawley:
If we intend to support Symfony `3.4` (and future 3.x releases) we cannot make this change outright; `web` is still the correct directory for the 3.x series. Instead, we need to _keep_ the current value **and** _add_ the new one pointing to `public`.

In the context of the `FileSystem` loader this can be done easily as it already accepts an array, but for the `WebPath` resolver, we need to update the `web_root` configuration option to accept an array and update the class(es) that use/rely on that value to handle such a type. Alternatively, for the `WebPath` resolver, we can make the default value dynamic based on the Symfony kernel version, but I generally hate version checks within any non-test code, IMHO.

As for the new parameters available, assuming `%kernel.project_dir%` is included in `3.4` (which I believe it is if I'm not mistaken) I agree that we should use it moving forward instead of the current usage.

lsmith77:
well for 2.x we can choose to make a new default. where possible it would be best of course to automatically determine the default. but in this case I would do it inside the DI extension instead of setting multiple dirs which could affect performance.

silverbackdan:
Personally, I'd like the new default and as suggested, perhaps in the DI a simple directory check to determine this possibly, although it may not be necessary with the major version change.

robfrawley:
How does #1044 look?

PR #1040, opened 2018-01-20
[CS] Implement new php-cs-fixer rules and related code cleanup

sebastianblum:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | -
| Fixed tickets | -
| License | MIT
| Doc PR | -

While fixing the tests in #1035 I saw a lot of wrong namespaces. Trying to run php-cs-fixer returns an error, so I updated the php-cs-fixer configuration.

The main rules are coming directly vom symfony/demo, https://github.com/symfony/demo/blob/master/.php_cs.dist

The question is, what merge strategy from 1.x to 2.x will be chosen and which rules make sense.

After the have defined the rules, I will add the changes to the php files and try to fix the scrutiniser checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
+/.idea/
+/.php_cs.cache
 /bin/
-/phpunit.xml
 /composer.lock
 /composer.phar
-/vendor/
-/Tests/Functional/app/web/media/cache
+/phpunit.xml
 /Tests/Functional/app/public/media/cache
-/.idea/
+/Tests/Functional/app/web/media/cache
 /var/
+/vendor/
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,47 +1,86 @@
 <?php
 
-$fileHeaderComment = <<<COMMENT
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+use PhpCsFixer\Config;
+use PhpCsFixer\Finder;
+
+$header = <<<HEADER
 This file is part of the `liip/LiipImagineBundle` project.
 
 (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
 
 For the full copyright and license information, please view the LICENSE.md
 file that was distributed with this source code.
-COMMENT;
+HEADER;
 
-$finder = PhpCsFixer\Finder::create()
-    ->in(__DIR__)
-    ->ignoreDotFiles(true)
-    ->ignoreVCS(true)
-    ->exclude(array('build', 'vendor'))
-    ->files()
-    ->name('*.php')
-;
-
-return PhpCsFixer\Config::create()
-    ->setFinder($finder)
+return Config::create()
+    ->setUsingCache(true)
     ->setRiskyAllowed(true)
+    ->setFinder((new Finder())->in(__DIR__))
     ->setRules([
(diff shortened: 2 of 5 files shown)


sebastianblum:
we should discuss the use of `native_function_invocation` to speed up the project. Since werequire php 7.1, I would enable this.

sebastianblum:
I see a lot commits adding the backslash for performance reasons: https://github.com/symfony/symfony/pull/25854

robfrawley:
My vote for `native_function_invocation` is :-1: as it reduces readability for the sake of (negligible) speed.

dbu:
php-cs-fixer is not currently installed in the 2.0 branch. i add it back in #1045

should we run the php-cs-fixer in travis-ci? if not, we won't see whether the configuration for styleci and php-cs-fixer conflict...

sebastianblum:
@dbu I will add it after the new build matrix is merged, see #1035 

We can add it in require-dev in composer.json and remove it for php 7.3 / nightly

sebastianblum:
PHP-CS-Fixer is now integrated in travis-ci

here are the results https://travis-ci.org/liip/LiipImagineBundle/jobs/336491426

dbu:
interesting. i think we miss a couple styleci rules, notably [yoda](https://styleci.readme.io/docs/fixers#section-yoda_style). in the end, if php-cs-fixer is slightly stricter than styleci, i think its acceptable, but those things we can we should also put into styleci so that automatic fixing is available. or we drop styleci completely and have only fixing by the user in their cli as an option.

robfrawley:
@dbu This is a new set of styles (compared to our old rule set this one adds a number of additional fixers), so there are some drastic changes this will apply outside the current `styleci` rules, but in the past, the two services _have_ been aligned. Moving forward, I agree we need to update `styleci` to reflect this new ruleset, assuming everyone signs off on these changes. I don't think the solution is to remove `styleci` completely, as it's great to have the badge on our readme to show our project complies with respectable code style standards at first glance.

@sebastianblum You _should_ add the yoda rule, assuming `phpcsfixer` has a compatible fixer (I believe it does).

sebastianblum:
@robfrawley Yoda style is included in @Symfony already.

I created this pull request to create the php-cs-fixer configuration and if @robfrawley and @dbu agree, I will create a new pull request optimizing the code.
In my plan we do these steps:

1. define rule set -  #1040 
2. optimize code -  #1046 
3. apply coding style - TODO
4. activate php-cs-fixer in travis configuration - #1040 

But we have to discuss how to merge new features from 1.0 to 2.0. Do we have a timeline for supporting 1.0?

robfrawley:
@sebastianblum About a week ago @lsmith77 and I spoke briefly about support for `1.0` and my understanding is that `1.0` will go into **bug-fix only mode** the moment the first `2.0` release is tagged. Before we can get this PR merged, can you separate the `php-cs-fixer` Travis addition into its own PR?

sebastianblum:
Good, I created #1049 for the future. Should we do the style fixing in this pull request or in a new one?

dbu:
the final call is up to you @sebastianblum and @robfrawley but the style rules seem all sensible to me.

if you add the rules in version 1 too, its probably easier to merge 1.0 to 2.0 when you want to merge bugfixes upstream.

robfrawley:
I'll handle merging this, running the new styles, and updating styleci in one at the same time. I think your work is done! I'll get this done shortly.

robfrawley:
@dbu Good call about applying them to 1.0 to avoid merge conflicts. Will handle that too then.

PR #1043, opened 2018-01-29
[Async] [DI] Make resolve_cache_processor a public service

silverbackdan:
liip_imagine.async.resolve_cache_processor should be public or an exception is thrown when sending enqueue command.

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | 
| Fixed tickets | 
| License | MIT
| Doc PR | 

I had an exception about this service needing to be public using SF4. All services are private by default now. Not sure where it was used and if dependency injection through the __construct method would be preferred but wanted to submit this PR as a quick easy fix which shouldn't result in any breaking changes with older symfony versions either.

### DIFF
--- a/Resources/config/enqueue.xml
+++ b/Resources/config/enqueue.xml
@@ -4,7 +4,7 @@
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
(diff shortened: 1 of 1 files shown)


robfrawley:
@makasim ping

makasim:
This how it should be, at least for now since we still support SF2,3,4

PR #1053, opened 2018-02-15
Update imagine.xml

hjanuschka:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | hope so
| License | MIT

we are in the process to upgrade to symfony to 3.4 (and later on to 4.0).
we stumbled upon a few deprecations with this bundle. services are by default now `non-public`
this PR makes all "not already explicitly defined" services public. 

edit: it also fixes ProcessBuilder deprecation, for the sake of it, as it is not a really big deal, i just added the processbuilder locally, instead of a heavy refactor

### DIFF
--- a/Compat/ProcessBuilder.php
+++ b/Compat/ProcessBuilder.php
@@ -1,0 +1,244 @@
+<?php
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace Liip\ImagineBundle\Compat;
+use Symfony\Component\Process\ProcessUtils;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Process\Exception\InvalidArgumentException;
+use Symfony\Component\Process\Exception\LogicException;
+//Compat layer for deprecated, removed later, ProcessBuilder
+/**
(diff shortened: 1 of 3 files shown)


robfrawley:
The `1.x` branch is not compatible and will not be upgraded to support Symfony `4.x`. Please use the `2.x` branch for Symfony `4.x` compatibility and see https://github.com/liip/LiipImagineBundle/milestone/6 for additional information about this upcoming stable release.

robfrawley:
@hjanuschka I appreciate the drive to work on this bundle and if you would like to submit changed to the `2.0` branch that further support for Symfony `4.x` we would gratefully include it. As for this contribution, we don't want to support Symfony `4.x` on the `1.x` branch.

hjanuschka:
@robfrawley many thx for your reply, i dont have a problem with rejecting the PR.
using the  `2.0` branch, enables us to upgrade without any deprecation notes to symfony 3.4.
by just changing `routing.xml` to `routing.yml` 🎉 once we finished our 3.3 to 3.4 upgrade, we immediately start creating a 3.4 to 4.0 branch, and if we stumble upon any issues with this or any other bundle we will do PR's to not run from private forks for too long. 👍 

our biggest pain point right now is assetic, even if it its deprecated, we already forked and ported it - to work with 3.4/4.0, so we can take a bit more time on getting away from assetic.

PR #1098, opened 2018-05-23
[Filters] [Config] [DI] Add Filter configuration class as public service

maximgubar:
| Q | A
| --- | ---
| Branch? | master 
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

In this PR I propose to introduce a service, which retrieves filter_sets, filters configuration and exposes it via interfaces.


robfrawley:
Overalls, this looks great. I do have some ideas I've been wanting to implement regarding filter-sets/filter that would be good to bring up now. The number of times people confuse the difference between filter-sets and filters is countless, I submit we should change the terminology of filter-set to "instructions" or something different. Thoughts? Other naming ideas?

robfrawley:
Also, with one of the PRs I'm converting from `1.x` to `2.0` I actually created a "configuration" class as well (instead of adding tons of constructor arguments), so perhaps we should move everything in this PR from `Config/` to `Config/Filter` namespace so we can define other configs for other components under the same `Config/` namespace without having tons of unrelated config classes together in the same namespace level.

Config/FilterInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+interface FilterInterface
```
dbu: i would like to do a phpdoc that explains what a filter is (the definition for one image transformation operation)

Config/FilterInterface.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @return array
+     */
+    public function getOptions(): array;
+
+    /**
+     * @param string $name
+     */
+    public function setName(string $name): void;
```
dbu: imo the interface should not define setters. the whole definition is a read-only thing.

Config/FilterSet.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function getDataLoader(): string
+    {
+        return $this->dataLoader;
+    }
+
+    /**
+     * @return int
+     */
+    public function getQuality(): int
```
dbu: this should have a phpdoc explaining what it is. (resp, as rob says the jpg quality and the png compression level values should have that doc, to explain the value range)

Config/Filter.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param string $name
+     */
+    public function setName(string $name): void
```
dbu: i'd rather use constructor arguments than setters

Config/FilterInterface.php
maximgubar: completely agree with you :)

Config/Filter.php
maximgubar: got it, will refactor it.

Config/Filter/Type/Thumbnail.php
```diff
@@ -0,0 +1,103 @@
...
+    private $allowUpscale;
+
+    /**
+     * @var string
+     */
+    private $filter;
+
+    /**
+     * @param string      $name
+     * @param array       $size
```
dbu: can you specify what kind of array this is? is it a hashmap with width and height?

Factory/Config/FilterFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Liip\ImagineBundle\Factory\Config;
+
+use Liip\ImagineBundle\Config\FilterInterface;
+
+interface FilterFactoryInterface
+{
+    /**
+     * @return string
```
dbu: can we say that the name must match the Filter::getName that this factory handles?

Config/FilterFactoryCollection.php
```diff
@@ -0,0 +1,59 @@
...
+
+final class FilterFactoryCollection implements FilterFactoryCollectionInterface
+{
+    /**
+     * @var FilterFactoryInterface[]
+     */
+    private $filterFactories;
+
+    /**
+     * FilterCollection constructor.
```
dbu: i'd remove docblocks that don't add any information over whats in the php code

Config/FilterFactoryCollection.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param FilterFactoryInterface ...$filterFactories
+     */
+    public function __construct(FilterFactoryInterface ...$filterFactories)
+    {
+        $this->filterFactories = $filterFactories;
```
dbu: imho we should convert this to a hashmap with name => factory so that we don't have to loop over the array in getFilterFactoryByName

Config/FilterFactoryCollectionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+use Liip\ImagineBundle\Factory\Config\FilterFactoryInterface;
+
+interface FilterFactoryCollectionInterface
```
dbu: imho there is only one way how this can be implemented and hence no point in having an interface

Config/FilterSet.php
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+use Liip\ImagineBundle\Exception\InvalidArgumentException;
+
+final class FilterSet implements FilterSetInterface
```
dbu: as discussed, rename this to `Stack`

Resources/config/imagine.xml
```diff
@@ -32,6 +32,39 @@
...
+        <service id="Liip\ImagineBundle\Factory\Config\FilterSetFactory" alias="liip_imagine.factory.config.filter_set"/>
+
+        <service id="liip_imagine.factory.config.filter.crop" class="Liip\ImagineBundle\Factory\Config\Filter\CropFactory" />
+        <service id="Liip\ImagineBundle\Factory\Config\Filter\CropFactory" alias="liip_imagine.factory.config.filter.crop"/>
+
+        <service id="liip_imagine.factory.config.filter.strip" class="Liip\ImagineBundle\Factory\Config\Filter\StripFactory" />
+        <service id="Liip\ImagineBundle\Factory\Config\Filter\StripFactory" alias="liip_imagine.factory.config.filter.strip"/>
+
+        <service id="liip_imagine.factory.config.filter.thumbnail" class="Liip\ImagineBundle\Factory\Config\Filter\ThumbnailFactory" />
+        <service id="Liip\ImagineBundle\Factory\Config\Filter\ThumbnailFactory" alias="liip_imagine.factory.config.filter.thumbnail"/>
```
dbu: i don't think that we need an autowiring capable alias for these. they should be private services not useable outside of DI configuration

Resources/config/imagine.xml
```diff
@@ -32,6 +32,39 @@
...
+
+        <!-- Config services -->
+        <service id="liip_imagine.config.filter_factory_collection" class="Liip\ImagineBundle\Config\FilterFactoryCollection">
+            <argument type="service" id="liip_imagine.factory.config.filter.crop"/>
+            <argument type="service" id="liip_imagine.factory.config.filter.strip"/>
+            <argument type="service" id="liip_imagine.factory.config.filter.thumbnail"/>
+        </service>
+        <service id="Liip\ImagineBundle\Config\FilterFactoryCollection" alias="liip_imagine.config.filter_factory_collection"/>
```
dbu: neither do we need autowiring for this. the only thing that should be available for autowiring should be the actual configuration object.

Config/FilterSetCollection.php
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * @var FilterSetBuilderInterface
+     */
+    private $filterSetBuilder;
+
+    /**
+     * @param FilterConfiguration       $filterConfiguration
+     * @param FilterSetBuilderInterface $filterSetBuilder
+     */
+    public function __construct(FilterConfiguration $filterConfiguration, FilterSetBuilderInterface $filterSetBuilder)
```
dbu: does it make sense to use the FilterConfiguration object? looks to me like that is a previous attempt at this problem that did not go far enough? it smells like we should deprecate FilterConfiguration and pass the plain array here, or not?

Config/FilterSetCollection.php
```diff
@@ -0,0 +1,58 @@
...
+    private $filterConfiguration;
+
+    /**
+     * @var FilterSetBuilderInterface
+     */
+    private $filterSetBuilder;
+
+    /**
+     * @param FilterConfiguration       $filterConfiguration
+     * @param FilterSetBuilderInterface $filterSetBuilder
```
dbu: having factories for so many things, why do we not have a factory for the StackCollection? we could then define the public service for this as a thing that uses the StackCollectionFactory, and not have the value object that is used outside have the factory in it. the FilterSetBuilder could take that role, to not explode the number of infrastructure classes that we use.

for the symfony config, it would look like this: https://symfony.com/doc/current/service_container/factories.html

Config/Filter.php
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+final class Filter implements FilterInterface
```
dbu: once you implemented all filters, this will become obsolete, right?

maximgubar: done

Config/FilterFactoryCollection.php
maximgubar: done

Factory/Config/FilterFactoryInterface.php
maximgubar: done

Config/FilterFactoryCollection.php
maximgubar: done

Config/FilterFactoryCollectionInterface.php
maximgubar: done

Resources/config/imagine.xml
maximgubar: done

Resources/config/imagine.xml
maximgubar: done

Config/FilterSetCollection.php
maximgubar: done

Config/FilterInterface.php
maximgubar: @dbu could you provide an example ? cause I'm not sure that I understand this comment.

Resources/config/imagine.xml
maximgubar: @dbu, btw, do we need to have SF4 compatibility ?

Config/FilterInterface.php
dbu:
```
/**
 * A filter contains the configuration for an image transformation operation.
 *
 * The type of operation is defined by the filter class. The operation parameters are specific to
 * the filter.
 */
```

Resources/config/imagine.xml
dbu: it would be nice, but i would not mix it into this pull request.

if we have new services that should be autowire-enabled, we can already define the right alias here, but i would not do anything outside the scope of modeling the configuration.

Config/FilterInterface.php
maximgubar: done

Config/Filter/Type/Crop.php
```diff
@@ -0,0 +1,68 @@
...
+    private $start;
+
+    /**
+     * @var array
+     */
+    private $size;
+
+    /**
+     * @param string $name
+     * @param array  $start start coordinates {x,y}
```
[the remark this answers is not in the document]
maximgubar: good idea, will add it

Config/FilterFactoryCollection.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * @var FilterFactoryInterface[]
+     */
+    private $filterFactories = [];
+
+    /**
+     * @param FilterFactoryInterface ...$filterFactories
+     */
+    public function __construct(FilterFactoryInterface ...$filterFactories)
```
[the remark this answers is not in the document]
maximgubar: got your point, but how usage of array and additional checks of type is better ?
since this object actually plays a role of collection and it is constructed with FilterFactoryInterface(s) why just not declare it directly ?

Config/Filter/Type/Background.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->color = $color;
+        $this->transparency = $transparency;
+        $this->position = $position;
+        $this->size = $size;
+    }
+
+    /**
+     * @return string
+     */
+    public function getName(): string
```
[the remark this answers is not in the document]
maximgubar: thanks for the hint, will refactor.

Config/FilterFactoryCollection.php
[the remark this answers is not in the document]
dbu: as you loop over the parameter anyways, you could also do an instanceof check, and have the `@param FilterFactoryInterface[] $filterFactories` to allow code sniffers to detect errors. 

if we ever need to add an (optional) constructor argument, the current form will be a problem as we can't add anything else without a BC break.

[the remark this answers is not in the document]
maximgubar: why would we ever need to to add something to constructor of this class ?
the purpose of this object is to store FilterFactoryInterfaces only and thats it. it shouldn't contain any logic. tbh, in this case specifically, I don't see advantages of array implementation.

Config/Filter/Type/Watermark.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @param string     $name
+     * @param string     $image
+     * @param string     $position
+     * @param float|null $size
+     */
+    public function __construct(string $name, string $image, string $position, float $size = null)
+    {
+        $this->name = $name;
```
dbu: is the name variable? should this not be a fixed string 'watermark'?

Config/Stack.php
```diff
@@ -0,0 +1,96 @@
...
+    private $quality;
+
+    /**
+     * @var FilterInterface[]
+     */
+    private $filters = [];
+
+    /**
+     * @param string            $name
+     * @param string|null       $dataLoader
```
dbu: what is the data loader?

Config/StackBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Liip\ImagineBundle\Config;
+
+use Liip\ImagineBundle\Factory\Config\StackFactory;
+
+final class StackBuilder implements StackBuilderInterface
+{
+    /**
+     * @var StackFactory
+     */
+    private $filterSetFactory;
```
dbu: please rename to stackFactory (applies to all `filterSet` mentions in this class)

Config/StackCollection.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+final class StackCollection
+{
+    /**
+     * @var array
+     */
+    private $filterSets = [];
```
dbu: please rename to $stacks

Config/StackCollection.php
```diff
@@ -0,0 +1,56 @@
...
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config;
+
+final class StackCollection
+{
+    /**
+     * @var array
```
dbu: i'd use `Stack[]` here instead of `array`

Factory/Config/Filter/AutoRotateFactory.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Liip\ImagineBundle\Factory\Config\Filter;
+
+use Liip\ImagineBundle\Config\Filter\Type\AutoRotate;
+use Liip\ImagineBundle\Config\FilterInterface;
+use Liip\ImagineBundle\Factory\Config\FilterFactoryInterface;
+
+final class AutoRotateFactory implements FilterFactoryInterface
+{
+    const NAME = 'auto_rotate';
```
dbu: as we are aiming for php 7.1, we should also qualify the visibility of these constants. i guess they should be public.

and imho, they should be in the Filter class, not the factory. and remove the $name constructor argument in favor of using the constant.

Factory/Config/FilterFactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace Liip\ImagineBundle\Factory\Config;
+
+use Liip\ImagineBundle\Config\FilterInterface;
+
+interface FilterFactoryInterface
+{
+    /**
+     * Filter name
```
dbu: Name of the filter that this factory can create.

Factory/Config/Filter/BackgroundFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return self::NAME;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $options): FilterInterface
+    {
+        $color = isset($options['color']) ? $options['color'] : null;
```
dbu: the most condensed for these checks would be `$color = $options['color'] ?? null;`

Factory/Config/Filter/InterlaceFactory.php
```diff
@@ -0,0 +1,43 @@
...
+        return self::NAME;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $options): FilterInterface
+    {
+        $mode = ImageInterface::INTERLACE_LINE;
+        if (!empty($options['mode'])) {
```
dbu: why empty and not isset? i'd go with the pattern `$mode = $options['mode'] ?? ImageInterface::INTERLACE_LINE;`

Factory/Config/Filter/ResizeFactory.php
```diff
@@ -0,0 +1,42 @@
...
+        return self::NAME;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $options): FilterInterface
+    {
+        $size = [];
+        if (isset($options['size']) && is_array($options['size'])) {
```
dbu: this is silently ignoring non-array size options, which is a problem imho. we should do the `$size = $options['size'] ?? [];` . the constructor type declaration on $size will detect if the size option has been set to an invalid value.

Factory/Config/StackFactory.php
```diff
@@ -0,0 +1,30 @@
...
+
+use Liip\ImagineBundle\Config\Stack;
+
+class StackFactory
+{
+    /**
+     * @param string      $name
+     * @param string|null $dataLoader
+     * @param int|null    $quality
+     * @param array       $filters
```
dbu: FilterInterface[]

Factory/Config/StackFactory.php
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @return Stack
+     */
+    public function create(string $name, string $dataLoader = null, int $quality = null, array $filters)
+    {
+        return new Stack($name, $dataLoader, $quality, $filters);
```
dbu: also, is there value to use a factory for this over just doing `new Stack()` in the place where we need the stack? feels to me like the factory is too much abstraction.

Factory/Config/Filter/BackgroundFactory.php
maximgubar: is it a typo ?
should it be like this ?
`$color = $options['color'] ?: null;`

Factory/Config/StackFactory.php
maximgubar: using `new` is breaking DI principle. it makes it hardcoded and non-extendable and not testable.

Factory/Config/Filter/BackgroundFactory.php
dbu: no. that will generate a warning when the index 'color' is not set. ?? is the [null coalesce operator](http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce) introduced in PHP 7 to shorten exactly the construct you used above.

Factory/Config/StackFactory.php
dbu: i don't see the problem. we don't want a mock stack anywhere, its a value object. to me it looks overengineered with this. we do not need to DI what kind of stack is created as its just a value object. but its not a huge deal either.

maximgubar: we already mock it in our unit tests

dbu: i don't see the need for mocking the Stack object in a unit test, its straightforward enough imho. yes, its not a perfectly atomic unit test, but good enough as all of the build process is just infrastructure code to build the configuration object.

Config/FilterFactoryCollection.php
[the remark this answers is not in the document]
maximgubar: all factories are declared as `final`, does it sill make sense to mark them additionally with `@internal` ?

Config/Filter/Type/Background.php
```diff
@@ -43,9 +43,9 @@
 
     /**
      * @param string      $name
-     * @param string|null $color
-     * @param string|null $transparency
-     * @param string|null $position
+     * @param string|null $color        background color HEX value
+     * @param string|null $transparency possible values 0..100
+     * @param string|null $position     position of the input image on the newly created background image. Valid values: topleft, top, topright, left, center, right, bottomleft, bottom, and bottomright
```
Koc: Maybe we can use value objects/enum here also.

Config/FilterFactoryCollection.php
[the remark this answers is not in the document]
dbu: yes, because `final` means you can't extend the class, but `@internal` means you should not interact with it at all, e.g. instantiate it in your code and so something with it.

Config/Filter/Type/Background.php
maximgubar: @Koc for all parameters or for specific one ?

Config/Filter/Argument/Size.php
```diff
@@ -0,0 +1,41 @@
...
+     * @var int
+     */
+    private $width;
+
+    /**
+     * @var int
+     */
+    private $height;
+
+    public function __construct(int $width = null, int $height = null)
```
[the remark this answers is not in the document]
maximgubar: @Koc there are cases, when one of the dimensions could be `null`, how would you suggest to implement it if this approach causes doubts ?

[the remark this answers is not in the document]
dbu: Lets add a phpdoc to this method? "To allow keeping aspect ratio, it is allowed to only specify one of width or height. It is however not allowed to specify neither dimension."

Config/Filter/Type/AutoRotate.php
```diff
@@ -0,0 +1,35 @@
...
+
+final class AutoRotate implements FilterInterface
+{
+    /**
+     * @var string
+     */
+    private $name;
+
+    /**
+     * @param string $name
```
dbu: lets remove this, its redundant

Config/Filter/Type/Interlace.php
```diff
@@ -0,0 +1,53 @@
...
+    private $name;
+
+    /**
+     * @var string
+     */
+    private $mode;
+
+    /**
+     * @param string $name
+     * @param string $mode
```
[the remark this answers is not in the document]
dbu: ^

Config/Filter/Type/RelativeResize.php
```diff
@@ -0,0 +1,88 @@
...
+     * @var float
+     */
+    private $scale;
+
+    /**
+     * @param string     $name
+     * @param float|null $heighten
+     * @param float|null $widen
+     * @param float|null $increase
+     * @param float|null $scale
```
dbu: please remove this phpdoc as it does not add anything

Config/Filter/Type/Strip.php
```diff
@@ -0,0 +1,35 @@
...
+
+final class Strip implements FilterInterface
+{
+    /**
+     * @var string
+     */
+    private $name;
+
+    /**
+     * @param string $name
```
dbu: please remove

Config/Filter/Type/Thumbnail.php
```diff
@@ -0,0 +1,89 @@
...
+     * @var string
+     */
+    private $filter;
+
+    /**
+     * @param string      $name
+     * @param Size        $size
+     * @param string|null $mode
+     * @param bool|null   $allowUpscale
+     * @param string|null $filter
```
dbu: please remove the docblock as it contains no information

Config/Filter/Type/Watermark.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @var float
+     */
+    private $size;
+
+    /**
+     * @param string     $name
+     * @param string     $image
+     * @param string     $position
+     * @param float|null $size
```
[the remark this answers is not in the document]
dbu: ^

Config/Filter/Type/Watermark.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @var float
+     */
+    private $size;
+
+    /**
+     * @param string     $name
+     * @param string     $image
+     * @param string     $position
```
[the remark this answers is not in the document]
dbu: ^

Config/Filter/Type/Watermark.php
```diff
@@ -0,0 +1,83 @@
...
+    private $position;
+
+    /**
+     * @var float
+     */
+    private $size;
+
+    /**
+     * @param string     $name
+     * @param string     $image
```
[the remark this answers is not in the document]
dbu: ^

Config/FilterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * A filter contains the configuration for an image transformation operation.
+ *
+ * The type of operation is defined by the filter class. The operation parameters are specific to
+ * the filter.
+ */
+interface FilterInterface
+{
+    /**
+     * @return string
```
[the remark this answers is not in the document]
dbu: ^

Config/Stack.php
dbu: ^

Config/StackBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @var FilterFactoryCollection
+     */
+    private $filterFactoryCollection;
+
+    /**
+     * @param StackFactoryInterface   $filterSetFactory
+     * @param FilterFactoryCollection $filterFactoryCollection
```
dbu: please remove, this is all visible in the method signature

Config/StackBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $this->filterSetFactory = $filterSetFactory;
+        $this->filterFactoryCollection = $filterFactoryCollection;
+    }
+
+    /**
+     * @param string $filterSetName
+     * @param array  $filterSetData
+     *
+     * @return StackInterface
```
dbu: please remove, this is all visible in the method signature

Config/StackCollection.php
dbu: ^

Config/StackCollection.php
dbu: ^

Config/StackCollection.php
```diff
@@ -0,0 +1,56 @@
...
+    private $stackBuilder;
+
+    /**
+     * @var array
+     */
+    private $filtersConfiguration;
+
+    /**
+     * @param StackBuilderInterface $stackBuilder
+     * @param array                 $filtersConfiguration
```
dbu: please remove, its all in the method signature

Config/StackInterface.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @return string
+     */
+    public function getName(): string;
+
+    /**
+     * @return string|null
+     */
+    public function getDataLoader();
```
dbu: as we are on php 7.1 or better, please use the return type declaration `getDataLoader(): ?string` and remove the phpdoc. same for getQuality

Config/Filter/Type/Background.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Liip\ImagineBundle\Config\Filter\Type;
+
+use Liip\ImagineBundle\Config\Filter\Argument\Size;
+use Liip\ImagineBundle\Config\FilterInterface;
+
+/**
+ * @codeCoverageIgnore
+ */
+final class Background extends FilterAbstract implements FilterInterface
```
dbu: you can remove the `implements FilterInterface` because thats already given by extending `FilterAbstract`

Config/StackBuilderInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Liip\ImagineBundle\Config;
+
+interface StackBuilderInterface
+{
+    /**
+     * @param string $filterSetName
+     * @param array  $filterSetData
+     *
+     * @return StackInterface
```
dbu: this does not add anything over the phpdoc, lets remove it

Factory/Config/FilterFactoryInterface.php
dbu: ^

Factory/Config/StackFactoryInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+interface StackFactoryInterface
+{
+    /**
+     * @param string            $name
+     * @param string|null       $dataLoader
+     * @param int|null          $quality
+     * @param FilterInterface[] $filters
+     *
+     * @return Stack
```
dbu: could we instead do the return type declaration on the method itself?

Config/Filter/Type/Background.php
maximgubar: done

Config/Stack.php
maximgubar: done

Config/StackBuilderInterface.php
maximgubar: done

Factory/Config/FilterFactoryInterface.php
maximgubar: done

Factory/Config/StackFactoryInterface.php
maximgubar: done

dbu:
@robfrawley i would prefer to have some final feedback from you before we merge this. do you have time to look at this soonish?

Binary/Locator/FileSystemInsecureLocator.php
```diff
@@ -21,9 +21,9 @@ class FileSystemInsecureLocator extends FileSystemLocator
      */
     protected function generateAbsolutePath(string $root, string $path): ?string
     {
-        if (false === mb_strpos($path, '..'.DIRECTORY_SEPARATOR) &&
-            false === mb_strpos($path, DIRECTORY_SEPARATOR.'..') &&
-            false !== file_exists($absolute = $root.DIRECTORY_SEPARATOR.$path)
+        if (false === mb_strpos($path, '..'.\DIRECTORY_SEPARATOR) &&
```
lsmith77: why the backslashes here?

Binary/Locator/FileSystemLocator.php
```diff
@@ -60,7 +60,7 @@ public function locate(string $path): string
      */
     protected function generateAbsolutePath(string $root, string $path): ?string
     {
-        if (false !== $absolute = realpath($root.DIRECTORY_SEPARATOR.$path)) {
+        if (false !== $absolute = realpath($root.\DIRECTORY_SEPARATOR.$path)) {
```
lsmith77: why the backslashes here?

DependencyInjection/Factory/Loader/FileSystemLoaderFactory.php
```diff
@@ -127,7 +127,7 @@ private function getBundleResourcePaths(ContainerBuilder $container)
         }
 
         return array_map(function ($path) {
-            return $path.DIRECTORY_SEPARATOR.'Resources'.DIRECTORY_SEPARATOR.'public';
+            return $path.\DIRECTORY_SEPARATOR.'Resources'.\DIRECTORY_SEPARATOR.'public';
```
lsmith77: why the backslashes here?

Tests/AbstractTest.php
```diff
@@ -50,8 +50,8 @@
 
     protected function setUp()
     {
-        $this->fixturesPath = realpath(__DIR__.DIRECTORY_SEPARATOR.'Fixtures');
-        $this->temporaryPath = sys_get_temp_dir().DIRECTORY_SEPARATOR.'liip_imagine_test';
+        $this->fixturesPath = realpath(__DIR__.\DIRECTORY_SEPARATOR.'Fixtures');
```
lsmith77: why the backslashes here?

Config/FilterFactoryCollection.php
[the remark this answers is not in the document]
lsmith77: I guess within Symfony the pattern used most often is an `add()` method to build the collection instead of a constructor argument. obviously this has more performance overhead however. I think this is an interesting approach but since variadic in user land is still a new concept there is less experience with it.

but I wonder if it wouldn't be better to use a factory method instead of the constructor. ie. something like: 

```
class FilterFactoryCollection
{
    static public function createFromList(FilterFactoryInterface ...$filterFactories)
    {
        $collection = new self;
        $collection->addList(FilterFactoryInterface ...$filterFactories);
    }
}
```

this keeps the options open for if we ever do need to add anything to the constructor. while keeping the convenience and more or less the overhead reduction benefits.

lsmith77:
maybe we should have a 2.1 branch for this? or should we actually start using master again for future minor/major new versions?

Config/FilterFactoryCollection.php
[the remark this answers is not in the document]
maximgubar: this object has only one purpose, to store the FilterFactoryInterfaces (which are clearly its dependencies and should be specified), I dont see any valid case which would lead to altering constructor. Also, I am not sure is it a good practice to do indirect object injections (via setters).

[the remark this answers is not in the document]
lsmith77: ok lets keep it as proposed. like I said, a good chance of this feeling "weird" is simply the lack of experience with variadic in user land .. and honestly this is not the most "risky" place .. if we do in the end realize we need support additional constructor parameters .. we can then still use a static factory method for that case.

maximgubar:
@lsmith77 we started to use `master` branch again, the base branch of this PR was changed.

PR #1103, opened 2018-05-30
[FlysytemResolve] Fix wrongly resolved path with empty prefix

joker806:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | none
| License | MIT
| Doc PR | none

When `$cachePrefix` is empty then `FlysystemResolver` puts double-slash into resolved path, because it always puts slash between `$this->cachePrefix` and `$filter` in `getFileUrl`. This PR fixes the issue by including the slash in `$this->cachePrefix` but only when it's needed (prefix is not empty).


dbu:
fixed in #1195

Issue #686, opened 2015-12-28
Roadmap 2.0


makasim:
* [ ] Decouple cache from filter and filter config.
* [ ] Decouple loader from filter and filter config.
* [ ] Create a facade service. move logic from filterAction to it.

mappedinn:
What about an implementation of google cloud storage resolver like aws s3? Is it planed?

makasim:
Not it is not. Feel free to contribute it .

mappedinn:
perfect! :)

Rattler3:
@makasim Do you have any plans to remove the /resolve/ from the filter path?

If the resolve was removed in filterAction you could return the binary rather than redirecting to the image path.

makasim:
@ralf57 it was done on purpose both resolve prefix and redirect from filterAction. I am not going to change it

Rattler3:
@makasim thanks for the speedy reply.

Only reason I was asking is when caching html response the path with "resolve" is cached, casing the request to always hit the controller until cache clears. This only happens on image generation, if image exists will cache true image path.

makasim:
I dont see a big problem here, the filterAction returns 301 redirects and browsers are smart enough to remember the url and do not request the action again.

Also You can force cache creation once image is uploaded, or from the cli.

Koc:
@makasim is it possible change behavior to get rid of calling `isStored` https://github.com/liip/LiipImagineBundle/blob/master/Imagine/Cache/CacheManager.php#L141 when generation url? Older versions of the bundle does not check file existance and it was created on demand.

Also there are a lot of issues where people want skip redirects parts. This problems have similar root.

makasim:
What is the problem with `isStored`

Koc:
file_exists call which can be omited like in older versions https://github.com/liip/LiipImagineBundle/blob/0.x/Imagine/Cache/CacheManager.php#L123 .

makasim:
Why do you need it?

Koc:
Cache stored on remote server mounted by sshfs. Yes, it is possible use cached resolver, but simpler sollution do not check existence each the time. Currently we are rewrited resolver and it returns always true for isStored and we are hawing same urls for resolve and target files

makasim:
sshfs is kind of special case I think it is good to solve it by overwriting the default resolver. The other way to go is to add a cache wrapper over the web one.

mitrae:
[...]

Issue #584, opened 2015-03-30
The "format" parameter, caching and file extensions


roderik:
For the KunstmaanBundlesCMS we lean heavily on the LiipImagineBundle for everything image related. But we are currently locked on v0.20.2 because at some point the "format" parameter disappeared. I noticed it's back now in the latest 1.2.x releases, including awesome features we want like postprocessing and interlacing. [Our WIP pull request to update with our configs etc can be found here](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/229)

I've tried upgrading and have gotten it up and running, but i'm running into a small issue. I contacted @lsmith77 on twitter but this deserves a longer format.

**Scenario**:

A user uploads an image file in the admin interface of the CMS. This image could be anything ranging from an optimised jpg to a huge png, bmp, we have even seen 1200dpi tiffs...

In the frontend we pull everything through imagine filters to optimise and resize them. We have three default filters to get people started, an "optimjpg" filter that converts to a jpg file, an "optimpng" that does the same to png, and one "optim" that keeps the format.

Let's say the user uploads a png file and we run it though the "optimjpg" filter. The image get's converted, optimised and cached, but while the content is a jpg, the URL in the cache still has a .png. While most browsers handle this gracefully, it's not like it should be. 

**Question**:

In v0.20.0 you could add the "formats" parameter and get the behaviour where the uploaded png was cached as jpg.

Upgrading is a must for us, but this is blocking it for now. What's the correct way to get this fixed? Do we need to configure something different, or do we need a custom cache implementation, or... ?

makasim:
The format conversion on the fly looks strange to me. The origin path is used as an identity and an extension part too. You want to modify it but it changes the origin image id. And we have to find out how to map cached image to its origin. Things can become too  complicated. Changing formats to guess origin image is a dangerous way. 

Also what about images stored in database or on s3. The origin image may not have an extension at all!.

The most simplest solution is to convert  the origin image on upload before you actually stored it to an internal format, let say jpeg, and then you can apply filters\post processing to that image.

roderik:
Converting on upload is not an option, since we use an "image library" concept one image can be used in lots of ways, each requiring the best format for their use. e.g. use the high res PNG version in a PDF generator, but a low res jpg on a page. 

Since the cache resolver redirects to the cached image that is served through the webserver (apache/nginx), and the webserver uses the extension to set the content type. Unless every file is always served via a controller, there is no way to override this.

Other option i see are:

[...]

PR #1122, opened 2018-08-29
Allow PDF to be find

Seb33300:
This make possible to work with PDF files when using `imagick` driver.

### DIFF
--- a/Imagine/Data/DataManager.php
+++ b/Imagine/Data/DataManager.php
@@ -149,7 +149,7 @@ class DataManager
         }
 
         if (0 !== mb_strpos($binary->getMimeType(), 'image/') && $binary->getMimeType() !== 'application/pdf') {
-            throw new \LogicException(sprintf('The mime type of image %s must be image/xxx got %s.', $path, $binary->getMimeType()));
+            throw new \LogicException(sprintf('The mime type of image %s must be image/xxx or application/pdf got %s.', $path, $binary->getMimeType()));
         }
 
         return $binary;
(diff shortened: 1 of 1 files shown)


Imagine/Data/DataManager.php
```diff
@@ -137,7 +137,7 @@ public function find($filter, $path)
             throw new \LogicException(sprintf('The mime type of image %s was not guessed.', $path));
         }
 
-        if (0 !== mb_strpos($binary->getMimeType(), 'image/')) {
+        if (0 !== mb_strpos($binary->getMimeType(), 'image/') && $binary->getMimeType() !== 'application/pdf') {
             throw new \LogicException(sprintf('The mime type of image %s must be image/xxx got %s.', $path, $binary->getMimeType()));
```
phorzycki: The error message should be modified to reflect your change:

```
The mime type of image %s must be image/xxx or application/pdf, got %s.
```

Seb33300: Fixed

maximgubar:
Hi @Seb33300 ,
Thank you for the merge request,
could you please merge latest master in your branch in order to pass the CI ?

fyrye:
If I am not mistaken, this will result in a file with the `image/png` mime-type  and `.pdf` extension. Until #686 is merged to resolve extension changes #584

dbu:
can you please add a functional test in https://github.com/liip/LiipImagineBundle/blob/master/Tests/Functional/Controller/ImagineControllerTest.php so that we can be sure the response mime type corresponds to the output?

Seb33300:
I rebased my pull request.
If someone has time to work on the functional test please do it.
I am not sure when I will be able to do it.

Issue #1143, opened 2018-12-08
Cache resolution fails if file name contains a plus (+)


sujayjaju:
### Preconditions

1. LiipImagineBundle 2.1
2. PHP 7.2

### Steps to reproduce

1. Try resolving a file named "test+file.jpg". Eg:
`{{ "uploads/test+file.jpg" | imagine_filter('thumb') }}`

### Expected result

* Resolved image

### Actual result

* Error" `Source image for path "uploads/test file.jpg" could not be found`

sujayjaju:
Same issue if there is a hash (#) in the file name

robfrawley:
A hash has a special context in a browser; same with a plus. You would need to encode the URL path for it to work properly.

sujayjaju:
@robfrawley 
Yes, but the problem is that the imagine_filter double encodes the file name, if it is encoded name is passed to the filter. The solution that worked was to extract just the file name after filter and replaced it with the encoded version. That seems like quite a convoluted solution. Should the filter handle those cases?

torinaki:
@sujayjaju I have the same issue. The problem is that `\Liip\ImagineBundle\Imagine\Cache\Resolver\ResolverInterface::resolve` method implementations (`WebPathResolver` and probably others) do not expect that file system path [might contain special characters](https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits). In case of URL, most of these characters should be encoded.

Probably the proper fix will be to convert filesystem path to URL path in following way:
```php
implode('/', array_map('rawurlencode', explode('/', $path)))
```

andyexeter:
I can confirm @torinaki's solution fixes this issue. Care to submit a PR for this?

torinaki:
@anderageru, @sujayjaju PR is created

torinaki:
Fixed in #1162

Seb33300:
I am still encountering issues with images containing a `+`

In filter action, there is no need to `urldecode` the filename because `+` are decoded as spaces and file is not found.

dbalabka:
@Seb33300 could you please describe your case in detail? It would be great to see the URL and expected file path examples.

PR #1162, opened 2019-03-17
Fix special characters encoding in URL path

torinaki:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1143
| License | MIT
| Doc PR | -


torinaki:
@dbu I had the same question and found the following line in `ImageController`
https://github.com/liip/LiipImagineBundle/blob/4c321e023e06e5541d0312eeea5fbc33e758fea8/Controller/ImagineController.php#L72
So the answer is yes. It is enough to use `urlencode` because `\` isn't encoded.

I can create another method just for consistency and use it in `ImageController`

dbu:
that would be great, then its more explicit that these operations are related.

dbu:
and can you please rebase on master? we fixed the build there...

torinaki:
@dbu I have made changes. Please review the latest changes in PR

PR #1149, opened 2018-12-17
Upgrade Enqueue to 0.9

makasim:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | MIT
| Doc PR | updated in this PR

Upgrade Enqueue integration.

PR #1169, opened 2019-04-02
 Upgrade Enqueue to 0.9

gromez:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | MIT
| Doc PR | -

References to PR https://github.com/liip/LiipImagineBundle/pull/1149
Merged last master

### DIFF
--- a/Binary/Loader/StreamLoader.php
+++ b/Binary/Loader/StreamLoader.php
@@ -39,7 +39,7 @@ class StreamLoader implements LoaderInterface
     {
         $this->wrapperPrefix = $wrapperPrefix;
 
-        if ($context && !is_resource($context)) {
+        if ($context && !\is_resource($context)) {
             throw new \InvalidArgumentException('The given context is no valid resource.');
         }
 
@@ -60,7 +60,7 @@ class StreamLoader implements LoaderInterface
          * The error suppression is solely to determine whether the file exists.
          * file_exists() is not used as not all wrappers support stat() to actually check for existing resources.
          */
-        if (($this->context && !$resource = @fopen($name, 'r', null, $this->context)) || !$resource = @fopen($name, 'r')) {
+        if (($this->context && !$resource = @fopen($name, 'rb', null, $this->context)) || !$resource = @fopen($name, 'rb')) {
             throw new NotLoadableException(sprintf('Source image %s not found.', $name));
         }
 
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "aws/aws-sdk-php": "^2.4",
         "doctrine/cache": "^1.1",
         "doctrine/orm": "^2.3",
-        "enqueue/enqueue-bundle": "^0.7|^0.8",
+        "enqueue/enqueue-bundle": "^0.9",
         "friendsofphp/php-cs-fixer": "^2.10",
         "league/flysystem": "^1.0",
         "psr/log": "^1.0",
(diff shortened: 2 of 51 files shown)


Binary/Loader/StreamLoader.php
```diff
@@ -60,7 +60,7 @@ public function find($path)
          * The error suppression is solely to determine whether the file exists.
          * file_exists() is not used as not all wrappers support stat() to actually check for existing resources.
          */
-        if (($this->context && !$resource = @fopen($name, 'r', null, $this->context)) || !$resource = @fopen($name, 'r')) {
+        if (($this->context && !$resource = @fopen($name, 'rb', null, $this->context)) || !$resource = @fopen($name, 'rb')) {
```
dbu: aparently `b`  is active by default and thats why the cs-fixer does not want to have it? https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4011

composer.json
```diff
@@ -35,7 +35,7 @@
         "aws/aws-sdk-php": "^2.4",
         "doctrine/cache": "^1.1",
         "doctrine/orm": "^2.3",
-        "enqueue/enqueue-bundle": "^0.7|^0.8",
+        "enqueue/enqueue-bundle": "^0.9",
```
dbu: we should also adjust the version in the `suggest` section

florianajir:
Any news ??

dbu:
@florianajir would you know about the question i asked, how compatible the enqueue 0.x versions are?

Binary/Loader/StreamLoader.php
florianajir: https://www.php.net/manual/en/function.fopen.php

> For portability, it is strongly recommended that you always use the 'b' flag when opening files with fopen().

florianajir:
@dbu according to https://packagist.org/packages/enqueue/enqueue-bundle the `enqueue/enqueue-bundle` v0.9 requires the same dependencies version `php: ^7.1.3` (patch version requirement upper is not a problem IMO) & `symfony/framework-bundle: ^3.4|^4` as LiipImagineBundle and I don't see any upgrade notice or BC between enqueue-bundle 0.7 & 0.9 so it should not be a problem. 
That being said, considering the optional aspect of this dependency, it should be only suggested and not required in dev part, but I think that this work can be done when this dependency will bump to the next major version. btw it's just my opinion.

Binary/Loader/StreamLoader.php
dbu: hm. i guess then we should go with the php doc and have the "b". it should not hurt, even if the php doc might be outdated and all current php versions now have "b" on. that means we have to disable that check in the .php_cs.dist file

dbu:
thanks for checking up on this. if there is no clear changelog, lets just bump the version and go on...

> That being said, considering the optional aspect of this dependency, it should be only suggested and not required in dev part, but I think that this work can be done when this dependency will bump to the next major version. btw it's just my opinion.

dev dependency is fine. it gets installed when testing this bundle, but anything requiring the bundle will not see those. what would make sense would be to split out the enqueue support into a separate bundle that can be attached to this bundle, instead of mixing things up. but that would indeed be a BC break.

@gromez do you have time to rebase this and solve the conflicts? otherwise @florianajir do you want to take this branch and add the things to wrap it up?

Issue #1166, opened 2019-03-20
Deprecation with Twig 2.7


enumag:
The spaceless tag used [here](https://github.com/liip/LiipImagineBundle/blob/master/Resources/views/Form/form_div_layout.html.twig#L2) is deprecated in Twig 2.7 and should be replaced by `{% filter spaceless %}`. This would however break compatibility with older Twig versions. Not sure how to fix it correctly.

mikemix:
Don't use spaceless at all.

Instead of:

```
{% block liip_imagine_image_widget %}
    {% spaceless %}
        {% if image_path %}
            <div>
                {% if link_url %}
                    <a href="{{ link_filter ? link_url|imagine_filter(link_filter): link_url }}" {% for attrname, attrvalue in link_attr %}{{ attrname }}="{{ attrvalue }}" {% endfor %}>
                {% endif %}

                <img src="{{ image_path|imagine_filter(image_filter) }}" {% for attrname, attrvalue in image_attr %}{{ attrname }}="{{ attrvalue }}" {% endfor %} />

                {% if link_url %}
                    </a>
                {% endif %}
            </div>
        {% endif %}

        {{ block('form_widget_simple') }}
    {% endspaceless %}
{% endblock %}
```

use 

```
{% block liip_imagine_image_widget %}
        {%- if image_path -%}
            <div>
                {%- if link_url -%}
                    <a href="{{ link_filter ? link_url|imagine_filter(link_filter): link_url }}" {% for attrname, attrvalue in link_attr %}{{ attrname }}="{{ attrvalue }}" {% endfor %}>
                {%- endif -%}

                <img src="{{ image_path|imagine_filter(image_filter) }}" {% for attrname, attrvalue in image_attr %}{{ attrname }}="{{ attrvalue }}" {% endfor %} />

                {%- if link_url -%}
                    </a>
                {%- endif -%}
            </div>
        {%- endif -%}

        {{ block('form_widget_simple') -}}
{% endblock %}
```

enumag:
@mikemix looks like you made the PR against the wrong repository.

mikemix:
Lol you're right!

dbu:
fixed in #1176

PR #1173, opened 2019-04-19
Prevent Twig 2.7 spaceless deprecation notice

mikemix:
The spaceless tag has been deprecated in Twig 2.7 and is going to be removed in 3.0

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1166
| License | MIT
| Doc PR | 

To keep backward compatibility we have to stop using `spaceless` at all.

### DIFF
--- a/Resources/views/Form/form_div_layout.html.twig
+++ b/Resources/views/Form/form_div_layout.html.twig
@@ -1,19 +1,17 @@
 {% block liip_imagine_image_widget %}
-    {% spaceless %}
-        {% if image_path %}
-            <div>
-                {% if link_url %}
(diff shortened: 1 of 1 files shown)


mikemix:
Anyone?

Seb33300:
I think this tag was deprecated in favor of the `[apply](https://twig.symfony.com/doc/2.x/tags/apply.html)` tag.
New syntax is `{% apply spaceless %}` instead of `{% spaceless %}`.

This looks overkill to add whitespace trimming everywhere.

But `apply` is only available starting Twig 2.9 so...

stephanvierkant:
I don't think upgrading minimum Twig version isn't a problem, since current supported versions of Symfony (3.4 and 4.2) already require `^2.9|^1.4`. I created a new PR: #1175

PR #1175, opened 2019-05-09
Replaced spaceless by apply filter

stephanvierkant:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes/no
| Fixed tickets | #1166
| License | MIT
| Doc PR | 

I replaced the `spaceless` tag by the `apply` filter.

This means raising the required Twig version, but that's something `symfony/symfony` already did for all supported versions (3.4 and 4.2).

### DIFF
--- a/Resources/views/Form/form_div_layout.html.twig
+++ b/Resources/views/Form/form_div_layout.html.twig
@@ -1,5 +1,5 @@
 {% block liip_imagine_image_widget %}
-    {% spaceless %}
+    {% apply spaceless %}
         {% if image_path %}
             <div>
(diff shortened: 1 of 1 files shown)


franmomu:
I guess the `composer.json` should be updated as well with the `twig/twig` version and maybe move the dependency to the `require` section.

dbu:
according to the ci, older twig versions work fine, the lowest version build works too. i merged #1176 with the same change.

PR #1199, opened 2019-08-03
Upgrade 2.0 from 1.0

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1181, https://github.com/liip/LiipImagineBundle/pull/1011
| License | MIT
| Doc PR | 

I tried to merge 1.0 in 2.0 as in https://github.com/liip/LiipImagineBundle/pull/1011 to bring back some lost commits

PR #1201, opened 2019-08-06
Merge 2.0 branch into master

franmomu:
| Q | A
| --- | ---
| Branch? | master
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

This is like https://github.com/liip/LiipImagineBundle/pull/1199, merging 2.0 branch into master so lost commits are recovered in `master` as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ matrix:
       env: SYMFONY_VERSION=4.0.*
     - php: 7.2
       env:
-        - DEPENDENCIES="symfony/phpunit-bridge:^4"
+        - DEPENDENCIES="symfony/phpunit-bridge:^4.2"
         - COMPOSER_UPDATE_FLAGS="--no-dev"
     - php: 7.2
       env:
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -106,7 +106,14 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->fixXmlConfig('filter_set', 'filter_sets')
             ->children()
-                ->scalarNode('driver')->defaultValue('gd')->end()
+                ->scalarNode('driver')->defaultValue('gd')
+                    ->validate()
+                        ->ifTrue(function ($v) {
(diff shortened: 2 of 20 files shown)


dbu:
i am confused by the branches in this repository. master currently only has the alias that 1.0-dev is version 1, but has no alias for what the master branch should see. if master is 2.x, we should not have a 2.0 branch. if master is 1.x we should not have had the 1.0 branch and should rather merge 1.0 to master than merge 2.0 to master.

note that there is no release of version 2 yet, so master can't mean 3.x. @lsmith77 or @michellesanver can one of you help us out here and clarify this?

lsmith77:
@maximgubar might remember the details .. but afaik 2.0 was used to prepare the 2.0 release .. after the 2.0 release, master became the source for 2.0 releases .. but I am not 100% sure

dbu:
oh, i overlooked the 2.0.0 and 2.1.0 releases because older tags use the `v` prefix. does 2.0 to master have other differences than what you added from 1.0 @franmomu ? if master has no BC breaks versus 2, i suggest we merge this and then remove the 2.0 branch to avoid confusion.

and imho there should be a branch alias for `dev-master: 2.x` if master represents version 2.

franmomu:
oops I thought 2.0 branch was for 2.x releases. Apart from the changes from 1.0, here are included a couple of commits (https://github.com/liip/LiipImagineBundle/commit/644059056f65be78997f6105288db35793fe23fe and https://github.com/liip/LiipImagineBundle/commit/647d51e0952a342a202deaa076b8f17ece44402e) that are in 2.0 and not in master

dbu:
looking at https://github.com/liip/LiipImagineBundle/compare/2.0...master there seem to be cs cleanups and one big new feature, the configuration model classes. i don't think they are a BC break, so imho we can merge this and then remove the 2.0 branch for clarity. wdyt @michellesanver ?

michellesanver:
Sounds like a perfect plan! :)

PR #1202, opened 2019-08-08
Improve travis-ci matrix

fbourigault:
- Drop support for Symfony 4.0 and 4.1 wich are no longer supported.
- Remove php-cs-fixer travis job as the project use styleci.
- Add PHP 7.4 in the matrix as allowed failure.
- Test against each supported Symfony versions.
- Test against upcoming Symfony 4.4 with allowed failure.
- Run code coverage with highest supported PHP version (7.3).
- Install `imagick` extension to avoid skipped tests.
- Fix risky `FilterSetTest::testSetFiltersWithValidFilterSuccess` test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,8 @@ before_install:
   - if [[ "$COVERALLS_ENABLED" == "true" ]]; then
       travis_retry composer require --dev satooshi/php-coveralls:^2.0 --no-update $COMPOSER_FLAGS;
     fi
+  - pecl channel-update pecl.php.net
+  - yes | pecl install imagick
 
 install:
   - travis_retry composer update --prefer-dist --no-interaction --no-suggest --no-progress --ansi $COMPOSER_FLAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,104 +1,91 @@
 language: php
-
 sudo: false
-
 cache:
-
   directories:
     - $HOME/.composer/cache
+    - $HOME/symfony-bridge/.phpunit
 
 env:
-
   global:
     - SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
     - PHPUNIT_FLAGS="-v"
     - PHPUNIT_ENABLED="true"
     - SYMFONY_PHPUNIT_VERSION="6.5"
-    - PHPCSFIXER_ENABLED="false"
     - COVERALLS_ENABLED="false"
+    - STABILITY=stable
 
 matrix:
-
   fast_finish: true
-
   include:
+    - php: '7.1'
+    - php: '7.2'
 
-    - php: 7.1
-    - php: 7.1
+    # Enable code coverage with the latest supported PHP version
+    - php: '7.3'
       env:
-        - PHPUNIT_ENABLED="false"
-        - PHPCSFIXER_ENABLED="true"
-    - php: 7.1
+        - COVERALLS_ENABLED="true"
+        - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
+
+    # Minimum supported dependencies with the latest and oldest supported PHP versions
+    - php: '7.1'
       env:
-        - SYMFONY_VERSION=4.0.*
         - COMPOSER_FLAGS="--prefer-lowest"
-    - php: 7.2
-    - php: 7.2
-      env: SYMFONY_VERSION=3.4.*
-    - php: 7.2
-      env: SYMFONY_VERSION=4.0.*
-    - php: 7.2
+    - php: '7.3'
       env:
-        - DEPENDENCIES="symfony/phpunit-bridge:^4.2"
-        - COMPOSER_UPDATE_FLAGS="--no-dev"
-    - php: 7.2
+        - COMPOSER_FLAGS="--prefer-lowest"
+
+    # Test each supported Symfony version with lowest supported PHP version
+    - php: '7.1'
       env:
-        - COVERALLS_ENABLED="true"
-        - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
-    - php: 7.2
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,87 +15,56 @@ env:
...
-    fi
   - if [[ "$STABILITY" != "" ]]; then
       travis_retry composer config minimum-stability $STABILITY;
     fi
   - if [[ "$COVERALLS_ENABLED" != "true" ]]; then
       phpenv config-rm xdebug.ini || true;
     fi
   - if [[ "$COVERALLS_ENABLED" == "true" ]]; then
-      travis_retry composer require --dev satooshi/php-coveralls:^2.0@dev --no-update $COMPOSER_FLAGS;
+      travis_retry composer require --dev satooshi/php-coveralls --no-update $COMPOSER_FLAGS;
```
dbu: what if version 3 of php-coveralls introduces a new set of rules? this would randomly start to fail once they release, no?

.travis.yml
```diff
@@ -18,87 +15,56 @@ env:
...
         - PHPCSFIXER_ENABLED="true"
-    - php: 7.1
-      env:
-        - SYMFONY_VERSION=4.0.*
-        - COMPOSER_FLAGS="--prefer-lowest"
-    - php: 7.2
-    - php: 7.2
-      env: SYMFONY_VERSION=3.4.*
-    - php: 7.2
-      env: SYMFONY_VERSION=4.0.*
```
dbu: what is the reasoning for not testing each allowed symfony minor version? that can sometimes find issues with symfony version differences.

fbourigault: Doesn’t the Symfony BC promise enough? If not what would be the rules? I often see hard to maintain Travis matrices when testing against each Symfony version.

What about testing each supported and maintained Symfony version with the lowest (or highest) supported php version?

.travis.yml
fbourigault: I will just remove the `@dev` then.

.travis.yml
fbourigault: Also, what about dropping Sf 4.0, 4.1 and 4.2 as those are no longer maintained? 4.2 could be kept as it’s still maintained for security.

This would help to keep the matrix clean.

fbourigault:
What about dropping the php-cs-fixer stuff from Travis as the project use styleci?

.travis.yml
dbu: i have seen things go wrong in unexpected ways before, would prefer to test each minor version we claim to support. +1 on dropping 4.0 and 4.1 support. 4.2 sounds like we should keep it while it still gets security support.

dbu:
> What about dropping the php-cs-fixer stuff from Travis as the project use styleci?
yeah one of these sounds like its enough.

fbourigault:
I'm trying to test each Symfony version using https://twitter.com/Yatsenco/status/1160159201643175936

fbourigault:
It looks like the travis-ci build are no longer triggering. Does anyone know what is happening?

dbu:
the hooks are still active, maybe it was a temporary hickup. can you rebase on master and re-push? maybe travis runs again now... there are conflicts with master anyways, so rebase is useful.

fbourigault:
Not better :/ I looked at Github and Travis statuses and both are green.

fbourigault:
It's working now. Had invalid yaml in `.travis.yml` :/

In the `composer.json`, we have:
```
    "minimum-stability": "dev",
    "prefer-stable": true,
```

In the `.travis.yml`, we have:
```
  - if [[ "$STABILITY" != "" ]]; then
      travis_retry composer config minimum-stability $STABILITY;
    fi
```

IMHO, we should drop one. I would drop the `composer.json` part as this would prevent to release this bundle with unstable dependencies.

What do you think?

dbu:
agreed. we could then have one single build that runs with stability dev, to detect if new and upcoming versions would have a break. there is a couple of symfony regressions that we prevented that way, and edge cases where we had relied on odd behaviour that got fixed, so that we could adjust in time...

.travis.yml
```diff
@@ -1,77 +1,71 @@
...
+    - php: 7.2
+    - php: 7.3
+      env:
+        - COVERALLS_ENABLED="true"
+        - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
 
+    # Minimum supported dependencies with the latest and oldest supported PHP versions
     - php: 7.1
+      env:
+        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
```
dbu: do we need --prefer-stable? i think not, it seems to be in composer.json (plus as you say, we should remove the minimum-stability setting in composer.json and then the flag becomes entirely obsolete)

dbu: you'll need to change composer.json too

.travis.yml
```diff
@@ -1,104 +1,93 @@
...
         - COMPOSER_FLAGS="--ignore-platform-reqs"
-
   allow_failures:
-
-    - env:
-        - COVERALLS_ENABLED="true"
-        - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
     - env:
-        - SYMFONY_VERSION=dev-master
         - STABILITY=dev
```
dbu: according to the [doc](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail) *all* conditions must match, there is no partial matching. which seems stupid, it would be elegant if this would work as you expect it to.

fbourigault:
This is now ready. Should I add to the changelog that PHP 7.3 is now officially supported and Symfony 4.0, 4.1 support is now dropped?

fbourigault:
I removed testing against Symfony 5.0 as Symfony flex is not yet able to force installation of this Symfony version (https://github.com/symfony/flex/issues/553).

fbourigault:
This PR is now ready!

PR #1209, opened 2019-08-22
Fix deprecated use of SYMFONY_DEPRECATIONS_HELPER

fbourigault:
This remove deprecated use of `weak_vendors` as `SYMFONY_DEPRECATIONS_HELPER` value. We are now using `max[self]=0` and using `symfony/phpunit-bridge` `^4.3`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
 
 env:
   global:
-    - SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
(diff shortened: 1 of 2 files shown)


dbu:
lowest version builds don't know that option yet. can you add the old name to the 2 matrix entries about lowest version?

fbourigault:
What about bumping `symfony/phpunit-bridge` to `^4.3`? This package is expected to be used in it's latest version whatever which version of PHP/Symfony we are testing.

PR #1185, opened 2019-05-31
Fixes/Refactor for Symfony 4.3

hjanuschka:
| Q | A
| --- | ---
| Branch? | master
| Bug fix? | no
| New feature? | no
| BC breaks? |  🤔
| Deprecations? | no
| Tests pass? | yes
| License | MIT

hey there 🤚,
we really love this project, and highly depend on it, we are in the process of upgrading all our symfony apps to `4.3` and stumbled upon some issues.

this PR - refactors the bundle and solves following things:

  - remove all deprecation notes
    - Event from Contract
    - EventDispatcher signature
    - MimeType Guesser
    - rootBuilder (i did the initial PR, but somehow missed the tests, https://github.com/liip/LiipImagineBundle/pull/1139)
    - remove `symfony/template` dependency (reference: https://symfony.com/blog/new-in-symfony-4-3-deprecated-the-templating-component-integration)

let me know what you think, and if you'd need me to change anything.

PR #1212, opened 2019-08-22
Fix event dispatcher deprecation sf 4.3

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

An attempt to get rid of event dispatcher deprecations using symfony 4.3. If someone can think of other way... I'm happy to change it :)

### DIFF
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -20,6 +20,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as ContractsEventDispatcherInterface;
 
 class CacheManager
 {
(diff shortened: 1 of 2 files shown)


dbu:
thanks for this one!

in #1185, someone mentioned a legacy event dispatch proxy. maybe you could use that?

franmomu:
yeah it was me 😆, after taking a look at that class I find it a little bit tricky, because, if we do this in the constructor (we can only use `LegacyEventDispatcherProxy` in sf >= 4.3):

```php
if (class_exists(LegacyEventDispatcherProxy::class)) {
    $this->dispatcher = LegacyEventDispatcherProxy::decorate($dispatcher);
} else {
    $this->dispatcher = $dispatcher;
}
```
we have to keep dispatching events like this (because with sf <4.3 it won't work):
```php
$this->dispatcher->dispatch(ImagineEvents::PRE_RESOLVE, $preEvent);
```
So the `LegacyEventDispacherProxy` executes this:
```php
@trigger_error(sprintf('The signature of the "%s::dispatch()" method should be updated to "dispatch($event, string $eventName = null)", not doing so is deprecated since Symfony 4.3.', $r->class), E_USER_DEPRECATED);
```
That's why I decided not to use it.

fbourigault:
I'm not super fan of using the Kernel version to detect with signature to use.

Someone may be stuck with http-kernel 4.2 and using event-dispatcher 4.3 and this would fail.

This is an edge case, but not impossible.

As since 4.3 the event-dispatcher implements `Symfony\Contracts\EventDispatcher\EventDispatcherInterface`, what about checking if the dispatcher implements this interface or not to choose the right signature?

dbu:
thank you!

franmomu:
you're very welcome!

PR #1215, opened 2019-08-23
Remove minimum-stability and prefer-stable configuration

fbourigault:
#1202 expected to remove `minimum-stability` and `prefer-stable` from composer.json.

Those where initially removed but came back while handling a rebase conflict.


dbu:
ups, thanks!

Issue #1132, opened 2018-10-12
Reset settings when the config is divided into parts


peter-gribanov:
I want to put the filter configurations into modules, but in this case the global settings `resolvers` and `loaders` will no longer apply to all filters.

### Preconditions

```
# php -v
PHP 7.2.8 (cli) (built: Jul 21 2018 07:56:11) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.8, Copyright (c) 1999-2018, by Zend Technologies
```

```
# composer show --latest 'liip/*'
liip/imagine-bundle 1.9.1 2.1.0 This bundle provides an image manipulation abstraction toolkit for Symfony-based projects.
```

### Steps to reproduce

1. Create global config `/config/packages/liip_imagine.yml` with:
```yml
liip_imagine:
    resolvers:
        default:
            web_path:
                cache_prefix: 'media/suit'
    filter_sets:
        autocomplete:
            filters:
                fixed: { width: 60, height: 60 }
```

2. Create config in module with:

```yml
liip_imagine:
    filter_sets:
        news_cover_in_list:
            filters:
                fixed: { width: 800, height: 600 }
```

3. Then use this filters in template:

```html
<img src="{{ 'media/test.jpg' | imagine_filter('autocomplete') }}" alt="" />
<img src="{{ 'media/test.jpg' | imagine_filter('news_cover_in_list') }}" alt="" />
```

### Expected result

```html
<img src="/media/suit/autocomplete/media/test.jpg" alt="" />
<img src="/media/suit/news_cover_in_list/media/test.jpg" alt="" />
```

### Actual result

```html
<img src="/media/cache/autocomplete/media/test.jpg" alt="" />
<img src="/media/cache/news_cover_in_list/media/test.jpg" alt="" />
```

c33s:
i also had this problem (and also 2 friends of mine @vworld, @MaxValEh)

we split our config in `edition.yaml` and `app.yaml` where `edition.yaml` are our global defaults and `app.yaml` are the settings for the current project/app.

our `kernel.php` looks a little bit different:
```
    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
    {
        parent::configureContainer($container, $loader);

        $confDir = $this->getProjectDir().'/config';
[...]

PR #1216, opened 2019-08-26
[RFC] removed webpath set to null on unset resolver

c33s:
| Q | A
| --- | ---
| Branch? | 2.0 would maybe also make sense to backport it
| Bug fix? | yes (for me yes)
| New feature? | no
| BC breaks? | yes (kind of, if someone buld upon this fallback)
| Deprecations? | no
| Tests pass? | ?
| Fixed tickets | #1132
| License | MIT
| Doc PR | 

alredy commented in https://github.com/liip/LiipImagineBundle/issues/1132#issuecomment-524867253 but i thought a PR is more visible to get comments.

haven't tried to run or fix tests because first it makes sense to know if you would merge this.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -95,10 +95,6 @@ class Configuration implements ConfigurationInterface
                         throw new \LogicException('Resolvers has to be array');
                     }
 
-                    if (false === \array_key_exists('default', $v['resolvers'])) {
-                        $v['resolvers']['default'] = ['web_path' => null];
-                    }
-
                     return $v;
                 })
             ->end();


michellesanver:
Thanks for this PR! I would definitely merge this bugfix and release in 2.3 as it is indeed a slight BC break. I'd be super happy if you go ahead and fix the related tests.

c33s:
sounds great. will have a look at the tests this week

dbu:
i don't think we can simply remove the default resolver. maybe we could add it in the extension class instead of the configuration class to avoid the problem when merging configurations?

PR #1218, opened 2019-08-28
Fix deprecated usage of Symfony\Component\EventDispatcher\Event class

fbourigault:
This fix usage of `Symfony\Component\EventDispatcher\Event` which was deprecated in Symfony 4.3.

### DIFF
--- a/Events/CacheResolveEvent.php
+++ b/Events/CacheResolveEvent.php
@@ -12,8 +12,22 @@
 namespace Liip\ImagineBundle\Events;
 
 use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\EventDispatcher\Event as ContractsEvent;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as ContractsEventDispatcherInterface;
 
-class CacheResolveEvent extends Event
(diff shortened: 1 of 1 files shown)


dbu:
oh, elegant solution!

this seems to fail with symfony 4.2 for some reason.

Events/CacheResolveEvent.php
```diff
@@ -12,8 +12,22 @@
 namespace Liip\ImagineBundle\Events;
 
 use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\EventDispatcher\Event as ContractsEvent;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as ContractsEventDispatcherInterface;
 
-class CacheResolveEvent extends Event
+if (is_subclass_of(EventDispatcherInterface::class, ContractsEventDispatcherInterface::class)) {
```
fbourigault: I'm using such test because in previous build I got the 4.2 event dispatcher installed alongside contracts 1.1. In such configuration, `Symfony\Component\EventDispatcher\EventDispatcherInterface` does not implements `Symfony\Contracts\EventDispatcher\EventDispatcherInterface` and so does require our event to extend `Symfony\Component\EventDispatcher\Event`.

dbu: yep, that makes sense to me :+1:

PR #1219, opened 2019-08-30
[Post-Processors] Add timeout option, refactor processors, add tests for all, add/remove/deprecate options

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/857 https://github.com/liip/LiipImagineBundle/pull/993
| License | MIT
| Doc PR | 

I cherry-picked the commit from https://github.com/liip/LiipImagineBundle/pull/993 and updated with `master` branch.

These are the changes I made from the original commit:
- `InvalidOptionException`: Removed the `var_export` call because it showed different results for `new stdClass` in `php 7.1` and `php 7.3`
- `AbstractPostProcessor `: When creating the `Process`, I removed `prefix` and `options` that were available in `ProcessBuilder` and not in `Process`.


Imagine/Filter/PostProcessor/AbstractPostProcessor.php
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * @param BinaryInterface $binary
+     * @param array           $options
+     * @param null            $prefix
+     *
+     * @return string
+     */
+    protected function writeTemporaryFile(BinaryInterface $binary, array $options = [], $prefix = null): string
```
dbu: $prefix is a string, right? should have the static type hint.

Imagine/Filter/PostProcessor/AbstractPostProcessor.php
```diff
@@ -0,0 +1,185 @@
...
+    {
+        return in_array($binary->getMimeType(), $types, true);
+    }
+
+    /**
+     * @param BinaryInterface $binary
+     * @param array           $options
+     * @param null            $prefix
+     *
+     * @return string
```
dbu: such docblocks don't add any information. in projects i work on, i don't do them anymore unless i actually add information (e.g. string[] for an array, or explanations about the parameters). when i add them, i only add the parameters / return where i actually have something to add. otherwise its redundancy imho.

i would suggest doing the same with new code being added.

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
```diff
@@ -46,129 +40,135 @@ class JpegOptimPostProcessor implements PostProcessorInterface
...
+    {
+        $arguments = [$this->executablePath];
+
+        if ($options['strip_all'] ?? $this->strip) {
+            $arguments[] = '--strip-all';
         }
 
-        $proc = new Process($processArguments);
-        $proc->run();
+        if (isset($options['max'])) {
```
dbu: could we do this in the dependency injection class already, and maybe also normalize the config, instead of at runtime?

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
```diff
@@ -46,129 +40,135 @@ class JpegOptimPostProcessor implements PostProcessorInterface
...
+                throw new InvalidOptionException('the "max" and "quality" options cannot both be set', $options);
+            }
+
+            $options['quality'] = $options['max'];
         }
 
-        $result = new Binary(file_get_contents($input), $binary->getMimeType(), $binary->getFormat());
+        if ($quality = $options['quality'] ?? $this->quality) {
+            if (!in_array($options['quality'], range(0, 100))) {
+                throw new InvalidOptionException('the "quality" option must be an int between 0 and 100', $options);
```
dbu: would also be nicer if this was discovered in the DI extension instead of at runtime. if this is too generic in the configuration, we could maybe have the processors provide a static validation method that the DI class can call.

Imagine/Filter/PostProcessor/AbstractPostProcessor.php
franmomu: totally agree, I'll take a look at those docblocks

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
franmomu: interesting thought, I'll give it a try

dbu: as this is not new checks but only refactorings, i think it makes sense to do a separate pull request. is this refactoring ready to be merged in your opinion?

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
```diff
@@ -46,129 +40,127 @@ class JpegOptimPostProcessor implements PostProcessorInterface
...
+        $this->strip = $strip;
 
         return $this;
     }
 
-    /**
-     * @param BinaryInterface $binary
-     * @param array           $options
-     *
-     * @throws ProcessFailedException
```
dbu: i would keep the exception annotations. they add information that is not part of the code.

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
```diff
@@ -46,129 +40,127 @@ class JpegOptimPostProcessor implements PostProcessorInterface
...
 
-        $processArguments[] = $input;
-        if ($binary instanceof FileBinaryInterface) {
-            copy($binary->getPath(), $input);
-        } else {
-            file_put_contents($input, $binary->getContent());
+        return $result;
+    }
+
+    private function getProcessArguments(array $options = []): array
```
dbu: do we know that these arrays are array of string? then i would use phpdoc annotations with `@param $options string[]` and `@return string[]` as there are no typed arrays in php.

Imagine/Filter/PostProcessor/MozJpegPostProcessor.php
```diff
@@ -24,84 +23,76 @@
...
         $this->quality = $quality;
 
         return $this;
     }
 
-    /**
-     * @param BinaryInterface $binary
-     * @param array           $options
-     *
-     * @throws ProcessFailedException
```
dbu: keep the exception annotation?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -36,83 +30,108 @@ class OptiPngPostProcessor implements PostProcessorInterface
...
-        $this->stripAll = $stripAll;
-        $this->tempDir = $tempDir ?: sys_get_temp_dir();
+        $this->strip = $strip;
     }
 
-    /**
-     * @param BinaryInterface $binary
-     * @param array           $options
-     *
-     * @throws ProcessFailedException
```
dbu: keep this line?

Imagine/Filter/PostProcessor/OptiPngPostProcessor.php
```diff
@@ -36,83 +30,108 @@ class OptiPngPostProcessor implements PostProcessorInterface
...
+        }
+
+        $result = new Binary(file_get_contents($file), $binary->getMimeType(), $binary->getFormat());
+
+        unlink($file);
+
+        return $result;
+    }
+
+    private function getProcessArguments(array $options = []): array
```
dbu: string[] ?

Imagine/Filter/PostProcessor/PngquantPostProcessor.php
```diff
@@ -25,76 +25,115 @@
...
+        $this->triggerSetterMethodDeprecation(__METHOD__);
         $this->quality = $quality;
 
         return $this;
     }
 
-    /**
-     * @param BinaryInterface $binary
-     *
-     * @throws ProcessFailedException
```
dbu: keep?

Imagine/Filter/PostProcessor/PngquantPostProcessor.php
```diff
@@ -25,76 +25,115 @@
...
+            throw new ProcessFailedException($process);
+        }
 
-        // 98 and 99 are "quality too low" to compress current current image which, while isn't ideal, is not a failure
-        if (!\in_array($proc->getExitCode(), [0, 98, 99], true)) {
-            throw new ProcessFailedException($proc);
+        return new Binary($process->getOutput(), $binary->getMimeType(), $binary->getFormat());
+    }
+
+    private function getProcessArguments(array $options = []): array
```
dbu: string[] ?

Imagine/Filter/PostProcessor/AbstractPostProcessor.php
```diff
@@ -0,0 +1,136 @@
...
+        if ($binary instanceof FileBinaryInterface) {
+            $this->filesystem->copy($binary->getPath(), $temporary, true);
+        } else {
+            $this->filesystem->dumpFile($temporary, $binary->getContent());
+        }
+
+        return $temporary;
+    }
+
+    protected function acquireTemporaryFilePath(array $options, $prefix = null): string
```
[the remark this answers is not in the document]
franmomu: 👍 I didn't know it was introduced in 7.2

Imagine/Filter/PostProcessor/JpegOptimPostProcessor.php
franmomu: oops, I overlooked this

Tests/Imagine/Filter/PostProcessor/MozJpegPostProcessorTest.php
```diff
@@ -11,25 +11,113 @@
...
+        $this->assertContains($expected, $result->getContent());
+        $this->assertContains($content, $result->getContent());
+
+        @unlink($file);
+    }
+
+    /**
+     * @dataProvider provideProcessData
+     *
+     * @expectedException \Symfony\Component\Process\Exception\ProcessFailedException
```
dbu: this is deprecated, the recommended way is to do `$this->expectException(ProcessFailedException::class);` on the line right before the one that should trigger the exception.

Tests/Imagine/Filter/PostProcessor/OptiPngPostProcessorTest.php
```diff
@@ -11,25 +11,164 @@
...
+    public function testInvalidInterlaceOption(): void
+    {
+        $this->getProcessArguments(['interlace_type' => 10]);
+    }
+
+    /**
+     * @group legacy
+     *
+     * @expectedException \Liip\ImagineBundle\Exception\Imagine\Filter\PostProcessor\InvalidOptionException
+     * @expectedExceptionMessage the "strip" and "strip_all" options cannot both be set
```
dbu: there is `$this->expectExceptionMessage()` as well

Tests/Imagine/Filter/PostProcessor/PngquantPostProcessorTest.php
```diff
@@ -11,25 +11,177 @@
...
  * @covers \Liip\ImagineBundle\Imagine\Filter\PostProcessor\PngquantPostProcessor
  */
-class PngquantPostProcessorTest extends AbstractTest
+class PngquantPostProcessorTest extends AbstractPostProcessorTestCase
 {
-    public function testPngquantPostProcessor()
+    /**
+     * @group legacy
+     *
+     * @expectedDeprecation The %s::setQuality() method was deprecated in %s and will be removed in %s. You must setup the class state via its __construct() method. You can still pass filter-specific options to the process() method to overwrite behavior.
```
dbu: i'd assume there is an expectDeprecation method as well on phpunit, but i could be mistaken.

franmomu:
the more thing you notice the better! 😬

franmomu:
> noticed some more things in the tests, sorry.
> 
> i was wondering if we should typehint in the constructor methods (that you only refactor). it might break for someone, but then again when the constructor arguments are not the types they should be, i don't think it would have worked now either.

Exactly, if `$executablePath` is not a `string`, I don't think it would work either. Since I added typehint to the `AbstractPostProcessor` I guess the ones extending this class can be also typehinted.

PR #1101, opened 2018-05-27
[DI] [Controller] Add controller redirect response code configuration option

robfrawley:
| Q | A
| --- | ---
| Branch? | `2.0`
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | #970
| License | MIT
| Doc PR | included in PR

This PR re-adds the `liip_imagine.controller.redirect_response_code` configuration option originally merged into `1.0` (as #970) but never merged into `2.0`.

Introduces a new `Liip\ImagineBundle\Config\Controller\ControllerConfig` service that is passed as the forth argument to `Liip\ImagineBundle\Controller\ImagineController`. Not passing an instance of the new `ControllerConfig` object will now throw a deprecation message and the ability to not do so will be removed in `3.0`.

The `ControllerConfig` object will also reject non-redirect HTTP codes; the configured code must be one of `201`, `301`, `302`, `303`, `307`, or `308` (otherwise an invalid argument exception will be thrown).

---

See https://github.com/liip/LiipImagineBundle/pull/1011#issuecomment-388824553 and its following conversation for more context on the missing PRs from `2.0` that were already merged into `1.x`.

PR #970, opened 2017-08-04
Allow to configure the HTTP response code for redirects

lstrojny:
| Q | A
| --- | ---
| Branch? | 1.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #969
| License | MIT

Makes redirect code configurable.

PR #1220, opened 2019-08-30
[DI] [Controller] Add controller redirect response code configuration option

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | #970 https://github.com/liip/LiipImagineBundle/pull/1101
| License | MIT
| Doc PR | 

Rebased version of https://github.com/liip/LiipImagineBundle/pull/1101

### DIFF
--- a/Config/Controller/ControllerConfig.php
+++ b/Config/Controller/ControllerConfig.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Config\Controller;
+
+use Liip\ImagineBundle\Exception\InvalidArgumentException;
+
+final class ControllerConfig
+{
+    public const REDIRECT_RESPONSE_CODES = [201, 301, 302, 303, 307, 308];
+
+    private $redirectResponseCode;
+
+    public function __construct(int $redirectResponseCode)
+    {
+        if (!in_array($redirectResponseCode, self::REDIRECT_RESPONSE_CODES, true)) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid redirect response code "%s" (must be 201, 301, 302, 303, 307, or 308).', $redirectResponseCode
+            ));
+        }
+
+        $this->redirectResponseCode = $redirectResponseCode;
+    }
+
+    public function getRedirectResponseCode(): int
+    {
+        return $this->redirectResponseCode;
+    }
+}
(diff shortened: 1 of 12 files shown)


Resources/config/imagine.xml
```diff
@@ -139,14 +141,18 @@
...
-            <argument type="service" id="liip_imagine.cache.signer" />
+        <!-- Config -->
+
+        <service id="Liip\ImagineBundle\Config\Controller\ControllerConfig" public="false">
+            <argument><!-- will be injected by FileSystemLoaderFactory --></argument>
         </service>
 
+        <!-- Controller -->
+
+        <service id="Liip\ImagineBundle\Controller\ImagineController" public="true" autowire="true" />
```
dbu: afaik best practice for bundles is to not use autowire as it can be a bit fragile.

Config/Controller/ControllerConfig.php
```diff
@@ -0,0 +1,44 @@
...
+     * @var int
+     */
+    private $redirectResponseCode;
+
+    /**
+     * @param int $redirectResponseCode
+     */
+    public function __construct(int $redirectResponseCode)
+    {
+        if (!in_array($redirectResponseCode, [201, 301, 302, 303, 307, 308], true)) {
```
dbu: i would prefer to put this validation into the configuration definition / DI class.

that way you get an exception right away when building the container, and not only when actually using the controller.

Resources/config/imagine.xml
```diff
@@ -139,14 +141,18 @@
             <argument type="service" id="logger" on-invalid="ignore" />
         </service>
 
-        <!-- Controller -->
+        <service id="Liip\ImagineBundle\Service\FilterService" alias="liip_imagine.service.filter" />
```
dbu: i'd only define such alias for autowiring services that we expect users of the bundle to need in their own services. not sure which of the services are expected to be used.
as said below, bundles should not use autowiring for their own services.

Resources/config/imagine.xml
franmomu: yes, you're wright

Config/Controller/ControllerConfig.php
franmomu: yeah, it makes perfect sense in the configuration definition, but would you leave also this validation here? I think it should also belong here

Resources/config/imagine.xml
franmomu: yes, I'll leave it as it was

Config/Controller/ControllerConfig.php
dbu: yep, it makes sense to validate here as people could be configuring the service themselves. how about defining a public constant with this array in this class and use that constant in the validation of Configuration.php ?

PR #981, opened 2017-08-30
[Console] Lazy commands, separate IO concerns, add forced resolve to command, and prittify and detail output

robfrawley:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

This is the `2.0` version of #967. Included are commands with the following new features:

- Use of commands as a service
- Removal of "ContainerAwareInterface" from commands (instead, the container is used to pass required dependencies)
- Updated styling that includes colored output, as well as the ability to disable colors and print machine-readable output
- Separation of concerns between the commands, the IO/style, etc
- Enable comparable output for the remove cache command

The use of explicit dependencies (via the constructor), as well as the separation of concerns, results in relatively simple commands, despite the complex behavior and output styling introduced. See:

- [ResolveCacheCommand.php](https://github.com/robfrawley/LiipImagineBundle/blob/c136b8fbad60ff31ae323b8f221c1e322f837872/Command/ResolveCacheCommand.php)
- [RemoveCacheCommand.php](https://github.com/robfrawley/LiipImagineBundle/blob/c136b8fbad60ff31ae323b8f221c1e322f837872/Command/RemoveCacheCommand.php)

It is also worth noting that the bulk of the PR size is new and updated unit tests. The new command output for the cache resolve command looks like the following:

imagine-cli-resolve_cached-resolved-failed_normal
imagine-cli-resolve_cached-resolved-failed_no-colors
imagine-cli-resolve_cached-resolved-failed_as-script
imagine-cli-resolve_cached-resolved_normal
imagine-cli-resolve_failed_all

PR #1222, opened 2019-09-09
[Console] Lazy commands, separate IO concerns, add forced resolve to command, and prettify and detail output

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/pull/981
| License | MIT
| Doc PR | 

As some of the previous PRs, this is to pick https://github.com/liip/LiipImagineBundle/pull/981 up, there was a lot of work done and it's a pity to leave it behind.


Command/CacheCommandTrait.php
```diff
@@ -0,0 +1,142 @@
...
+            $filterSetsSize = count($filters);
+            $allActionsSize = ($filterSetsSize * $imagePathsSize) - $this->failures;
+            $allActionsWord = $wordPluralizer($allActionsSize, 'operation');
+
+            switch (substr(__CLASS__, -18, 6)) {
+                case 'Remove':
+                    $allActionsWord = $wordPluralizer($allActionsSize, 'removal');
+                    break;
+
+                case 'esolve':
```
dbu: esolve? should this be Resolve?

Command/RemoveCacheCommand.php
```diff
@@ -20,56 +21,89 @@
...
         $this
-            ->setDescription('Remove cache for given paths and set of filters.')
-            ->addArgument('paths', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, 'Image paths')
-            ->addOption(
-                'filters',
-                'f',
-                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
-                'Filters list'
-            )
+            ->setName('liip:imagine:cache:remove')
```
dbu: why move the name from constructor to here? afaik the symfony 4 way is a static protected property for the name, which can avoid instantiating the command in some situations.

Command/RemoveCacheCommand.php
```diff
@@ -20,56 +21,89 @@
...
+            ->setAliases(['imagine:del'])
```
dbu: this is a new thing, right? not sure if we should start doing this - symfony console already would find this as `l:i:c:rem` unless you have some very similar commands available.

Command/ResolveCacheCommand.php
```diff
@@ -21,68 +22,109 @@
...
-    http://localhost/media/cache/thumb2/path1</info>
-
-If you omit --filters parameter then to resolve given paths will be used all configured and available filters in application:
-<info>php app/console %command.name% path1</info>
-As a result you will get<info>
-    http://localhost/media/cache/thumb1/path1
-    http://localhost/media/cache/thumb2/path1</info>
+            ->setName('liip:imagine:cache:resolve')
+            ->setAliases(['imagine:get'])
+            ->setDescription('Resolve cache entries for the given paths and filters.')
```
dbu: can we be a bit more concrete what this does? print out the existing cached files for given source image and optional path? "resolve" repeats the name and is not obvious to me what it means in this context.

Command/ResolveCacheCommand.php
```diff
@@ -21,68 +22,109 @@
...
+            ->addArgument('path', InputArgument::REQUIRED | InputArgument::IS_ARRAY,
+                'Image file path(s) to run resolution on.')
+            ->addOption('filter', 'f', InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+                'Filter(s) to use for image resolution; if none explicitly passed, use all filters.')
+            ->addOption('force', 'F', InputOption::VALUE_NONE,
+                'Force re-resolution of image, regardless of whether it has been previously cached.')
```
dbu: does this mean the image in the cache will be re-generated? and does that mean this command not only shows what cached images exist, but does a cache warmup?

Command/RemoveCacheCommand.php
```diff
@@ -20,56 +21,89 @@
...
+        }
+
+        $this->io->group($image, $filter, 'blue');
+        $this->io->space();
 
-        if (empty($filters)) {
-            $filters = null;
+        if ($this->cacheManager->isStored($image, $filter)) {
+            $this->cacheManager->remove($image, $filter);
+            $this->io->status('removed', 'green');
```
dbu: not sure about the colors. removed could be red i feel...

Command/ResolveCacheCommand.php
```diff
@@ -21,68 +22,109 @@
...
-                $output->writeln($this->filterService->getUrlOfFilteredImage($path, $filter));
+        $this->io->group($image, $filter, 'blue');
+        $this->io->space();
+
+        try {
+            if ($forced || !$this->cacheManager->isStored($image, $filter)) {
+                $this->cacheManager->store($this->filterManager->applyFilter($this->dataManager->find($filter, $image), $filter), $image, $filter);
+                $this->io->status('resolved', 'green');
+            } else {
+                $this->io->status('cached', 'yellow');
```
dbu: do we have something like "light green"? yellow indicates a warning, but existing cache is not a problem. maybe just white?

Component/Console/Style/ImagineStyleInterface.php
```diff
@@ -0,0 +1,44 @@
...
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Component\Console\Style;
+
+/**
+ * @internal
+ */
+interface ImagineStyleInterface
```
dbu: can we re-validate which of these are actually used (outside of the test)? i have the impression only half of these are actually used. we should not define more than we need at the moment, and add them as we go. the interface is marked internal, so adding method seems possible. (not sure if we even need an interface or should rather just have the concrete class)

Tests/Fixtures/BufferedOutput.php
```diff
@@ -0,0 +1,41 @@
...
+     * @var string
+     */
+    private $buffer = '';
+
+    public function getBuffer(): string
+    {
+        return $this->buffer;
+    }
+
+    public function write($messages, $newline = false, $options = self::OUTPUT_NORMAL): void
```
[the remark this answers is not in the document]
franmomu: we cannot, this method is overriden from `Output`

Command/CacheCommandTrait.php
franmomu: It was a little bit weird, I changed the method a little bit

Command/RemoveCacheCommand.php
franmomu: No idea, I added the protected property in both commands

Command/RemoveCacheCommand.php
franmomu: yes, I agree, so I removed these aliases

Tests/Fixtures/BufferedOutput.php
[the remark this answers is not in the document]
dbu: oh of course, sorry

Command/RemoveCacheCommand.php
dbu: thanks

Command/RemoveCacheCommand.php
```diff
@@ -20,56 +21,89 @@
...
 
-    public function __construct(CacheManager $cacheManager)
+    use CacheCommandTrait;
+
+    public function __construct(CacheManager $cacheManager, FilterManager $filterManager)
     {
-        $this->cacheManager = $cacheManager;
+        parent::__construct();
 
-        parent::__construct('liip:imagine:cache:remove');
```
dbu: do we only allow symfony versions that support `$defaultName`? otherwise we need to pass the protected static field to the constructor.

Command/CacheCommandTrait.php
dbu: thanks

Command/ResolveCacheCommand.php
franmomu: mmmm yes, but it looks like it does the [same thing than before (it stores the image if it does not exist)](https://github.com/liip/LiipImagineBundle/blob/master/Service/FilterService.php#L82), but with this option it doesn't check if it's already stored

Component/Console/Style/ImagineStyleInterface.php
franmomu: I felt the same about the interface, I'll take a look an see what can I do

Command/RemoveCacheCommand.php
franmomu: I think so, [it looks like it's supported since 3.4](https://github.com/symfony/symfony/pull/24609)

Command/RemoveCacheCommand.php
franmomu: in this case, I guess it make sense green because what you're trying to do is actually to remove cache, so it worked fine

Command/ResolveCacheCommand.php
franmomu: ``
Available foreground and background colors are: black, red, green, yellow, blue, magenta, cyan and white.
``
yep, white could be fine

Command/ResolveCacheCommand.php
dbu: that makes sense, given the name of this flag. can we tweak the description of the command and of this parameter to be a bit more explicit what this thing is doing?

Command/RemoveCacheCommand.php
dbu: you are probably right. red could make people think that there was a problem with those files.

Command/ResolveCacheCommand.php
```diff
@@ -21,68 +22,108 @@
...
-As a result you will get<info>
-    http://localhost/media/cache/thumb1/path1
-    http://localhost/media/cache/thumb2/path1</info>
-
-If you omit --filters parameter then to resolve given paths will be used all configured and available filters in application:
-<info>php app/console %command.name% path1</info>
-As a result you will get<info>
-    http://localhost/media/cache/thumb1/path1
-    http://localhost/media/cache/thumb2/path1</info>
+            ->setDescription('Displays and creates the cache entries for the given image sources and optional filters.')
```
dbu: i suggest "Warms up the cache for the specified image sources with all or specified filters applied, and prints the list of cache files."

Command/ResolveCacheCommand.php
dbu: we could formulate this as "Force generating the image and writing the cache, regardless of whether a cached version already exists."

Command/ResolveCacheCommand.php
```diff
@@ -21,68 +22,108 @@
...
+            ->addArgument('path', InputArgument::REQUIRED | InputArgument::IS_ARRAY,
+                'Image file path(s) to run resolution on.')
```
dbu: Image file path(s) for which to generate the cached images.

Command/ResolveCacheCommand.php
franmomu: perfect, thanks!

Command/ResolveCacheCommand.php
franmomu: much better, thank you

robfrawley:
@franmomu: After reviewing the many PRs you've resurrected in the past few months (all of which were in a state of extreme merge-conflict with the current git HEAD, to the point were no one else, even their original contributors like myself, had any interest/time/patience in resolving), I felt *entirely compelled* to take a moment and extend my sincere gratitude for the excellent work you put in to get #1220, #1219, #1201, and #1199 successfully merged. I honestly resigned myself to the fact that the respective original PRs would never be introduced into master.

The patience, effort, and man-hours required to merge these PRs deserves praise and I wanted to ensure that I took the time to do so (it's the very *least* I could do :-). Many of the previously stalled PRs that you resurrected include significant new features, bug fixes, and structural refactoring that the entire LiipImagineBundle community will benefit from.

Really great work. Thanks so much for your contributions to LiipImagineBundle!

franmomu:
Thank you very much for your kind words! You made my day :D Those PR had a lot of good work done and it was a pity to let them rot and also it was a way to learn :)

dbu:
I can only second what @robfrawley is saying, thanks a TON for all the effort you are putting into this @franmomu ! And thanks a lot @robfrawley for all the work you put into this bundle into the past! And thanks my co-worker at Liip @michellesanver who stepped up to maintain this bundle and waded through all those old issues last week! :green_heart:

Issue #1186, opened 2019-06-03
Version/milestones?


stephanvierkant:
It's been a long time since the last stable release of this bundle. The milestones aren't up to date right now: https://github.com/liip/LiipImagineBundle/milestones.

Can someone tag a new version or update the milestones, so we can track when a new version will be released?

garak:
I would also appreciate a new release, since Imagine has a lot of deprecations and I can't use its latest version (fixed in master, unreleased).

dbu:
changelog looks quite unmaintained. if someone could look over https://github.com/liip/LiipImagineBundle/compare/2.1.0...master and do a PR to add things to the changelog as necessary, that would be a great help.

dbu:
@michellesanver @franmomu wdyt, can we tag a release now, with symfony 4 no deprecations support?

michellesanver:
I think it would make sense

franmomu:
It looks good to me!

michellesanver:
2.2.0 was released this morning! (YAY! FRIDAY RELEASE PARTY! :D) 

Thanks everyone for your contributions. Means a lot to this library. And let me know if anything went under or I missed anything. Just getting into this library as its new maintainer 👍

PR #1223, opened 2019-09-10
update changelog

garak:
See https://github.com/liip/LiipImagineBundle/issues/1186#issuecomment-523774394

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | (none)
| License | MIT
| Doc PR | (none)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ for a given releases. Unreleased, upcoming changes will be updated here periodic
 [milestones](https://github.com/liip/LiipImagineBundle/milestones) page for the latest changes.
 
 ## [Unreleased](https://github.com/liip/LiipImagineBundle/tree/HEAD)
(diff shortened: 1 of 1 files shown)


dbu:
thanks, looks good to me, and thanks for helping out.

afaik @michellesanver wanted to look into whether there is something in the bundle to generate the changelog automatically. if there is we should rather not merge this to avoid confusion. but i did not see any tool configuration, manual curated changelog seems fine to me.

franmomu:
I've tried https://github.com/github-changelog-generator/github-changelog-generator and it works pretty well

michellesanver:
Thanks a lot! Helps to have an updated changelog. We use some automatic tools with git at Liip, but I think that this works well now.

PR #1224, opened 2019-09-12
added reference type to generate url

eroluysal:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | yes 
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

I want to use cdn with **imagine_filter** and **asset** methods. But **imagine_filter** returns absolute url everytime and asset ignores my defined assets url.  

I was added **referenceType** argument to **generateUrl** method of cache manager.

### DIFF
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -208,6 +208,21 @@ class CacheManager
     }
 
     /**
+     * BC Layer for Symfony < 4.3
+     *
+     * @param CacheResolveEvent $event
+     * @param string $eventName
+     */
+    private function dispatchWithBC(CacheResolveEvent $event, string $eventName): void
+    {
+        if ($this->dispatcher instanceof ContractsEventDispatcherInterface) {
+            $this->dispatcher->dispatch($event, $eventName);
+        } else {
+            $this->dispatcher->dispatch($eventName, $event);
+        }
+    }
+
+    /**
      * @see ResolverInterface::store
      *
      * @param BinaryInterface $binary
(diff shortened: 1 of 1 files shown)


Imagine/Cache/CacheManager.php
```diff
@@ -138,10 +138,11 @@ public function getRuntimePath($path, array $runtimeConfig)
      * @param string $filter        The name of the imagine filter in effect
      * @param array  $runtimeConfig
      * @param string $resolver
+     * @param int    $referenceType The type of reference to be generated (one of the constants)
      *
      * @return string
      */
-    public function generateUrl($path, $filter, array $runtimeConfig = [], $resolver = null)
+    public function generateUrl($path, $filter, array $runtimeConfig = [], $resolver = null, $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
```
dbu: this is a BC break. the default value must be the same as was hardcoded before: ABSOLUTE_URL

Imagine/Cache/CacheManager.php
```diff
@@ -138,10 +138,11 @@ public function getRuntimePath($path, array $runtimeConfig)
      * @param string $filter        The name of the imagine filter in effect
      * @param array  $runtimeConfig
      * @param string $resolver
+     * @param int    $referenceType The type of reference to be generated (one of the constants)
```
dbu: lets say "(one of the UrlGenerator constants)" to be more specific about what we mean.

ossinkine:
@eroluysal I guess you make something like `asset('path'|imagine_filter('filter'))`, so you need a relative path. But you can use ProxyResolver, it can override the origin domain to your CDN domain.

Issue #954, opened 2017-06-21
resizing don't coop with auto_rotation


annamakowskaespeo:
| Q | A
| --- | ---
| Bug Report? | yes
| Feature Request? | no
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | 1.8.0

Hi! I'm trying to auto rotate image and then resize it. My problem is, that resizing filters (I tried many) don't coop with auto_rotate filter. 

For example, I have taken vertical photo by phone. This photo is saved as horizontal image 640x480 that should auto rotate to vertical.  I want to make thumbnail, that will be 120px width and will be displayed correctly (vertical). 

With configuration:
```
liip_imagine:
    driver: imagick
    resolvers:
        default:
            web_path: ~

    filter_sets:
        cache: ~
        thumbi_small:
            quality: 100
            filters:
                auto_rotate: ~
                relative_resize: { widen: 120 }
```

I get image that is displayed vertical (as I wanted), but the width is 90px (height is 120px). What is surprising, this image is not physically auto_rotated - it's just displayed correctly (without auto_rotate it's displaying wrong).
How can I solve this?

example image: [image](https://soma-dev.espeo.pl/uploads/Items/2/0bc8b123eef3e1b99585779a5ab53c3f.JPG)

I have exif installed.
I have tried gd liblary, but auto_rotate is not working with it
I'm using imagick module version => 3.4.3

I have also tried configuration:
```
filters:
    auto_rotate: ~
    thumbnail: { size: [120, 200], mode: inset }
```
```
filters:
    auto_rotate: ~
    scale:
        dim: [ 120, 200 ]
```

JakubKontra:
+1 same problem

robfrawley:
@annamakowskaespeo @JakubKontra Has this problem persisted?

JakubKontra:
yes @robfrawley

belabed90:
same problem

franmomu:
mmm it works to me, I created a small project https://github.com/franmomu/liip-imagine-bundle-examples to see it

belabed90:
Sorry, The problem was in my php server, exif_read_data function was not enabled.

i think the filter must throw an exception when there is such things!

dbu:
good idea! @belabed90 please do a pull request to add such a check in the filter class to have it throw an understandable exception.

michellesanver:
Is this still an issue with #1230 where that filter class was removed in favour of using imagine directly?

[...]

Issue #741, opened 2016-05-27
Auto rotate problem in real server


jekeyeke:
Hi everyone.

I have a problem with auto_rotate option.
In local, it works perfect but when I upload the code to the real server, the auto_rotate option doesn't work.
I'm working with Symfony and i don't know which is the problem.

This is my code:

```
liip_imagine:
    driver: gd
    cache:  default
    data_loader:   default

    resolvers:
        default:
            web_path:
                web_root: %kernel.root_dir%/../web
                cache_prefix: media/images/cache
    filter_sets:
        original:
            quality: 75
            filters:
                auto_rotate: ~
                thumbnail: { size: [1200, 1200], mode: inset }
```

I appreciate any help.

Thanks

BartVB:
If this is working locally and not on the server then chances are that you're using an outdated version of GD on the server. Could you check the versions?

BTW any particular reason why you are using GD and not one of the other libs?

davidromani:
Same problem here.

Working local machine (OSX 10.10 + PHP 5.6.19) config:
```
$ php -i | grep -i --color gd
GD Support => enabled
GD Version => bundled (2.1.0 compatible)
gd.jpeg_ignore_warning => 0 => 0
```

**Not working** server machine (Ubuntu 14.04 + PHP 5.5.9) config:
```
GD Support => enabled
GD Version => 2.1.1-dev
gd.jpeg_ignore_warning => 0 => 0
```

lsmith77:
Yeah I think this is more likely a GD issue.

https://twitter.com/PierreJoye/status/754939618982699008
"@lsmith some cases have been improved in 2.2, and quality much better in master.
Please try using libgd master. Ideally open an issue there"

can you contact the GD maintainers about this?

lsmith77:
if someone can setup a reproduce-able test case Pierre offered to have a look

jekeyeke:
Hi everyone,
@BartVB when i started in the project, liip was already installed so i continued since there.

Working local machine (Ubuntu 14 + PHP 5.5.9) config:
```
$ php -i | grep -i --color gd
GD Support => enabled
GD Version => bundled (2.1.0 compatible)
gd.jpeg_ignore_warning => 0 => 0
```

Not working server machine (Debian 3.2.68 + PHP 5.5.25) config:

```
GD Support => enabled
GD Version => bundled (2.1.0 compatible)
gd.jpeg_ignore_warning => 0 => 0
```

Could you tell me another lib to use?

[...]

Issue #737, opened 2016-05-19
Automatic rotation in conjunction with thumbnails


pscheit:
Hey Liip,

thank you so much for the bundle.

I have a issue with an not-yet-rotated image, which I try to thumbnail. Here's the config:

``` yaml
  filter_sets:
    xs:
      quality: 75
      filters:
        auto_rotate: ~
        thumbnail: { size: [120, 90], mode: inset }
```

My Image Size is: 2592x1944, so its landscape, but should be rotated to portrait 1944x2592
I have another Picture which is already rotated which is 1944x2592, too.

The thumbnails produces from the non-yet-rotated photo gives me: 90x120
And the already rotated one (portrait) gives me: 68x90 (which is the expected result).

I thought auto_rotate would apply first and then the thumbnail filter? Is my yaml wrong?

``` yaml
  filter_sets:
    xs:
      quality: 75
      filters:
        thumbnail: { size: [120, 90], mode: inset }
        auto_rotate: ~
```

this configuration gives me exactly the same results. So it looks like no matter what, its always thumbnailed first and then auto-rotated. Which isn't what I have expected. Are my expectations wrong?

BartVB:
I'm using a similar config which works as expected:

```
liip_imagine:
    driver: imagick
    filter_sets:
        thumb:            
            quality: 75
            filters:
                auto_rotate: ~
                thumbnail: { size: [120, 90], mode: outbound }   
```

No idea why this isn't working for you. The documentation mentions that you should run the auto_rotate filter as early as possible:
http://symfony.com/doc/master/bundles/LiipImagineBundle/filters.html#the-auto-rotate-filter

You could try using a different driver? This might solve the issue (but I'm only guessing here).

lsmith77:
maybe related to https://github.com/liip/LiipImagineBundle/issues/741

michellesanver:
Closing this issue because of #1230

More details: https://github.com/liip/LiipImagineBundle/pull/1230#issuecomment-537934662

If it's still an issue for you, please open a new PR, and thanks a lot :)

PR #1230, opened 2019-10-03
Use Autorotate Filter from Imagine library

franmomu:
| Q | A
| --- | ---
| Branch? |2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

When I was taking a look at ``AutoRotateFilterLoader``, I've realised that this was also implemented in the [Imagine library](https://github.com/avalanche123/Imagine/blob/develop/src/Filter/Basic/Autorotate.php). So this commit removes the implementation of this bundle to rely on the library one. I ended up removing the test because it's just a call, like ``StripFilterLoader``.


Imagine/Filter/Loader/AutoRotateFilterLoader.php
```diff
@@ -20,90 +21,14 @@
...
-
-            // Flips if necessary.
-            if ($this->isFlipped($orientation)) {
-                $image->flipHorizontally();
-            }
-        }
+        $filter = new Autorotate();
+        $image = $filter->apply($image);
 
         return $image;
```
dbu: could be shortened to `return $filter->apply($image);`

franmomu:
Also there are some issues related with automatic rotation that are not really solved, but apparently could be about having `exif_read_data` enabled on the server.

https://github.com/liip/LiipImagineBundle/issues/737 https://github.com/liip/LiipImagineBundle/issues/532 https://github.com/liip/LiipImagineBundle/issues/741 https://github.com/liip/LiipImagineBundle/issues/954

and I guess that after this, if there should throw an exception or not, should be part of Imagine instead of the bundle.

franmomu:
> good catch, i am all in favor of this.
> 
> i could see keeping a single test case as a kind of "integration test" to notice if imagine does a BC break or something like that.

I was thinking what to do... what about just this:
```php
$this->assertTrue(class_exists(Autorotate::class));
```
Because at first I thought about checking a call to `metadata`, but I don't like checking the implementation of the library.

dbu:
> $this->assertTrue(class_exists(Autorotate::class));

i would like us to instantiate the autorotate and call filter on it, to see things are wired together properly. while i agree we get into the grey area of testing the upstream library, i think testing the integration is useful, and if the upstream library is misbehaving, we do care as our users will possibly complain here rather than figure out that its a library problem.

franmomu:
> Yay! Thanks for this MR! So much less code, it's the best kind.

It feels always good to remove code 😬

robfrawley:
The upstream implementation only checks for the `ifd0.Orientation` EXIF tag, whereas our prior implementation also checked for the `exif.Orientation` tag. The inclusion of the latter check was originally merged via #476 without any specific comments about its use, so I have absolutely no idea as to whether the check for the `exif.Orientation` tag resolved some previously reported issue(s) that are still relevant, is completely unnecessary and never served any useful purpose, or only applies to an insignificant number of EOF/legacy/non-compliant EXIF metadata writers and can safely be ignored moving forward.

To be completely clear, I am not recommending reverting this PR; we should absolutely be using the upstream implementation whenever possible (only creating our own duplicated logic in the face of *severe* upstream deficiencies with critical user-facing consequences). I only point this out to document the change (such that it can be properly searched for in case it does cause any issues) and to bring it to the maintainers' attention so that **if** the `exif.Orientation` tag (or related autorotation issues) show up in new issues we can head on over to [github.com/avalanche123/Imagine](https://github.com/avalanche123/Imagine) and submit an upstream PR to bring back the secondary check there.

Otherwise, as to the entirety of this PR, removed code is always great; noone ever complained about less code to maintain! :-)

Great PR @franmomu!

franmomu:
Thank you @robfrawley, thanks for pointing out about the `exif.Orientation`, good catch.

michellesanver:
Thanks @robfrawley!

PR #1059, opened 2018-02-20
[Cache Resolver] Implement relative web path resolver

nmuntyanov:
#| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1050
| License | MIT

Added new path resolver for relative paths creation 
(RelativeWebPathResolver && RelativeWebPathResolverFactory) 

usage example: 
```php
liip_imagine:
  resolvers:
     default:
       relative_web_path:
         web_root:  "%kernel.root_dir%/../web"
         cache_prefix: assets
```

PR #1233, opened 2019-10-05
[WIP][Cache Resolver] Implement relative web path resolver

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

This is https://github.com/liip/LiipImagineBundle/pull/1059 merged with the current master, I fixed the tests and conflicts, ~~but I haven't checked the implementation yet~~.

EDIT:
- The constructor of `WebPathResolver` shouldn't be change in case someone has extended this class.
- I would just instantiate the `PathResolver` inside the class and I'm not sure if the `PathResolverInterface` is needed.


franmomu:
Hi, I don't think someone is working on it, feel free to take over. Based on the comments I wrote:

> The constructor of WebPathResolver shouldn't be change in case someone has extended this class.

This should be done in a BC-break way.

> I would just instantiate the PathResolver inside the class and I'm not sure if the PathResolverInterface is needed.

Check if it's really needed the interface.

Issue #1236, opened 2019-10-07
Remove command and delete all images


kalypso63:
### Preconditions

1. liip/imagine-bundle

### Steps to reproduce

1. Delete all image via php bin/console liip:imagine:cache:remove

### Expected result

1. Delete all existing image in any filter

### Actual result

1. Path is required by default

bobvandevijver:
I found this as well: the documentation states that this is possible though (https://github.com/liip/LiipImagineBundle/blob/master/Resources/doc/commands.rst)

franmomu:
It looks like it was after https://github.com/liip/LiipImagineBundle/pull/1222, not sure if I can take a look this weekend.

PR #1239, opened 2019-10-26
Fix RemoveCacheCommand with empty path

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1236
| License | MIT
| Doc PR | 

In https://github.com/liip/LiipImagineBundle/pull/1222 the `path` for `RemoveCacheCommand` was made mandatory, so there is no way to remove all images and filters as before. 

As paths can be multiple I changed it back to `paths`. ~I don't like very much the solution because if you execute the command you would get:~

```
 - [OKAY] Completed 0 removals (0 images, 7 filters)
```
~which is a little bit weird, so~ any suggestion is more than welcome.

Now it says:
```
 -
 - [OKAY] Completed 7 removals (7 filters)
 -
```

### DIFF
--- a/Command/CacheCommandTrait.php
+++ b/Command/CacheCommandTrait.php
@@ -60,7 +60,7 @@ trait CacheCommandTrait
     private function resolveInputFiltersAndPaths(InputInterface $input): array
     {
         return [
-            $input->getArgument('path'),
+            $input->getArgument('paths'),
             $this->normalizeFilterList($input->getOption('filter')),
         ];
     }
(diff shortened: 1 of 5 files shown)


michellesanver:
Thanks a lot for this fix! :)

Issue #1172, opened 2019-04-19
Image cannot be resolved if contains query parameters


proArtex:
Stream loader via http/https may contain query params that are omitted during cache resolve.

### Preconditions

1. Symfony 4.2.7
2. liip/imagine-bundle: 2.1.0
3. imagine/imagine: v0.7.1

### Steps to reproduce
```
# services.yaml

Liip\ImagineBundle\Binary\Loader\StreamLoader:
        arguments:
            $wrapperPrefix: ''
        tags:
            - { name: 'liip_imagine.binary.loader', loader: 'remote_loader' }
```

```
# packages/liip_imagine.yaml

liip_imagine:
    driver: imagick
    resolvers:
        default:
            web_path:
                cache_prefix: '%thumbnails_path%'
    filter_sets:
        user_640x640:
            data_loader: remote_loader
            filters:
                strip: ~
                upscale:
                    min: [640, 640]
                thumbnail:
                    size: [640, 640]
                    mode: outbound
```

```
# template.twig

<img src="{{ 'https://pp.userapi.com/Ezx-RHE3h-pOqTrvqi3jrWnYGWTNW-F_yG5uoA/Wppvp-5waFQ.jpg?ava=1' | imagine_filter('user_640x640') }}">
```

### Expected result
Resolved image /thumbnails/user_640x640/https://pp.userapi.com/Ezx-RHE3h-pOqTrvqi3jrWnYGWTNW-F_yG5uoA/Wppvp-5waFQ.jpg?ava=1

### Actual result
Exception "No route found for "GET /thumbnails/user_640x640/https://pp.userapi.com/Ezx-RHE3h-pOqTrvqi3jrWnYGWTNW-F_yG5uoA/Wppvp-5waFQ.jpg"

michellesanver:
Thanks for the report. An easy fix would be to strip query parameters from the URL, I think this would be a good first issue for someone to pick up.

ClementGautier:
Hi :wave:,

I was trying to make a fix for that but it's unclear to me what is the expected result here. I tried to reproduce and the query param is correctly stripped. Is the expected behavior to keep the query param?

I tried to render the following
```
<img src="{{ 'http://via.placeholder.com/640x360?ava=1' | imagine_filter('user_640x640') }}">
<img src="{{ 'http://via.placeholder.com/640x360' | imagine_filter('user_640x640') }}">
```

and got the renderer result
```
[...]

Issue #1116, opened 2018-07-29
assets with version configured


TiMESPLiNTER:
### Preconditions

1. Set a version under `framework.assets.version`
2. Require the image using like `asset(i.imagePath) | imagine_filter('item_list')` in twig

### Steps to reproduce

1. View a page where the image is embedded like mentioned above
2. The url then is `http://exmaple.com/media/cache/resolve/filter/IMG_6776.jpg%3Fv5`
3. The imagine twig extension then tries to find the file `IMG_6776.jpg%3Fv5` which obviously does not exist

### Expected result

1. It should work and any query parameters should be ignored

### Actual result

1. The image can not be displayed

rejinka:
This also filters away the scheme. We use gaufrettes StreamWrapper to reference files like app://images/some_image.jpg

PR #1240, opened 2019-11-18
Revert "Only use actual path without any query parameters from the url"

GKFX:
This reverts commit b77558c98e389bef3a642f4a03745b47925b90d4.
This commit broke backwards compatibility with code which wanted to use
more than just the path component of a URL (e.g. wanted to use full URLs
with a domain via `Liip\ImagineBundle\Binary\Loader\StreamLoader`). This was released on upgrade from 2.1.0 to 2.2.0 so if the project follows semver then I don't think this should have happened.

It's not clear from the commit history why a need was felt to strip query parameters, but if a particular user needs it then this should probably be done by them?

Sample project using this was [camdram/camdram](https://github.com/camdram/camdram/blob/56517df4efe2678817eba5c33da6e56d0390ee11/app/config/services.yml#L56).

| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes/no
| Fixed tickets | 
| License | MIT
| Doc PR |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ for a given releases. Unreleased, upcoming changes will be updated here periodic
 - Fix Symfony 4.2 Deprecation Warnings ([hjanuschka](https://github.com/hjanuschka))
 - Fix special characters encoding in URL path ([dbalabka](https://github.com/dbalabka))
 - Update imagine/imagine dependency to 1.1 ([maximgubar](https://github.com/maximgubar))
-- Only use actual path without any query parameters from the url ([maximgubar](https://github.com/maximgubar))
+- __\[BC BREAK\]__ Only use actual path ([parse_url's](https://www.php.net/manual/en/function.parse-url.php) PHP_URL_PATH) without any query parameters from the url ([maximgubar](https://github.com/maximgubar))
 - __\[Dependency Injection\]__ Add aliases for data and filter manager ([fpaterno](https://github.com/fpaterno))
 - Use Autorotate Filter from Imagine library ([franmomu](https://github.com/franmomu))
 - Fix Mime deprecations for Symfony 4 ([franmomu](https://github.com/franmomu))
--- a/Templating/FilterTrait.php
+++ b/Templating/FilterTrait.php
@@ -40,7 +40,7 @@ trait FilterTrait
      */
     public function filter($path, $filter, array $config = [], $resolver = null)
     {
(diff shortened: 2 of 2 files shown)


michellesanver:
For context, this commit comes from this issue: https://github.com/liip/LiipImagineBundle/issues/1116

dbu:
it sounds to me like the problem does exist and simply reverting would be a regression for people who expect the version thing to not break the file format detection.

it smells to me like the fix is in the wrong place, or the semantics of what should be passed is incomplete. i guess the correct thing would be to have the implementation of getBrowserPath figure out if it needs to process the url, or the component that later handles the result that is returned by this method.

michellesanver:
@GKFX Can you remove the commit reverting the change in the filter trait? Then I'll merge the BC break documentation, and we have to see in #1172 how we fix this issue for the next release.

Schyzophrenic:
I just see this now after submitting #1291 about the very same line. 
The problem with b77558c98e389bef3a642f4a03745b47925b90d4 is that it strips out the domain and will therefore breaks all absolute remote url. I assume this is one of the reason why people are struggling with the current documentation as it cannot work as it is. 
In addition, if you just revert to the previous code, the query parameters will be kept but the rename file process will fail as it will try to rename with the query parameters as well (e.g. something.png?v=1321321). 
The more I think about it, the more I wonder whether we should revert b77558c98e389bef3a642f4a03745b47925b90d4 and just ensure the renaming is done correctly. 

Side note: I am new to this bundle so let me know if I am awfully wrong! Thank you

dbu:
i wonder if we could ask other places to strip query parameters, e.g. the controller actions, instead of doing it in this central place.

dbu:
@GKFX looking at #1291 which is doing something similar, i looked into the code and say:

i spent some time with a collegue to try to understand what is going on.

unless we are completely mistaken, when the image is not in the local cache, it means that we provide the full image url in the path of the ImagineController::filterAction which then proceeds to download the image and cache it with the filter applied. i think that LiipImagineBundle should not support this - it would allow an attacker to provide the controller with an endless number of images from other domains. that way one could place all kinds of illegal content into the cache, or overload the cache.

did i misunderstand, or is this how the flow works?

i will look at documenting this better, but to me the change was a bug/security fix that had side effects for your use case. as it has been this way for a long time now, i think reverting it would not make sense.

dbu:
i created #1393 to discuss the security implications of loading images from anywhere

dbu:
fixed in #1397

PR #1255, opened 2019-12-28
Eliminate root_dir usage

Guite:
| Q | A
| --- | ---
| Branch? | symfony5
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

Replaced `root_dir` by `project_dir`. Not sure whether this is the correct way you want to do it, but seems working so far.

### DIFF
--- a/Imagine/Filter/Loader/PasteFilterLoader.php
+++ b/Imagine/Filter/Loader/PasteFilterLoader.php
@@ -25,12 +25,12 @@ class PasteFilterLoader implements LoaderInterface
     /**
      * @var string
      */
-    protected $rootPath;
+    protected $projectDir;
 
-    public function __construct(ImagineInterface $imagine, $rootPath)
+    public function __construct(ImagineInterface $imagine, $projectDir)
     {
         $this->imagine = $imagine;
-        $this->rootPath = $rootPath . '/app';
+        $this->projectDir = $projectDir;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


franmomu:
[The docs of watermark](https://github.com/liip/LiipImagineBundle/blob/master/Resources/doc/filters/general.rst#watermark) should also be updated.

Imagine/Filter/Loader/PasteFilterLoader.php
```diff
@@ -30,7 +30,7 @@ class PasteFilterLoader implements LoaderInterface
     public function __construct(ImagineInterface $imagine, $rootPath)
     {
         $this->imagine = $imagine;
-        $this->rootPath = $rootPath;
+        $this->rootPath = $rootPath . '/app';
```
franmomu: IMO this could be called $projectDir and do not append `/app` and then in the `load` function I would do something like:

```php
$file = $this->projectDir.'/'.$options['image'];

if (!file_exists($file) {
    @trigger_error(
        'The ' . $file . ' does not exists, change the path based on kernel.project_dir parameter',
        E_USER_DEPRECATED
    );
    $file = $this->rootPath.'/app/'.$options['image']
}
```

Guite:
@franmomu added both changes

PR #1258, opened 2020-01-03
Cache warmers

gklvns:
This is a feature, used by the company I work for. 
Hopefully, this feature ends up in the master branch.

Steps to create a cache warmer:
1. Create a service that implements `WarmerInterface` and warms up the images (descriptions for methods can be found in the interface);
2. Tag it with `liip_imagine.cache.warmer`. Tag also requires a `warmer` parameter, which contains the name of the cache warmer. It can be used in the console command (read the next point);
4. The console command `liip:imagine:cache:warm` finds all cache warmers and runs them, but you can also specify specific cache warmers separated by space: `liip:imagine:cache:warm products_warmer brands_warmer`. Run `liip:imagine:cache:warm --help` to read more.

Service example:
```yaml
app.products_warmer:
    class: AppBundle\Imagine\ProductImagesWarmer
    tags:
        - { name: 'liip_imagine.cache.warmer', warmer: 'products_warmer' }
```

Filter example:
```yaml
product_thumb_new_v2:
    quality: 85
    filters:
        thumbnail: { size: [450, 301], mode: inset }
        background: { size: [450, 301], color: '#FFFFFF' }
    warmers: [products_warmer]
    post_processors:
        jpegoptim: {}
```

If the description is missing information, it is misleading or hard to understand, let me know, so I can update it.


dbu:
hi @keshancs , i am going through the backlog of older issues and merge requests.

thanks for the pull request. code looks good to me, but can you explain what the use case is? i would expect that usually, people keep the cached images between deployments. are you deploying image asset files that you convert with LiipImagineBundle to different sizes?

as this is a significant amount of code, i wonder if it would be better to make a separate bundle out of this. i would be happy to mention it in the documentation for people that have the same use case.

PR #1259, opened 2020-01-06
Symfony 5 to master

michellesanver:
Time for 2.3.0 release! :)


Command/CacheCommandTrait.php
```diff
@@ -60,7 +60,7 @@ private function setupOutputStyle(InputInterface $input, OutputInterface $output
     private function resolveInputFiltersAndPaths(InputInterface $input): array
     {
         return [
-            $input->getArgument('path'),
+            $input->getArgument('paths'),
```
dbu: is this a bugfix? if not, this could be a BC break

dbu: ah sorry, its an argument, not an option! so no name on the cli that the user of the command types.

Issue #1007, opened 2017-11-17
Allow paste multiple watermarks


Koc:
| Q | A
| --- | ---
| Bug Report? | no
| Feature Request? | yes
| BC Break Report? | no
| RFC? | no
| Imagine Bundle Version | latest

For now it is impossible insert multiple watermarks because filter name is key of the array.

peter-gribanov:
You can solve this problem by creating a additional filter

```yml
services:
    liip_imagine.filter.loader.watermark2:
        class: '%liip_imagine.filter.loader.watermark.class%'
        arguments: [ '@liip_imagine', '%kernel.root_dir%' ]
        tags:
            - { name: liip_imagine.filter.loader, loader: 'watermark2' }
```

and apply it

```yml
liip_imagine:
    filter_sets:
        my_watermark_filter:
            filters:
                watermark:
                    image: 'assets/images/watermark1.png'
                    position: 'topleft'
                watermark2:
                    image: 'assets/images/watermark2.png'
                    position: 'bottomright'
```
But this does not eliminate need to resolve the problem.

dbu:
seems fixed in #1281

PR #1281, opened 2020-05-07
Extend WatermarkFilter with multiple option

tzemp:
Signed-off-by: Timothy Zemp <timothy.zemp@gmail.com>

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1007 
| License | MIT
| Doc PR | no

Extends the watermark functionally to allow multiple watermarks of the same type. It tries to paste the watermark onto the image as often as possible.

I'm not sure if "position" is the right way to address this problem, but i can be solved within few lines of code. It can also uses the "size" attribute -> 0.25% -> 4 watermarks per width.

Maybe some validation is required to prevent sizes > 1.0, as this is not validated yet.

Feedback appreciated :-)

### DIFF
--- a/Imagine/Filter/Loader/WatermarkFilterLoader.php
+++ b/Imagine/Filter/Loader/WatermarkFilterLoader.php
@@ -70,6 +70,21 @@ class WatermarkFilterLoader implements LoaderInterface
             $watermarkSize = $watermark->getSize();
         }
 
+        if ('multiple' === $options['position']) {
+            // we loop over the coordinates of the image to apply the watermark as much as possible
+            $pasteX = 0;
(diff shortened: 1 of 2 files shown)


dbu (approved):
cool!

peter-gribanov:
As far as i understand, the watermark is paved in the picture over its entire area. Perhaps the option should have been named "pave"?

Problem #1007 talks about adding several different watermarks to one picture with the ability to choose the location of the watermarks, and not about pave the entire picture with one watermark.

PR #1282, opened 2020-05-31
Fix symfony 5.1 deprecations

mynameisbogdan:
| Q | A
| --- | ---
| Branch? | 2.3
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |  #1287 
| License | MIT
| Doc PR |

### DIFF
--- a/Resources/config/templating.xml
+++ b/Resources/config/templating.xml
@@ -7,7 +7,7 @@
         <service id="liip_imagine.templating.filter_helper" class="Liip\ImagineBundle\Templating\Helper\FilterHelper">
             <tag name="templating.helper" alias="imagine" />
             <argument type="service" id="liip_imagine.cache.manager" />
-            <deprecated>The "%service_id%" service is deprecated since LiipImagineBundle 2.2 and will be removed in 3.0.</deprecated>
+            <!-- deprecated service, deprecation is set in the container extension -->
         </service>
     </services>
 </container>
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "symfony/filesystem": "^3.4|^4.3|^5.0",
         "symfony/finder": "^3.4|^4.3|^5.0",
         "symfony/framework-bundle": "^3.4|^4.3|^5.0",
+        "symfony/mime": "^4.3|^5.0",
         "symfony/options-resolver": "^3.4|^4.3|^5.0",
         "symfony/process": "^3.4|^4.3|^5.0",
         "symfony/templating": "^3.4|^4.3|^5.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,10 @@ matrix:
       env:
         - SYMFONY_VERSION=5.0.*
 
+    - php: '7.2.9'
+      env:
+        - SYMFONY_VERSION=5.1.*
+
     # Remember to test upcoming Symfony versions below with lowest supported PHP version and dev dependencies
 
     # Test upcoming PHP versions with dev dependencies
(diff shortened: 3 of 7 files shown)


Resources/config/templating.xml
```diff
@@ -7,7 +7,7 @@
         <service id="liip_imagine.templating.filter_helper" class="Liip\ImagineBundle\Templating\Helper\FilterHelper">
             <tag name="templating.helper" alias="imagine" />
             <argument type="service" id="liip_imagine.cache.manager" />
-            <deprecated>The "%service_id%" service is deprecated since LiipImagineBundle 2.2 and will be removed in 3.0.</deprecated>
+            <!-- deprecated service, deprecation is set in the container extension -->
```
Pixelshaped, line 10: As per Symfony deprecation message, couldn't this have been fixed by just adding "package" and "version" attributes to the `<deprecated>` node?

mynameisbogdan: That would work for Symfony 5.1+, but fail for older versions. That was also my initial solution and changed to this.

Pixelshaped: That's a bummer, but your solution works. The PR looks fine to me.

mynameisbogdan: Thank you!

Pixelshaped:
@dbu please consider merging this PR

composer.json
```diff
@@ -24,6 +24,7 @@
         "symfony/filesystem": "^3.4|^4.3|^5.0",
         "symfony/finder": "^3.4|^4.3|^5.0",
         "symfony/framework-bundle": "^3.4|^4.3|^5.0",
+        "symfony/mime": "^4.3|^5.0",
```
dbu, line 27: did we miss to explicitly require this, and symfony 5.1 drops some indirect requiring of it?

mynameisbogdan: as you can see [here](https://github.com/symfony/symfony/commit/11cef32ff7d6cf66f74d5e25a41ab2afa751492f) symfony/mime was set as an optional dependency so it's needed to explicitly require it since SimpleMimeTypeGuesser depends on it.

Pixelshaped: There is a related issue, might I add: https://github.com/liip/LiipImagineBundle/issues/1287

mynameisbogdan: that's correct.

dbu: (i am not active on this project, but have merge rights as i work at liip, so want to help out to merge this)

thanks for the pointers! reading #1287, i am unsure how this PR fixes the issue? don't we also need to change to use the new or old mime type class depending on which is available?

mynameisbogdan: This is already done in `SimpleMimeTypeGuesser`. The only issue is that since now `symfony/mime` is set as an optional dependency for `symfony/http-foundation` we need to explicitly require it. So far till symfony 5.1 this wasn't required since it was pulled from the main `symfony/framework-bundle` dependency.

We highly appreciate your help.

dbu: thanks for the explanation!

dbu:
are you aware of anything else that should be merged before i tag a new version? i merge a couple of other cleanups about composer requirements and documentation cleanups.
i intend to merge #1290 if there is no build error, and then tag the release.

mynameisbogdan:
No, sadly I'm not aware.

dbu:
i am more happy than sad that there are no blockers :wink: 

here we go: https://github.com/liip/LiipImagineBundle/releases/tag/2.3.1 :rocket:

Issue #328, opened 2014-02-14
Does LiipImagineBundle support remote image? 


saharabear:
I have try the following code: 

```
{{'http://www.other-domain.com/img.jpg'|imagine_filter('face')}}
{{'http://www.my-domain.com/my-img.jpg'|imagine_filter('face')}}
```

LiipImagineBundle does not generate cache, but I tried the following code: 

```
{{'/my-img.jpg'|imagine_filter('face')}}
```

Now it works. 

Any way to process pictures with absolute URL or remote pictures? 

Thanks.

flug:
@makasim Is there a reason why this is not built-in the bundle ? This would be a great feature. I could contribute it if necessary.

makasim:
@flug Could you please remember me what problem you are trying to solve?

greg0ire:
@makasim he was trying to get this bundle to work with remote images, gave up and wrote this : https://github.com/flug/imagine-bundle

makasim:
sorry, but I am trying to understand the real problem which you all guys want to solve. It is not clear too me. the code posted is not looking good and clear too. So please describe the problem so I can help you with the solution.

greg0ire:
We have images stored on a remote server. We want to resize them and store them locally.

makasim:
Try https://github.com/liip/LiipImagineBundle/blob/master/Binary/Loader/StreamLoader.php loader. I believe it must fit your needs.

greg0ire:
We'll have a look at that, thanks a lot !

aistis-:
Could someone post a clean solution with `StreamLoader`?

ibasaw:
the solution of : @mechanicalgux doesn't work anymore. How to use CDN url absolute path asset with filter ?

    image.absolutepath|imagine_filter('my_filter_medium')

maxlipsky:
@ibasaw You need define StreamLoader as service:
```
app.imagine.cdn_data_loader:
    class: "%liip_imagine.binary.loader.stream.class%"
    arguments:
        - '' # if you store full path in the database
    tags:
        - { name: 'liip_imagine.binary.loader', loader: 'cdn_data_loader' }
```

And to config:

```
filter_sets:
    my_filter_medium:
        data_loader: cdn_data_loader
        filters: # your filter
```

xmon:
@maxlipsky I can not run the code, do you know of any place where there is a complete functional example? 

thanks a lot

alainib:
[...]

PR #1291, opened 2020-07-07
Fix Remote Files and allow compatibility with versioning of assets (?v=123) 

### LINKED ISSUE #936: Working example for remote images via http?
Hello,

I need an working example for symfony 3.2 with custom data loader and custom cache. I've found nothing about that in documentation. My project needs to fetch the images through http [file_get_contents($remotePath)] and then cache them localy, before displaying it. 

Can anyone provide such an example please?

Schyzophrenic:
| Q | A
| --- | ---
| Branch? | 2.4
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #936 #328 #1116 
| License | MIT
| Doc PR | No

Potentially fixes #936 #328 #1116 .
Please note I am not familiar enough with the bundle to completely grasp the implications. Please double check this PR before merging :)
The previous implementation was stripping the server name from the path, preventing any remote loading to occur.


glouton:
Hi all,

I have to agree with @Schyzophrenic as I spent hours trying to figure out why my stream loader refused to work before finding the uncaught NotLoadableException in dev.log and noticing the protocol and domain name were not there.

The parse_url($path, PHP_URL_PATH) explain it indeed.

I also like to know why this has been changed and, if there is no good reason, to have this fix merge.

Regards,

dbu:
the commit b77558c98e389bef3a642f4a03745b47925b90d4 that changed this  talked about wanting to remove query parameter, i guess it was not aware that this could also have a domain

Templating/FilterTrait.php
```diff
@@ -42,7 +42,7 @@ public function __construct(CacheManager $cache)
      */
     public function filter($path, $filter, array $config = [], $resolver = null, $referenceType = UrlGeneratorInterface::ABSOLUTE_URL)
     {
-        return $this->cache->getBrowserPath(parse_url($path, PHP_URL_PATH), $filter, $config, $resolver, $referenceType);
+        return $this->cache->getBrowserPath($path, $filter, $config, $resolver, $referenceType);
```
dbu, line 41: as the parse_url was added in b77558c98e389bef3a642f4a03745b47925b90d4 to remove query parameter, could we keep removing those but without losing the domain? otherwise we trade the fix for a regression in other cases, i think.

Schyzophrenic: I actually think we should not be removing the query parameters when getting the image. Indeed, some image provider will actually serve images based on them (could include size, path, etc). Nonetheless, parameters should be stripped at some points and this is why I also changed the WebPathResolver to take this into account. 

I am happy to revisit this if needed as I do believe that allowing handling remote files is a critical feature.

glouton: > I actually think we should not be removing the query parameters when getting the image. Indeed, some image provider will actually serve images based on them (could include size, path, etc). Nonetheless, parameters should be stripped at some points and this is why I also changed the WebPathResolver to take this into account.
> 
> I am happy to revisit this if needed as I do believe that allowing handling remote files is a critical feature.

Agreed! And it's actually the conclusion of the discussion here :
https://github.com/liip/LiipImagineBundle/issues/1172#issuecomment-554983445

dbu:
this seems reasonable to me, but i don't have a deep enough overview of all the parts and bits of this bundle to tell if it will again explode somewhere else if we do this. 

and clearly there are not enough tests for all use cases, or we would have noticed issues in the previous pull requests.

i am unsure what the best course of action is. maybe merge and make it a new minor version and see if then other things explode again because of the change?

this would also replace #1240, right?

Schyzophrenic:
> this would also replace #1240, right?
It should replace it indeed. In terms of testing, I don't have a lot of experience in this field, but it would indeed make sense to have at least:
- Local file
- Remote file (http://mydomain.com/image.png)
- Remote file (http://mydomain.com/image?name=path.png&additionalValues)

Of course, it means the cached image may be generated multiple times if the query parameters are different, but I believe this is the logical way of doing it. It will be up to the user to strip out any non required parameters if needed. 
@glouton any thoughts on that?

glouton:
I fully support this point en view @Schyzophrenic 

In my humble opinion, as it has been stated in the issue about query params, Liip Imagine Bundle job is to transform an image.
It's not up to it to decide whether the path of the resource we feed it with is ok or not.
- Hey Liip, please transform the resource at this location
Then
- Ok, I found it and cached the result
Or
- Sorry, couldn't find it

And that's it!
Actually the problem is it's not finding it BECAUSE it plays with the path. Makes not sense to me.

Schyzophrenic:
Makes sense. I wrote this PR so long ago, we will need to double check the exact behavior and document it as well. 
I can potentially do it but I need to some kind of commitments this will be merged (in principle) if I put more time into it.

dbu:
@Schyzophrenic i agree on the direction we are going here and can merge this change once you double checked for side effects. thanks a lot!

dbu:
have you been able to validate if this PR has unwanted side effects?

Templating/FilterTrait.php
c33s: why remove the query parameter? why not slug the whole string?

kilab:
@Schyzophrenic any news? 

It would be nice to see your changes merged in this repository. I checked your changes in two cases (local file + remote file (with and without query parameters) and it seems have no side effects.

Templating/FilterTrait.php
Schyzophrenic: @c33s I took your advice into account and change the method to have something more generic.

Schyzophrenic:
@kilab I actually found a use case where the original code change would fail to generate the image so I updated the getFullPath function. It now seems to work OK with the use cases I tested. 
@dbu I know it has been a long time but I am comfortable merging this. Once merged, I will be able to add a comment in the main doc on how to deal with remote files which seems to be a common issue (and the reason why I looked into this in the first place).
@glouton Thanks for the support along the way ;)

glouton:
Hi there,
Thank you @Schyzophrenic for the good work and all for the support. :)
And thanks for this bundle in the first place ;o)

Imagine/Cache/Resolver/WebPathResolver.php
```diff
@@ -169,6 +169,25 @@ private function getFullPath($path, $filter)
...
+        $path = preg_replace('~[^-/.\w]+~', '', $path);
+
+        // trim
+        $path = trim($path, '-');
+
+        // remove duplicate -
+        $path = preg_replace('~-+~', '-', $path);
+
+        // lowercase
+        $path = strtolower($path);
```
dbu: cs-fixer wants mb_strtolower. can we use that or is there a specific reason to not use the mb_ version?

dbu:
i am worried by the phpunit tests failing. but it seems to be unrelated to your changes. problem is it hides if something breaks...

i just re-ran tests on the 2.x branch. can you please rebase on that branch so we see if things get green?

Imagine/Cache/Resolver/WebPathResolver.php
glouton: Hi,

Can't say for sure but from what I can tell mb_string sort of adds support for more characters such as, and I quote the doc, "A-umlaut (Ä)" (reminds me my german classes :p).
As we're talking about a path here, there shouldn't be any of these characters and, as he who can do more can do less, I don't see any reason why mb_string would raise any problem.

dbu: yeah that makes sense. it is considered general best practice to use mb_ thats why the cs fixer is complaining. but for this place it indeed seems unnecessary. is there a php comment you can put above this line to have the cs fixer ignore the line?

Schyzophrenic: I am happy to update the code with the mb version. Not sure what is the comment I need to put otherwise though.

dbu: don't know by heart either. lets use mb_strtolower. this also will not mess up if we somehow miss multibyte characters in the path (though it might mess up later in the process if the character is not supported...)

Schyzophrenic (commented):
I updated strtolower to mb_strtolower

Schyzophrenic:
I found a very edge case scenario: This [URL ](https://images2.productserve.com/?w=200&h=200&bg=white&trim=5&t=letterbox&url=ssl%3Ad8mkdcmng3.imgix.net%2Fabbe%2F753009.jpg%3Fauto%3Dformat%26bg%3D0FFF%26fit%3Dfill%26h%3D600%26q%3D100%26w%3D600%26s%3Da30c66d56411f5b7da75fe3c75cfc560&feedId=5778&k=b35628ab8b9d898c2d3a4873711daaaed5d11f8a) is converted into [this url](https://images2.productserve.com/?w=200&h=200&bg=white&trim=5&t=letterbox&url=ssl:d8mkdcmng3.imgix.net/abbe/753009.jpg?auto=format&bg=0FFF&fit=fill&h=600&q=100&w=600&s=a30c66d56411f5b7da75fe3c75cfc560&feedId=5778&k=b35628ab8b9d898c2d3a4873711daaaed5d11f8a) which breaks the remote file download. 

I tracked this change to PathHelper: https://github.com/liip/LiipImagineBundle/blob/7c469724fdf818a3611249fe619d8e0001eeb649/Imagine/Cache/Helper/PathHelper.php#L29

Which is called here: https://github.com/liip/LiipImagineBundle/blob/7c469724fdf818a3611249fe619d8e0001eeb649/Controller/ImagineController.php#L85

Removing the urldecode will fix the problem, but I am not sure why this is here in the first place. @glouton , @dbu or @c33s any idea? 

I now have a working copy on the latest branch that I haven't uploaded yet, but I wished to get your opinion first. Hopefully we will be able to close this soon.

Schyzophrenic:
@dbu I have updated the tests to reflect the changes made. Hopefully, we are good to merge now! 👍

Schyzophrenic:
Any chance this could be merged? Let me know if anything is missing. 
Thanks

glouton:
Hello @Schyzophrenic ,

I have no clue regarding the urldecode in PathHelper::urlPathToFilePath
Sorry I don't have much time to look into it but thanks for your job anyway. :1st_place_medal:

Imagine/Cache/Resolver/WebPathResolver.php
```diff
@@ -167,6 +167,22 @@ private function getFullPath($path, $filter)
...
+        $path = iconv('utf-8', 'us-ascii//TRANSLIT', $path);
+
+        // remove unwanted characters
+        $path = preg_replace('~[^-/.\w]+~', '/', $path);
+
+        // trim
+        $path = trim($path, '-');
+
+        // remove duplicate -
+        $path = preg_replace('~-+~', '-', $path);
```
dbu, line 180: i am afraid that this can break existing caches. and it seems unrelated to the change you are doing. is there any harm in having `https---...`?

Tests/Functional/Controller/ImagineControllerTest.php
```diff
@@ -309,11 +309,11 @@ public function testShouldResolveWithCustomFiltersFromCache(): void
     public function testShouldResolvePathWithSpecialCharactersAndWhiteSpaces(): void
     {
         $this->filesystem->dumpFile(
-            $this->cacheRoot.'/thumbnail_web_path/images/foo bar.jpeg',
+            $this->cacheRoot.'/thumbnail_web_path/images/foo/bar.jpeg',
```
dbu, line 312: i am afraid this could break existing caches. do we need it?

Controller/ImagineController.php
```diff
@@ -82,7 +82,6 @@ public function __construct(
      */
     public function filterAction(Request $request, $path, $filter)
     {
-        $path = PathHelper::urlPathToFilePath($path);
```
dbu, line 85: see also #1382

Imagine/Cache/Resolver/WebPathResolver.php
```diff
@@ -167,6 +167,22 @@ private function getFullPath($path, $filter)
...
+
+        if (empty($path)) {
+            $path = 'n-a';
```
dbu, line 183: if we have more than 1 image that ends up empty, this would collide, right?

could it make sense to keep the original $path and do an md5 from that if we don't manage to build a nice string?

Templating/FilterTrait.php
```diff
@@ -38,7 +38,7 @@ public function __construct(CacheManager $cache)
...
```
dbu, line 41: isn't this also used with local files? if we talk with a filesystem, the query string on the path will break things... or should it be the job of the cache to strip a potential query string if it wants to lookup on a filesystem?

dbu (commented):
i had a look at this PR again, and have a couple questions:

Schyzophrenic:
Hi @dbu , thank you for taking the time to look into this. 
I unfortunately moved on and no longer have an active project using this bundle but I know this was working fine when I tested it on my project at the time with both local and remote files. 
You nonetheless made a very good point on the existing cache, and I am pretty sure I didn't pay attention to this as I had been regenerating the cache during my tests. Taking this into consideration, would it make sense to merge it to the next non BC version?

dbu:
i wonder if we can do it in a backwards-compatible way, even.

i am going through the issues on the bundle, will come back to it.

Templating/FilterTrait.php
dbu: is the use case to allow requesting an image from a different server in a web request? or does this only make sense when coming through another path? if the later, we could make do the parse_url part in the controller instead of here...

Schyzophrenic: The use case is to be able to call the image http://myImgServer.com/image.png?v=123 from http://myAppServer.com/mypage
We need to maintain the whole path, including the parameter. Some images are served via specialized servers such as: https://images2.productserve.com/?w=200&h=200&bg=white&trim=5&t=letterbox&url=ssl%3Awww.cdiscount.com%2Fpdt2%2F9%2F2%2F5%2F1%2F700x700%2FNIN4902370537925.jpg&feedId=38753&k=421fa9c064b91b3719c999b1a900cfb30d7e23de

dbu: yep, i got that. my question was more like "where is the image url coming from?". as one code path to get here is through the controller of this bundle that is asked for an image. i would not expect that we want that controller to get a domain passed and load random images from any 3rd party host...

so maybe we could move the parse_url logic into that controller instead, and say that if you use other means we expect you to specify exactly the file you need, but if you come through the controller, we make sure to reduce things to a path and strip domain or query string if they exist...

Schyzophrenic: Well, this was exactly my use case. I had loads of image paths (literally thousands) loaded into DB via a feed load. These images would then get displayed to the users but we wished to avoid relying on 3rd parties as much as possible so we built the cache along the way.
This was also a good way to deliver the image as we deemed fit with the change operations done when building the cache.
Before we made this fix we used to download the pictures when loading the data but it created other issues. Leveraging the cache and filters is really what made more sense.

dbu:
see also #1240

Templating/FilterTrait.php
dbu: i spent some time with a collegue to try to understand what is going on.

unless we are completely mistaken, when the image is not in the local cache, it means that we provide the full image url in the path of the ImagineController::filterAction which then proceeds to download the image and cache it with the filter applied. i think that LiipImagineBundle should not support this - it would allow an attacker to provide the controller with an endless number of images from other domains. that way one could place all kinds of illegal content into the cache, or overload the cache.

did i misunderstand, or is this how the flow works?

i will look at documenting this better, but to me the change was a bug/security fix that had side effects for your use case. as it has been this way for a long time now, i think reverting it would not make sense.

dbu:
i created #1393 to discuss the security implications of loading images from anywhere

dbu:
fixed in #1397

PR #1300, opened 2020-09-21
Add FormatExtensionResolver

### LINKED ISSUE #1198: Formats and extensions
I know that this issue been already discussed many times, but it's 2019th and all issues are set to closed while bug is still there. Steps to reproduce:
```
liip_imagine:
    driver: "gd"
    filter_sets:
        cover_full:
            data_loader:           cover
            format:                jpeg
            jpeg_quality:          100
            filters:
                auto_rotate: ~
                strip: ~

                thumbnail:
                    size:          [835, 470]
                    mode:          outbound
                    allow_upscale: true
```

Given a PNG image with '.png' extension, say "123.png", imagine_filter produces a link to "media/cache/.../123.png" which is not OK because it is really a JPEG file.

```
        "liip/imagine-bundle": "dev-master",
```

ossinkine:
| Q | A
| --- | ---
| Branch? | 2.0 (was branched from master)
| Bug fix? | yes
| New feature? | yes
| BC breaks? |no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1198
| License | MIT
| Doc PR | 

I've added a new resolver to resolve issue with file extension when filter changes image format, see related issues.
Just add it as decorator to your primary resolver and images will be stored with proper extensions:
```yaml
services:
    acme.imagine.cache.resolver.format_extension:
        class: Liip\ImagineBundle\Imagine\Cache\Resolver\FormatExtensionResolver
        arguments:
            - "@liip_imagine.cache.resolver.default"
            - "@liip_imagine.filter.configuration"
        tags:
            - { name: "liip_imagine.cache.resolver", resolver: "format_extension" }
```

### DIFF
--- a/Resources/doc/cache-resolver/format_extension.rst
+++ b/Resources/doc/cache-resolver/format_extension.rst
@@ -1,0 +1,65 @@
+
+.. _cache-resolver-format-extension:
+
+FormatExtensionResolver
+=======================
+
+The ``FormatExtensionResolver`` cannot be used by itself. Instead, it is a "decorator" for
+another resolver. It adds the ability to set the correct file extension when a filter converted the image format.
+
+Configuration
+-------------
+
+To use this cache resolver, you must first define the cache resolver it will decorate.
+In this example, we will use the :ref:`Web Path Resolver <cache-resolver-web-path>`.
+
+Next, we need to define a service for this cache resolver and inject the web path cache resolver service to decorate.
+
+.. code-block:: yaml
+
+    # app/config/services.yml
+
+    services:
+        acme.imagine.cache.format_extension:
+            class: Liip\ImagineBundle\Imagine\Cache\Resolver\FormatExtensionResolver
+            arguments:
+                - "@acme.imagine.cache.resolver.web_path"
+                - "@liip_imagine.filter.configuration"
+            tags:
+                - { name: "liip_imagine.cache.resolver", resolver: "format_extension" }
+
+
+With this configuration, the format extension resolver will rewrite the extension to match the filter format.
+For example, you have the source image ``image.png`` and you apply filter with format ``jpg`` you will get jpeg-image ``image.jpg``.
+
+Usage
+-----
+
+After configuring ``FormatExtensionResolver``, you can set it as the default cache resolver
+for ``LiipImagineBundle`` using the following configuration.
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+
+    liip_imagine:
+        cache: format_extension
+
+
+Usage on a Specific Filter
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Alternatively, you can set ``FormatExtensionResolver`` as the cache resolver for a specific
+filter set using the following configuration.
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+
+    liip_imagine:
+        filter_sets:
+            cache: ~
+            my_thumb:
+                cache: format_extension
+                filters:
+                    # the filter list
--- a/Imagine/Cache/Resolver/FormatExtensionResolver.php
+++ b/Imagine/Cache/Resolver/FormatExtensionResolver.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Imagine\Cache\Resolver;
+
+use Liip\ImagineBundle\Binary\BinaryInterface;
+use Liip\ImagineBundle\Imagine\Filter\FilterConfiguration;
+
+class FormatExtensionResolver implements ResolverInterface
+{
+    /**
+     * @var ResolverInterface
+     */
+    private $resolver;
+
+    /**
+     * @var FilterConfiguration
+     */
+    private $filterConfig;
+
+    public function __construct(ResolverInterface $resolver, FilterConfiguration $filterConfig)
+    {
+        $this->resolver = $resolver;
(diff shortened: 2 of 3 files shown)


robfrawley (changes requested):
Are you able to include documentation regarding this new feature? Our documentation source files are located under the [`Resources/doc`](https://github.com/liip/LiipImagineBundle/tree/master/Resources/doc) project directory and can be locally built using the [build guide](https://github.com/liip/LiipImagineBundle/wiki/Building-RST-Documentation) from the project Wiki to confirm your documentation changes and/or additions properly compile and render as expected. If you have no familiarity with [reStructuredText (RST)](https://docutils.sourceforge.io/rst.html) (which is the markup language used for our docs) or you do not have the time or ability to update the docs for this feature, please advise and I will look into doing so myself prior to merging this PR.

ossinkine:
@robfrawley I've added the documentation for this feature but unfortunatelly I was not able to build the doc according to build manual. Feel free to fix it if something is wrong.

Perni1984:
@ossinkine many thanks for this!

Resources/doc/cache-resolver/format_extension.rst
```diff
@@ -0,0 +1,64 @@
+
+.. _cache-resolver-format-extension:
+
+FormatExtensionResolver
+=======================
+
+The ``FormatExtensionResolver`` cannot be used by itself. Instead, it is a "decorator" for
+another resolver. It add the ability to set the correct file extension when filter converts image format.
```
dbu:
```suggestion
another resolver. It adds the ability to set the correct file extension when a filter converted the image format.
```

Resources/doc/cache-resolver/format_extension.rst
```diff
@@ -0,0 +1,64 @@
...
+    services:
+        acme.imagine.cache.format_extension:
+            class: Liip\ImagineBundle\Imagine\Cache\Resolver\FormatExtensionResolver
+            arguments:
+                - "@acme.imagine.cache.resolver.web_path"
+            tags:
+                - { name: "liip_imagine.cache.resolver", resolver: "format_extension" }
+
+
+With this configuration, the cache resolver will generate paths such with file extension according to filter format.
```
dbu:
```suggestion
With this configuration, the format extension resolver will rewrite the extension to match the filter format.
```

Resources/doc/cache-resolver/format_extension.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. code-block:: yaml
+
+    # app/config/services.yml
+
+    services:
+        acme.imagine.cache.format_extension:
+            class: Liip\ImagineBundle\Imagine\Cache\Resolver\FormatExtensionResolver
+            arguments:
+                - "@acme.imagine.cache.resolver.web_path"
```
dbu, line 26: don't we need to configure the $filterConfig as well? or does that happen automatically because of the liip_imagine.cache.resolver tag?

ossinkine: You're right, I forgot it because I use autowiring in my project. Fixed.

ossinkine:
@dbu Thank you for fixes 👍

dbu (approved):
looks good to me. okay to merge, @robfrawley ?

acrobat:
@ossinkine why did you close this PR?

ossinkine:
Looks like I accidentally deleted the branch while creating another PR.

tarjei:
Hi, any news on this PR? `heic` files are becoming an issue so this is needed more an more.

plandolt:
Looking forward to this resolver. Is there anything to add/help?

Another question is about a possible conflict when change the format on two or more files with the same filename except a different extension. e.g. 1.jpg and 2.png transformed to webp format will both result in 1.webp. But those file could be totally different. Should the extension instead of being replaced better being added? Or should this behaviour be configurable?

Plus: This one does not solve the problem of no redirect on the first request in which the cached version will be generated.

ossinkine:
@scuben I have no idea why this PR still is not merged, so I don't how you can help 🤷‍♂️ 

What about file extensions, I though about that and it seemed to me that filenames like `1.jpg.webp` are ugly enough. We are using hash names to avoid conflicts.

trsteel88:
@lsmith77 are you able to merge this branch if it is fine? I'm looking to implement webp image formats (by defining format: webp) on additional filtersets and it doesn't alter the path name :(

trsteel88:
Actually just looking at this code. It doesn't look like this is just something you can turn on. You would need to implement it in your services each time. Would it be better to have this as an option? Or alternatively, always replace (or append) the new extension?

ossinkine:
@trsteel88 We can't always change the extension because it breaks BC. The such functionality in this library extends by decorating cache resolvers. Why do you think the adding an option is better then adding the decorator service?

@lsmith77 Indeed, what is keeping us from merging?

lsmith77:
the main issue is time reviewing ..

@fbourigault do you have time to look into this PR?

robfrawley:
@lsmith77 I can actually take a look at this later this week if no one else gets around to it.

trsteel88:
> @trsteel88 We can't always change the extension because it breaks BC. The such functionality in this library extends by decorating cache resolvers. Why do you think the adding an option is better then adding the decorator service?

Make it an option would mean BC would still be respected.

e.g. "redirect_format_extension" = false by default.

Then on any new installs, we could just specify true to make use of the automated redirect.

trsteel88:
Not sure if this is a big deal, but the getBrowserPath() does not render the correct extension before the image is stored.

This means that the path will render as media/cache/resolve/foo/bar.jpg if the image does not exist yet. Then it will redirect to media/cache/foo/bar.webp

After the file is generated, it will then render as media/cache/foo/bar.webp which is correct.

dbu:
thanks! and sorry for the delay.

i think for version 3 we should make this the default behaviour - it seems weird to not always do it. (also, with content negotiation for webp and stuff, did we currently only store one version, or does that take care of the extension?)

PR #1302, opened 2020-10-27
Update aws_s3.rst with clarifications re service setup

manuelkiessling:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | MIT
| Doc PR |

After struggling to set up LiipImagineBundle in an AWS EC2 environment where we do not want to use hard coded AWS API credentials, but instead rely on attaching an IAM role to our EC2 instances, I realized that the current documentation lacks some relevant information.

### DIFF
--- a/Resources/doc/cache-resolver/aws_s3.rst
+++ b/Resources/doc/cache-resolver/aws_s3.rst
@@ -92,13 +92,12 @@ While creating the resolver through a factory using the configuration above does
 it can sometimes be neccessary to create the resolver as a service.
 
 .. tip::
-    For example, this is required if you do not want
-    to connect to S3 using hardcoded AWS credentials, but instead your code will run on an AWS EC2 instance with
-    an assumed IAM role for AWS API authentication. In this case, simply setting up the resolver through a
-    factory configuration block and leaving out the `credentials` block won't cut it. Instead, you need to
-    set up the resolver as a service as shown below, passing an S3Client - this way, the resolver will talk
-    to the AWS API via the S3Client, and the S3Client can be configured without explicit credentials, in which
-    case it will resolve to using the assumed role attached to the EC2 instance.
+    For example, this is required if you do not want to hard code AWS credentials for your resolver, but
+    instead make it use the `Default Credential Provider Chain`_ within AWS (for example, because you are
+    running this on an EC2 instance which is already attached to an IAM role with the required permissions).
+    In this case, simply setting up the resolver through a factory configuration block and leaving out
+    the credentials block won't cut it. Instead, you need to set up the resolver as a service as shown
+    below, passing an S3Client.
 
 You have to set up the services required:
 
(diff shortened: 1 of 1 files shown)


Resources/doc/cache-resolver/aws_s3.rst
```diff
@@ -88,6 +88,18 @@ Create Resolver from a Factory
...
+it can sometimes be neccessary to create the resolver as a service.
+
+.. tip::
+    For example, this is required if you do not want
+    to connect to S3 using hardcoded AWS credentials, but instead your code will run on an AWS EC2 instance with
+    an assumed IAM role for AWS API authentication. In this case, simply setting up the resolver through a
+    factory configuration block and leaving out the `credentials` block won't cut it. Instead, you need to
+    set up the resolver as a service as shown below, passing an S3Client - this way, the resolver will talk
+    to the AWS API via the S3Client, and the S3Client can be configured without explicit credentials, in which
+    case it will resolve to using the assumed role attached to the EC2 instance.
```
dbu, line 101: i tried shortening this. what do you think?

When you create the resolver as a service, this allows you to pass an S3Client. When running in an AWS EC2 instance, the S3Client can be configured without explicit credentials, making it resolve the credentials using the assumed role attached to the EC2 instance. That way, you can set up an S3 resolver without having to configure the credentials (which you can't know because they depend on the instance).

dbu (approved):
thanks! yeah explaining the "why" is a good practice in documentation.

i am not sure if EC2 provides the credentials as environment variables, so that newer symfony versions would actually allow to configure the credentials that way, but when the S3Client handles this automatically, your solution sounds to me like the right way to handle this.

Resources/doc/cache-resolver/aws_s3.rst
manuelkiessling: @dbu Love it, much better. One change proposal for your change proposal, though - leave the last sentence out!

> i am not sure if EC2 provides the credentials as environment variables, so that newer symfony versions would actually allow to configure the credentials that way, but when the S3Client handles this automatically, your solution sounds to me like the right way to handle this.

Not automatically. However, using an S3Client as the middle-man makes the setup automatically use the "Default Credential Provider Chain" of AWS, as explained in https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html

Maybe this is the way to put it:

>  For example, this is required if you do not want to hard code AWS credentials for your resolver, but instead make it use the Default Credential Provider Chain within AWS (for example, because you are running this on an EC2 instance which is attached to an IAM role with the required permissions). In this case, simply setting up the resolver through a factory configuration block and leaving out the `credentials` block won't cut it. Instead, you need to set up the resolver as a service as shown below, passing an S3Client.

dbu: sounds good :+1: 

you could link "Default Credential Provider Chain" in the text to that link you gave, for extra context.

manuelkiessling: @dbu Done.

PR #1304, opened 2020-10-29
fix naming of imagine in composer dependency as uppercase is no longer ok

dbu:


dbu:
hm, thats version 1 branch, i guess we should not bother with it anymore...

PR #1305, opened 2020-10-29
allow to use Doctrine/Persistence:^2

jkabat:
| Q | A
| --- | ---
| Branch? |  2.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| License | MIT

### DIFF
--- a/Tests/Binary/Loader/AbstractDoctrineLoaderTest.php
+++ b/Tests/Binary/Loader/AbstractDoctrineLoaderTest.php
@@ -11,8 +11,10 @@
 
 namespace Liip\ImagineBundle\Tests\Binary\Loader;
 
-use Doctrine\Common\Persistence\ObjectManager;
-use Doctrine\Common\Persistence\ObjectRepository;
+use Doctrine\Common\Persistence\ObjectManager as LegacyObjectManager;
+use Doctrine\Common\Persistence\ObjectRepository as LegacyObjectRepository;
+use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\ObjectRepository;
 use Liip\ImagineBundle\Binary\Loader\AbstractDoctrineLoader;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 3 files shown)


Tests/Binary/Loader/AbstractDoctrineLoaderTest.php
```diff
@@ -34,12 +36,16 @@ class AbstractDoctrineLoaderTest extends TestCase
 
     public function setUp(): void
     {
-        if (!interface_exists(ObjectManager::class)) {
+        if (interface_exists(LegacyObjectManager::class)) {
+            $omClassName = LegacyObjectManager::class;
+        } elseif (interface_exists(ObjectManager::class)) {
+            $omClassName = ObjectManager::class;
+        } else {
             $this->markTestSkipped('Requires the doctrine/orm package.');
```
fbourigault: As `doctrine/orm` is in `require-dev` it will always be installed when running tests.Therefore you can remove the `else` statement.

composer.json
```diff
@@ -35,6 +35,7 @@
         "aws/aws-sdk-php": "^2.4",
         "doctrine/cache": "^1.1",
         "doctrine/orm": "^2.3",
```
fbourigault: I think this `require-dev` entry is useless. There isn't any code in this bundle which depends on `Doctrine\\ORM`!

We should remove it and keep `doctrine/persistence` as the true optional dependency thus we only provide an `AbstractDoctrineLoader` to help developers to implement their own doctrine ORM/ODM loader.

lsmith77:
I just merged https://github.com/liip/LiipImagineBundle/pull/1337

fbourigault:
Could you rebase? #1337 replaced the `doctrine/orm` dev dependency with `doctrine/persistence` `^1.3`.

jkabat:
@fbourigault Ill do that today/tomorrow.

jkabat:
rebased and fixed based on suggestions

Tests/Binary/Loader/AbstractDoctrineLoaderTest.php
jkabat: done.

lsmith77:
thank you!

Issue #1309, opened 2020-12-04
Add support for php 8


garak:
I guess no description is needed...

dbu:
to get this, we need a pull request to change composer.json and add php 8 tests in .travis.yml. afaik they 8.0 is not available there yet, so instead the php version would be `nightly` for now.

garak:
It's available as `8.0snapshot`

PR #1313, opened 2020-12-12
allow php 8

garak:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1309
| License | MIT
| Doc PR | not needed

Allow PHP 8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
     - PHPUNIT_FLAGS="-v"
     - PHPUNIT_ENABLED="true"
     - COVERALLS_ENABLED="false"
+    - IMAGICK_ENABLED="true"
+    - ENQUEUE_ENABLED="true"
     - STABILITY=stable
 
 matrix:
@@ -18,11 +20,15 @@ matrix:
   include:
     - php: '7.1'
     - php: '7.2'
+    - php: '7.3'
+    - php: '7.4'
 
     # Enable code coverage with the latest supported PHP version
-    - php: '7.3'
+    - php: '8.0'
       env:
         - COVERALLS_ENABLED="true"
+        - IMAGICK_ENABLED="false"
+        - ENQUEUE_ENABLED="false"
         - XDEBUG_MODE=coverage
         - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
 
@@ -30,31 +36,31 @@ matrix:
     - php: '7.1'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
-    - php: '7.3'
+    - php: '8.0'
       env:
-        - SYMFONY_PHPUNIT_VERSION="7.5"
+        - SYMFONY_PHPUNIT_VERSION="8.5"
         - COMPOSER_FLAGS="--prefer-lowest"
+        - IMAGICK_ENABLED="false"
+        - ENQUEUE_ENABLED="false"
 
     # Test each supported Symfony version with lowest supported PHP version
-    - php: '7.1'
+    - php: '7.1.3'
       env:
         - SYMFONY_VERSION=3.4.*
-    - php: '7.1'
+    - php: '7.1.3'
       env:
-        - SYMFONY_VERSION=4.3.*
-
-    - php: '7.2.9'
-      env:
-        - SYMFONY_VERSION=5.0.*
-
-    - php: '7.2.9'
+        - SYMFONY_VERSION=4.4.*
+    - php: '7.2.5'
       env:
         - SYMFONY_VERSION=5.1.*
+    - php: '7.2.5'
+      env:
+        - SYMFONY_VERSION=5.2.*
 
     # Remember to test upcoming Symfony versions below with lowest supported PHP version and dev dependencies
 
     # Test upcoming PHP versions with dev dependencies
-    - php: '7.4snapshot'
+    - php: 'nightly'
       env:
         - STABILITY=dev
         - COMPOSER_FLAGS="--ignore-platform-reqs --prefer-stable"
@@ -79,10 +85,15 @@ before_install:
       phpenv config-rm xdebug.ini || true;
     fi
   - if [[ "$COVERALLS_ENABLED" == "true" ]]; then
-      travis_retry composer require --dev satooshi/php-coveralls:^2.0 --no-update $COMPOSER_FLAGS;
+      travis_retry composer require --dev php-coveralls/php-coveralls:^2.0 --no-update $COMPOSER_FLAGS;
     fi
-  - pecl channel-update pecl.php.net
-  - yes | pecl install imagick
+  - if [[ "$IMAGICK_ENABLED" == "true" ]]; then
+      pecl channel-update pecl.php.net;
+      yes | pecl install imagick;
+    fi
+  - if [[ "$ENQUEUE_ENABLED" == "false" ]]; then
+      composer remove --no-update --dev enqueue/enqueue-bundle;
+    fi
 
 install:
   - travis_retry composer update --prefer-dist --no-interaction --no-suggest --no-progress --ansi $COMPOSER_FLAGS
--- a/Tests/Component/Console/Style/ImagineStyleTest.php
+++ b/Tests/Component/Console/Style/ImagineStyleTest.php
@@ -225,8 +225,13 @@ class ImagineStyleTest extends AbstractTest
      */
     public function testInvalidFormatAndReplacements(string $format, array $replacements): void
     {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ matrix:
     - php: '7.2'
```
fbourigault, line 21: As far as I remember, this part of the matrix should contains all the supported PHP versions minus the latest which has code coverage enabled.

.travis.yml
```diff
@@ -21,7 +21,7 @@ matrix:
...
 
     # Enable code coverage with the latest supported PHP version
-    - php: '7.3'
+    - php: '7.4'
```
fbourigault, line 24: This should be 8.0. This would match the comment.

.travis.yml
```diff
@@ -40,20 +40,23 @@ matrix:
         - SYMFONY_VERSION=3.4.*
     - php: '7.1'
       env:
-        - SYMFONY_VERSION=4.3.*
+        - SYMFONY_VERSION=4.4.*
 
     - php: '7.2.9'
       env:
         - SYMFONY_VERSION=5.0.*
```
fbourigault, line 47: Should this be kept as Symfony 5.0 isn't supported anymore?

.travis.yml
```diff
@@ -40,20 +40,23 @@ matrix:
...
 
-    - php: '7.2.9'
+    - php: '7.4'
```
fbourigault, line 49: This should not change as minimum supported version of Symfony 5.1 is 7.2.

.travis.yml
```diff
@@ -40,20 +40,23 @@ matrix:
...
       env:
         - SYMFONY_VERSION=5.1.*
 
+    - php: '8.0'
```
fbourigault, line 53: This should be 7.2.9 as Symfony 5.2 requires at least this version.

fbourigault (changes requested):
Thank you for handling this!

To keep the matrix pilosophy I introducted in https://github.com/liip/LiipImagineBundle/pull/1202, I requested some changes.

The major ideas in the matrix is to:
- test each supported php version with latest dependencies.
- test latest supported php version with code coverage enabled.
- test lowest dependencies with oldest and latest supported php versions.
- test each supported Symfony version with it's minimum required PHP.

.travis.yml
```diff
@@ -30,30 +32,28 @@ matrix:
...
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
-    - php: '7.3'
+    - php: '8.0'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
 
     # Test each supported Symfony version with lowest supported PHP version
-    - php: '7.1'
+    - php: '5.5.9'
```
fbourigault, line 40: This one must be kept on 7.1 as it's the minimal PHP version required by this bundle!

fbourigault:
The matrix looks good now, but the build is failing.

7.X failures are unrelated with this PR and I started to fix those in https://github.com/liip/LiipImagineBundle/pull/1315.

dbu:
#1315 fixed the existing builds, can you please rebase this PR on master?

garak:
No luck for php8, imagemagick does not compile :disappointed:

garak:
Related: https://github.com/Imagick/imagick/issues/358

fbourigault:
You can add an `IMAGINE_ENABLED` environment variable with de default value of `true`, install `imagick` extension only if this variable value is `true` and set the variable to `false` for PHP 8.0 jobs.

A few tests will be skipped as imagick won't be available, but that's acceptable as users won't be able to have such setup (php 8 + imagick).

garak:
Still not working since a dependency can't install. Related issue: https://github.com/php-enqueue/enqueue-dev/issues/1108

dbu:
as enqueue is optional, could we have the php 8 build remove it from composer before installing and have the tests that rely on enqueue skip when enqueue is not available?

garak:
OK, I think we're almost there...
It looks like phpunit 7.5 doesn't work well with php8, so I think we should give up with the possibility of testing php8 with it.
I would remove `SYMFONY_PHPUNIT_VERSION="7.5"` from line 41 of travis config, what do you think?

dbu:
yeah, phpunit 7 does not work on php 8. afaik phpunit 8 now works on both php 7 and 8, lets try that? phpunit 9 brings a couple of BC breaks that means we would need to adjust assertions and such. we eventually want to do that, but should not mix in too many things into one PR

can you try explicitly setting it to the current 8.x version?

garak:
Unfortunately, even phpunit 8.5 is failing (due to missing return type on `tearDown`)

dbu:
this has been fixed in symfony 3.4.23
i suggest changing the composer.json to `"symfony/framework-bundle": "^3.4.23|^4.3|^5.0",`

.travis.yml
```diff
@@ -11,50 +11,56 @@ env:
...
 
     # Minimum supported dependencies with the latest and oldest supported PHP versions
     - php: '7.1'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
-    - php: '7.3'
+    - php: '8.0'
       env:
-        - SYMFONY_PHPUNIT_VERSION="7.5"
+        - SYMFONY_PHPUNIT_VERSION="8.5"
```
fbourigault, line 41: You can remove this line and let PHPUnit Bridge choose the right version (see https://symfony.com/doc/current/components/phpunit_bridge.html#modified-phpunit-script)

.travis.yml
```diff
@@ -79,10 +85,15 @@ before_install:
...
   - if [[ "$COVERALLS_ENABLED" == "true" ]]; then
-      travis_retry composer require --dev satooshi/php-coveralls:^2.0 --no-update $COMPOSER_FLAGS;
+      travis_retry composer require --dev php-coveralls/php-coveralls:^2.0 --no-update $COMPOSER_FLAGS;
+    fi
+  - if [[ "$IMAGICK_ENABLED" == "true" ]]; then
+      pecl channel-update pecl.php.net;
+      yes | pecl install imagick;
+    fi
+  - if [[ "$ENQUEUE_ENABLED" == "false" ]]; then
+      composer remove --dev enqueue/enqueue-bundle;
```
fbourigault, line 95: You should add the `--no-update` option as the `update` is performed at `install` step.

.travis.yml
garak: Isn't the ratio to try testing with older versions?

garak:
Any idea of the reasons of last failures? :disappointed: Why images are not found?

fbourigault:
Could you rebase please? I will look at those failure localy.

.travis.yml
dbu: i suggested fixing to phpunit 8 for now, because likely there are some assertions that will stop working with phpunit 9. that way we can keep upgrading to phpunit 9 separate, to not add more into one PR than necessary.

.travis.yml
```diff
@@ -11,50 +11,56 @@ env:
...
+        - ENQUEUE_ENABLED="false"
         - XDEBUG_MODE=coverage
         - PHPUNIT_FLAGS="-v --coverage-text --coverage-clover var/build/clover.xml"
 
     # Minimum supported dependencies with the latest and oldest supported PHP versions
     - php: '7.1'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
-    - php: '7.3'
+    - php: '8.0'
```
fbourigault, line 39: This is the last failing job. I think we should stick with php 7.4 here as having php 8.0 with --prefer-lowest would work only if any dependencies doesn't declare supporting php with the `>=` operator without being compatible with php 8.0.

Note: Symfony does the same (see https://github.com/symfony/symfony/blob/5.x/.travis.yml#L23-L30)

dbu: yeah, i agree. we could also drop the "latest php version prefer-lowest" completely. the main thing we care about is "lowest php version with prefer-lowest".

fbourigault: The highest supported PHP version with lowest dependencies tests the case where, in a project, an unrelated dependency lock one of our dependency to a lower version than expected.

I took the idea from Symfony IIRC but maybe this is already covered with other jobs. I’m neutral about keep/remove this job from the matrix.

dbu: there is always a chance of some combinations of versions having a problem. but builds are fast and parallel, so lets change this to 7.4 for now.

lsmith77:
closing in favor of https://github.com/liip/LiipImagineBundle/pull/1325

PR #1329, opened 2020-12-30
Enable WebP support in tests

fbourigault:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | no
| Fixed tickets |
| License | MIT
| Doc PR |

This is a tentative to run tests with WebP support enabled.

PR #1331, opened 2020-12-31
Create "foo bar.jpeg.webp"

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

Create `foo bar.jpeg.webp` for fix `ImagineControllerTest::testShouldResolvePathWithSpecialCharactersAndWhiteSpaces()`.
See https://github.com/liip/LiipImagineBundle/pull/1329#issuecomment-752631427

### DIFF
--- a/Tests/Functional/Controller/ImagineControllerTest.php
+++ b/Tests/Functional/Controller/ImagineControllerTest.php
@@ -280,6 +280,11 @@ class ImagineControllerTest extends AbstractSetupWebTestCase
             $this->cacheRoot.'/thumbnail_web_path/images/foo bar.jpeg',
             'anImageContent'
         );
+        $this->filesystem->dumpFile(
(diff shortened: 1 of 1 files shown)


fbourigault:
The `ImagineControllerTest::testShouldResolvePathWithSpecialCharactersAndWhiteSpaces()` test is not expected to re-generate the image from the cache.

I didn't find any explicit reference, but I think the goal of this test is to ensure that when special chars/white spaces are used in the path, the cache is correctly resolved thus the cache is not re-build from source at every request.

peter-gribanov:
@fbourigault Yes. You were right.

fbourigault (approved):

Issue #1319, opened 2020-12-17
WebP support is broken


fbourigault:
Given the following code:
```
<img src="{{ asset('DSC_3644.JPG') | imagine_filter('thumbnail') }}" alt="DSC_3644.JPG" />
```

When there is no cache, the src value is `https://localhost:8000/media/cache/resolve/thumbnail/DSC_3644.JPG` which is handled by the `ImagineController` and redirect to `https://localhost:8000/media/cache/thumbnail/DSC_3644.JPG.webp`

When the cache is warmed, the src value is `https://127.0.0.1:8000/media/cache/thumbnail/DSC_3644.JPG` which is not handled by the `ImagineController` and returns the JPEG version of the image.

I think we should not support WebP this way and consider using proper content negotiation which will require users to use the `<picture>` HTML element but will allow deterministic behavior and also support other formats such as AVIF (see https://github.com/liip/LiipImagineBundle/issues/1314).

I didn't tested it, but this WebP implementation may break HTTP Caching.

I propose to revert https://github.com/liip/LiipImagineBundle/pull/1307 and then to follow this plan:

- Move to GitHub Actions as travis-ci.com does not support webp (see #1318).

- Ensure that WebP can be used as a supported format by adding tests and strong configuration validation to ensure that when webp format is requested by the end user, the used driver also support it. This would avoid issues about webp format not working.

At this point, it's possible to do content negotiation through the use of multiple filters_set (one per format).

- Allow to use configure multiple formats per filter set to avoid filters_set duplications because of the format and add a `format` option to the `imagine_filter` twig filter.

At this point, it's possible to do content negotiation with `<picture>` and a single filters_set (configured with multiple formats).

- Allow server-side content-negotiation. IMHO, this should be done at the twig level. to avoid the issue I described above. A service could also be provided to ease integration with other systems (APIs ?).

At this point, it will be possible to do both client-side and server-side content negotiation with no compromise and also being open to add AVIF and any future format.

martinbroos:
I'm still very exited to use webp with LiipImagine, but haven't tried the new implementation yet. But your cache issue makes sense.. I'm still not against using html picture element as content negotiation.

lsmith77:
sorry. I didn’t see the issue before releasing 2.4.0. 

I will look at how to proceed today. 

@peter-gribanov can you maybe also have a look at the issue?

lsmith77:
@fbourigault proper content negotiation is indeed the way to go.

[...]

PR #1333, opened 2020-12-31
Not resolve WebP in CacheManager

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

Fix problem described in #1319

> Given the following code:
> 
> ```twig
> 
> ```
> 
> When there is no cache, the src value is https://localhost:8000/media/cache/resolve/thumbnail/DSC_3644.JPG which is handled by the ImagineController and redirect to https://localhost:8000/media/cache/thumbnail/DSC_3644.JPG.webp
> 
> When the cache is warmed, the src value is https://127.0.0.1:8000/media/cache/thumbnail/DSC_3644.JPG which is not handled by the ImagineController and returns the JPEG version of the image.

### DIFF
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -55,22 +55,30 @@ class CacheManager
     protected $defaultResolver;
 
     /**
+     * @var bool
+     */
+    private $webpGenerate;
+
+    /**
      * Constructs the cache manager to handle Resolvers based on the provided FilterConfiguration.
      *
      * @param string $defaultResolver
+     * @param bool   $webpGenerate
      */
     public function __construct(
         FilterConfiguration $filterConfig,
         RouterInterface $router,
         SignerInterface $signer,
         EventDispatcherInterface $dispatcher,
-        $defaultResolver = null
+        $defaultResolver = null,
+        $webpGenerate = false
     ) {
         $this->filterConfig = $filterConfig;
         $this->router = $router;
         $this->signer = $signer;
         $this->dispatcher = $dispatcher;
         $this->defaultResolver = $defaultResolver ?: 'default';
+        $this->webpGenerate = $webpGenerate;
     }
 
     /**
@@ -103,12 +111,12 @@ class CacheManager
         if (!empty($runtimeConfig)) {
             $rcPath = $this->getRuntimePath($path, $runtimeConfig);
 
-            return $this->isStored($rcPath, $filter, $resolver) ?
+            return !$this->webpGenerate && $this->isStored($rcPath, $filter, $resolver) ?
                 $this->resolve($rcPath, $filter, $resolver) :
                 $this->generateUrl($path, $filter, $runtimeConfig, $resolver, $referenceType);
         }
 
-        return $this->isStored($path, $filter, $resolver) ?
+        return !$this->webpGenerate && $this->isStored($path, $filter, $resolver) ?
             $this->resolve($path, $filter, $resolver) :
             $this->generateUrl($path, $filter, [], $resolver, $referenceType);
     }
--- a/Resources/config/imagine.xml
+++ b/Resources/config/imagine.xml
@@ -126,6 +126,7 @@
             <argument type="service" id="liip_imagine.cache.signer" />
             <argument type="service" id="event_dispatcher" />
             <argument>%liip_imagine.cache.resolver.default%</argument>
+            <argument>%liip_imagine.webp.generate%</argument>
         </service>
 
         <service id="liip_imagine.filter.configuration" class="Liip\ImagineBundle\Imagine\Filter\FilterConfiguration">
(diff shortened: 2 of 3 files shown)


Imagine/Cache/CacheManager.php
```diff
@@ -109,12 +117,12 @@ public function getBrowserPath($path, $filter, array $runtimeConfig = [], $resol
...
             $rcPath = $this->getRuntimePath($path, $runtimeConfig);
 
-            return $this->isStored($rcPath, $filter, $resolver) ?
+            return !$this->webpGenerate && $this->isStored($rcPath, $filter, $resolver) ?
                 $this->resolve($rcPath, $filter, $resolver) :
                 $this->generateUrl($path, $filter, $runtimeConfig, $resolver, $referenceType);
         }
 
-        return $this->isStored($path, $filter, $resolver) ?
+        return !$this->webpGenerate && $this->isStored($path, $filter, $resolver) ?
```
fbourigault, line 119: This can’t be the solution as when WebP is enabled, all URLs will go through the controller. This work in your project but doesn’t with this bundle and won’t work with s3, etc...

fbourigault (changes requested):

Imagine/Cache/CacheManager.php
peter-gribanov: Why won't it work? The request goes through the controller and is cached by the client.  A repeated call by the same client will not reach the controller and will be immediately redirected by the browser to s3, etc.

fbourigault: If you serve the image to 1M clients, you will get 1M requests to the Controller.

Previously, only the first request goes through the Controller. The 999,999 others are directly served from s3.

peter-gribanov: Yes, but we receive 1M requests to s3 from the server to check for the existence of the file. This is also not a free operation. This affects the page rendering speed. We can move the LiipImagine controller to a separate endpoint and better manage the load.

peter-gribanov: We can improve performance by caching the full page as static, but then we definitely cannot resolve WebP in templates.
We can introduce one more option in the configurations that will allow us to resolve WebP in templates, but there can be much more unexpected problems with this.

fbourigault: > Yes, but we receive 1M requests to s3 from the server to check for the existence of the file. This is also not a free operation.

This is not free, but it's what happen when WebP support is disabled. Having the same behavior is a must have IMHO.

> We can improve performance by caching the full page as static, but then we definitely cannot resolve WebP in templates.

Page caching is out of scope of this bundle.

> We can introduce one more option in the configurations that will allow us to resolve WebP in templates, but there can be much more unexpected problems with this.

I don't think you need more configuration options. As you do content negotiation in the controller, you should do the same in the twig extension.

There is probably some logic to extract from the `FilterService` (you can't use it directly because it would generate the cached file when the twig is rendered).

peter-gribanov: > Page caching is out of scope of this bundle.

That is why i say that we cannot resolve WebP in templates, since we do not know how these templates will be used. And we cannot do this:

> This is not free, but it's what happen when WebP support is disabled. Having the same behavior is a must have IMHO.

fbourigault (approved):
I talked about this with @lsmith77 on Slack and we both agree to have this limitations when WebP support is enabled.

lsmith77:
could you please also add a note in the documentation?

Imagine/Cache/CacheManager.php
trsteel88: Is there was a reverse proxy in place caching HTML, it should be configured to vary on the ACCEPT.

I don't agree that the controller should be doing all the work for every request. Once a static file is generated, it should resolve to it in the templates.

You give an example of s3, but not everybody uses that.

peter-gribanov: There are many ways to avoid calling a controller, but they all require individual server configuration. Resolving in the controller is an easy way to make WebP support out of the box for everyone, including shared hosting.

This is a simple example that saves you the trouble of accessing the controller:

```
map $http_accept $webp_suffix {
    "~*webp"  ".webp";
}

server {

   # ...

    location ~ \.(png|jpe?g)(\?[a-f0-9]+)?$ {
        expires 365d;
        access_log off;
        try_files $uri$webp_suffix $uri /index.php$is_args$args;
    }

   # ...
}
```

> *I hope i take the time to describe in detail how users can improve the rendering speed of pages and images when using [Amazon S3](https://symfony.com/doc/2.x/bundles/LiipImagineBundle/cache-resolver/amazons3.html) or [file system](https://symfony.com/doc/2.x/bundles/LiipImagineBundle/cache-resolver/web_path.html) as cache.*

PR #1336, opened 2021-01-04
Run static code analysis on GitHub Actions

fbourigault:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | MIT
| Doc PR |
 
As discussed with @dbu, here is a PR to replace most of Scrutinizer work with static code analysis.

I choose PHPStan as I use it daily at work and it's the tool used by SonataAdminBundle which is my main source of inspiration for all my GitHub Actions PRs.

PR #1338, opened 2021-01-05
Fix unsafe usages of new static().

fbourigault:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | ?
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | MIT
| Doc PR |

PHPStan (#1336) complains about those two `new static` calls.

I'd like to have your opinion to know if it is a BC break or not.

### DIFF
--- a/Async/CacheResolved.php
+++ b/Async/CacheResolved.php
@@ -77,6 +77,6 @@ class CacheResolved implements \JsonSerializable
             throw new LogicException('The message uris must not be empty array.');
         }
 
-        return new static($data['path'], $data['uris']);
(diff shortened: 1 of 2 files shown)


lsmith77:
Hmm ... this is a BC break but its super unlikely to affect anyone.
But I would hold off until 3.0, especially if we are just doing this to please phpstan.

fbourigault:
I'm fine with this. I will exclude those two little mistakes in PHPStan.

I will deprecate those calls in an other PR.

PR #1347, opened 2021-01-13
Generate WebP in liip:imagine:cache:resolve CLI command and async resolve cache messages

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1319
| License | MIT
| Doc PR | 

The `liip:imagine:cache:resolve` CLI command warms up the cache for the specified image sources with all or specified filters applied, and prints the list of cache files. I forgot that this command should also create a image in WebP format if [WebP generation](https://symfony.com/doc/2.x/bundles/LiipImagineBundle/basic-usage.html#webp-image-format) is enabled.

* Add new mthod `FilterService::warmsUpCache()`;
* The mthod `FilterService::bustCache()` now return result of busting;
* The `liip:imagine:cache:resolve` command now warms up image in WebP format too;
* The `liip:imagine:cache:remove` command now remove image in WebP format too;
* The `ResolveCacheProcessor` wat async warms up cache now warmed up image in WebP format too;
* Add unit tests for `FilterPathContainer`;
* Add unit tests for `FilterService`.

### DIFF
--- a/Command/ResolveCacheCommand.php
+++ b/Command/ResolveCacheCommand.php
@@ -14,6 +14,7 @@ namespace Liip\ImagineBundle\Command;
 use Liip\ImagineBundle\Imagine\Cache\CacheManager;
 use Liip\ImagineBundle\Imagine\Data\DataManager;
 use Liip\ImagineBundle\Imagine\Filter\FilterManager;
+use Liip\ImagineBundle\Service\FilterPathContainer;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -23,6 +24,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 class ResolveCacheCommand extends Command
 {
     use CacheCommandTrait;
+
     protected static $defaultName = 'liip:imagine:cache:resolve';
 
     /**
@@ -30,13 +32,30 @@ class ResolveCacheCommand extends Command
      */
     private $dataManager;
 
-    public function __construct(DataManager $dataManager, CacheManager $cacheManager, FilterManager $filterManager)
-    {
+    /**
+     * @var bool
+     */
+    private $webpGenerate;
+
+    /**
+     * @var mixed[]
+     */
+    private $webpOptions;
+
+    public function __construct(
+        DataManager $dataManager,
+        CacheManager $cacheManager,
+        FilterManager $filterManager,
+        bool $webpGenerate,
+        array $webpOptions
+    ) {
         parent::__construct();
 
         $this->dataManager = $dataManager;
         $this->cacheManager = $cacheManager;
         $this->filterManager = $filterManager;
+        $this->webpGenerate = $webpGenerate;
+        $this->webpOptions = $webpOptions;
     }
 
     protected function configure(): void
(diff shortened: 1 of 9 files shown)


Command/ResolveCacheCommand.php
```diff
@@ -110,11 +129,23 @@ private function runCacheImageResolve(string $image, string $filter, bool $force
...
+                    $filteredBinary = $this->filterManager->applyFilter($binary, $filter, $filterPathContainer->getOptions());
+                    $this->cacheManager->store($filteredBinary, $filterPathContainer->getTarget(), $filter);
+
+                    $this->io->status('resolved', 'green');
+                } else {
+                    $this->io->status('cached', 'white');
+                }
             }
 
             $this->io->line(sprintf(' %s', $this->cacheManager->resolve($image, $filter)));
```
mbabker: Maybe this should also be moved into the `$filterPathContainers` foreach?  Especially as it's the cache manager's `resolve()` method that ultimately generates the image's URL, and when you're using cache, saves the URL there.  Otherwise, only the non-webp variation is going to have its URL resolved here.

```php
$this->io->line(sprintf(' %s', $this->cacheManager->resolve($filterPathContainer->getTarget(), $filter)));
```

peter-gribanov: Interesting idea :+1:

peter-gribanov:
Build failed because for PHP 8 used PHPUnit 9 what is not compatible with current tests. See fix #1381

Command/ResolveCacheCommand.php
peter-gribanov: Of course, it would be nice to see URLs to WebP pictures in the console, but this requires opening the internal implementation of `FilterService::buildFilterPathContainers()` to the outside. I wouldn't want that.

peter-gribanov:
@lsmith77 it would be nice to merge this PR. It fixes some bugs not implemented in #1307, adds unit tests and adds a cache warming function. If this PR is too large, then i can split it into several PRs.

Service/FilterService.php
```diff
@@ -70,21 +70,44 @@ public function __construct(
...
+
+        return $busted;
+    }
+
+    /**
+     * @param bool $forced Force warm up cache
+     *
+     * @return bool Returns true if the cache is warmed up
+     */
+    public function warmsUpCache(
```
dbu: why warmsUpCache and not warmUpCache? it feels like a request to do the warmup to me.

Service/FilterService.php
```diff
@@ -70,21 +70,44 @@ public function __construct(
     /**
      * @param string $path
      * @param string $filter
+     *
+     * @return bool Returns true if the cache is busted
```
dbu:
```suggestion
     * @return bool Returns true if we removed at least one cached image
```

dbu (commented):
sorry for the late reply. this looks good to me, and a nice refactoring to move the warmup into the filter service.

i have some nitpicks, can you please have a look?

dbu:
if https://github.com/liip/LiipImagineBundle/pull/1360/ is merged first, we need to adjust the logic in the symfony messenger processor to follow the changes done here.

peter-gribanov:
@dbu renamed the method to `warmUpCache`.

peter-gribanov:
The order of the merger is not critical. I can make changes to my PR after #1360  merge.

Tests/Service/FilterServiceTest.php
```diff
@@ -0,0 +1,930 @@
...
+use Liip\ImagineBundle\Imagine\Filter\FilterManager;
+use Liip\ImagineBundle\Service\FilterService;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @covers \Liip\ImagineBundle\Service\FilterService
+ */
+final class FilterServiceTest extends TestCase
```
dbu, line 27: there is a LOT of mocking going on in this test. i wonder if we could not move towards a more functional test that uses the real managers instead of mocking. if tests fail, it will be a bit harder to know where the mistake was - but if they succeed, they create more trust that things actually work as expected.

Tests/Service/FilterPathContainerTest.php
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Liip\ImagineBundle\Tests\Service;
+
+use Liip\ImagineBundle\Service\FilterPathContainer;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \Liip\ImagineBundle\Service\FilterPathContainer
+ */
+final class FilterPathContainerTest extends TestCase
```
dbu, line 20: thanks for adding a test for this class :+1:

dbu (commented):
thanks a lot! i really like that we move from talking about "resolving" to "warmup" :+1: 

#1360 won the race ;-) can you please rebase this branch on 2.x and apply your refactoring to WarmupCacheHandler?

i am worried a bit by how much mocking we are doing in one of the unit tests. but if you are happy with it, we can leave it like that, just want to ask the question.

peter-gribanov:
> #1360 won the race ;-) can you please rebase this branch on 2.x and apply your refactoring to WarmupCacheHandler?

done

Tests/Service/FilterServiceTest.php
peter-gribanov: I agree that it would be great to supplement the unit tests with functional tests, but unfortunately i don't have time to do this :confused:

dbu:
thanks a lot!

i will look into creating a new minor release next week.

Issue #1346, opened 2021-01-09
review WebP documentation


lsmith77:
Followup to https://github.com/liip/LiipImagineBundle/pull/1333 but also in general given all the discussions we had, I think it makes sense to review the documentation to be sure it is accurate.

specifically https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/basic-usage.rst and https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/configuration.rst

/cc @peter-gribanov / @fbourigault

PR #1348, opened 2021-01-13
Add Twig filter for resolve path to cache

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1346
| License | MIT
| Doc PR | 

Add a `imagine_filter_cache` new Twig filter what always return URL to cached image, even if there is no cache. Proper use of this filter can improve performance.

If you always [warm up the cache](https://symfony.com/doc/current/bundles/LiipImagineBundle/resolve-cache-images-in-background.html), then you do not need to check the existence of the cached image every time when rendering templates. You can immediately use the link to the cached version of the image.

Also, you can avoid unnecessary controller calls if you use the local filesystem to store the image cache and use the same cache path as in the controller.

```
# /etc/nginx/nginx.conf
map $http_accept $webp_suffix {
    "~*webp"  ".webp";
}

server {
    # ...

    location ~ ^/media/cache {
        expires 365d;
        access_log off;
        try_files $uri$webp_suffix $uri /index.php$is_args$args;
    }

    # ...
}
```

```yaml
# config/packages/liip_imagine.yaml
liip_imagine:
    resolvers:
       profile_photos:
          web_path:
            web_root: "%kernel.root_dir%/../public"
            cache_prefix: "media/cache"
    webp:
        generate: true
```

```yaml
# config/routes/liip_imagine.yaml
liip_imagine_filter:
    path: /media/cache/{filter}/{path}
    defaults:
        _controller: '%liip_imagine.controller.filter_action%'
    methods:
        - GET
    requirements:
        filter: '[A-z0-9_-]*'
        path: .+

```

### DIFF
--- a/Resources/doc/basic-usage.rst
+++ b/Resources/doc/basic-usage.rst
@@ -114,6 +114,11 @@ when the first page request is served. The transformed image is then cached
 for subsequent requests. The final cached image path would be similar to
 ``/media/cache/my_thumb/relative/path/to/image.jpg``.
 
+.. tip::
+
+    You can :doc:`prepare the cache in advance <commands>` and use the ``imagine_filter_cache`` filter to always
+    return a link to the final cached image.
+
 .. note::
 
     Using the ``dev`` environment you might find that images are not properly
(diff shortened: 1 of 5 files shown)


dbu:
thanks a lot!

do you want to add the nginx trick to use this even without async cache warming to the documentation? (it would also prevent race conditions when a new image is uploaded and immediately looked at)

PR #1351, opened 2021-01-18
add CacheManagerInterface.php that will allow CacheManager.php to be …

Warxcell:
…decorated.

| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | yes
| BC breaks? | no I think
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR |

### DIFF
--- a/DependencyInjection/LiipImagineExtension.php
+++ b/DependencyInjection/LiipImagineExtension.php
@@ -14,6 +14,7 @@ namespace Liip\ImagineBundle\DependencyInjection;
 use Liip\ImagineBundle\DependencyInjection\Factory\Loader\LoaderFactoryInterface;
 use Liip\ImagineBundle\DependencyInjection\Factory\Resolver\ResolverFactoryInterface;
 use Liip\ImagineBundle\Imagine\Cache\CacheManager;
+use Liip\ImagineBundle\Imagine\Cache\CacheManagerInterface;
 use Liip\ImagineBundle\Imagine\Data\DataManager;
 use Liip\ImagineBundle\Imagine\Filter\FilterManager;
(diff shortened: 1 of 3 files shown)


lsmith77:
can you remove all the formatting changes from the PR?
also can you briefly explain the specific use case you want to achieve with this change?
ie. for what use case do you want to decorate the CacheManger?

Warxcell:
Don't remember - I think I decided to go with Adapter Pattern in my case, but I think it's useful anyway. :)

Warxcell:
Basically I needed to add `$runtimeConfig` for particular `path`, globally - not just for one file. So to avoid Repeat-Your-Self I could decorate CacheManager and do 

```php
if(stripos($path, 'XXX') !== false)  {
    $runtimeConfig['YY'] = ['xxx' => 'xxx']
}
return $this->decorated->getBrowserPath($path, $filter, $runtimeConfig.... );`
```

lsmith77:
ok .. as long as there is a use-case I am fine adding an interface :)

fbourigault:
The `CacheManager` has two responsibilities.

The first one, for the `isStored`, `store`, `remove` and `resolve` methods, is to find the right resolver and call the same method this resolver.

The second one, for the `getBrowserPath`, `getRuntimePath` and `generateUrl`, is to get an URL or a path.

I see here a nice opportunity to split `CacheManager`.

The code is quite complex, but I would start with introducing some kind of `ResolverManagerInterface` to move the first responsibility out of `CacheManager`.

dbu:
i like the suggestion of @fbourigault - do you want to split the interface in the way he suggested @Warxcell ?

PR #1356, opened 2021-02-18
Return default image with redirect response code

bszyszka:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | no
| Fixed tickets | 
| License | MIT
| Doc PR |

### DIFF
--- a/Controller/ImagineController.php
+++ b/Controller/ImagineController.php
@@ -143,7 +143,7 @@ class ImagineController
             return new RedirectResponse($url(), $this->controllerConfig->getRedirectResponseCode());
         } catch (NotLoadableException $exception) {
             if (null !== $this->dataManager->getDefaultImageUrl($filter)) {
-                return new RedirectResponse($this->dataManager->getDefaultImageUrl($filter));
+                return new RedirectResponse($this->dataManager->getDefaultImageUrl($filter), $this->controllerConfig->getRedirectResponseCode());
             }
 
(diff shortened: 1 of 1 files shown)


lsmith77:
can you add a bit more context here? I guess this should be classified as a bug fix? could you look into adding a test case?

Controller/ImagineController.php
```diff
@@ -143,7 +143,7 @@ private function createRedirectResponse(\Closure $url, string $path, string $fil
             return new RedirectResponse($url(), $this->controllerConfig->getRedirectResponseCode());
         } catch (NotLoadableException $exception) {
             if (null !== $this->dataManager->getDefaultImageUrl($filter)) {
-                return new RedirectResponse($this->dataManager->getDefaultImageUrl($filter));
+                return new RedirectResponse($this->dataManager->getDefaultImageUrl($filter), $this->controllerConfig->getRedirectResponseCode());
```
peter-gribanov, line 146: I do not think it's a good idea. You make a 301 redirect to the default image if the requested image is not found. If the image is added, the user will not see it because his request has been cached.

dbu (commented):
thanks @bszyszka . i agree with @peter-gribanov here, we should not treat the not found image the same as when things worked, to avoid the browser remembering.

PR #1361, opened 2021-03-05
[Doc] Simplified the "strong" role

javiereguiluz:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | MIT
| Doc PR | -

Hi! We're going to introduce some changes in the RST parser of symfony.com and the `:strong:` role is not going to be supported. That's why we suggest you to replace it by the other common way of adding strong emphasis (see https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#strong-emphasis).

PR #1367, opened 2021-04-03
[Doc] Simplified the strong role, follow up for, #1361

lsmith77:
@javiereguiluz is this correct?


javiereguiluz (approved):
Thanks for backporting this to 1.x branch!

PR #1265, opened 2020-02-07
Create Lazy-Loaded Twig Extension

emmanuelballery:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1261
| License | MIT
| Doc PR | 

Hi!

Twig supports [Lazy-Loaded Extensions](https://symfony.com/doc/current/templating/twig_extension.html#creating-lazy-loaded-twig-extensions) since `1.35.0` and `2.4.4`.

However, I don't really know how we can be sure that Twig is indeed `>= 1.35.0` or `>= 2.4.4`.

I updated `composer.json` `require-dev` for `twig/twig` to `^1.35|^2.4.4|^3.0`.

Tests have been updated too.

(Also, @mbabker provided details and Blackfire comparison in #1261)

Issue #1261, opened 2020-01-10
Use a Twig runtime service for the Twig extension


mbabker:
**Is your feature request related to a problem? Please describe.**
In one of my applications, we are using AWS S3 through Flysystem as our backend storage.  While profiling the application during a round of performance review, we noticed that the FilterExtension from this bundle was the only Twig extension that shows in the Blackfire timeline, and this is primarily because of the time spent instantiating the AWS S3Client.  The current setup results in the CacheManager service and the S3Client being created on every request, even if there is no need for them to be loaded.

By creating a Twig runtime service, these services can be lazily loaded when they're needed and avoid the overhead of instantiating adapters for third party services when not necessary.

(Note, this is a really minor performance hit, but significant enough with the application's current setup that I figured the feature request would make sense)

**Describe the solution you'd like**
Use a Twig runtime service for the Twig extension so that the CacheManager does not get instantiated on every request

**Describe alternatives you've considered**
N/A

**Additional context**
I made a couple of changes inside my application for testing, and below are screenshots from the Blackfire comparison showing the difference on our homepage when a Twig runtime service is used compared to the current setup; unfortunately I didn't have time to look at the bundle's tests to submit this as a pull request.

![Screen Shot 2020-01-10 at 2 25 13 PM](https://user-images.githubusercontent.com/368545/72184869-0be7e500-33b7-11ea-95bc-5acf19c177b1.png)

![Screen Shot 2020-01-10 at 2 32 42 PM](https://user-images.githubusercontent.com/368545/72184888-13a78980-33b7-11ea-827f-ae49da781d87.png)

dbu:
fixed in #1376

PR #1376, opened 2021-05-21
Create Lazy-Loaded Twig Extension

emmanuelballery:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | #1261
| License | MIT
| Doc PR | 

Hi!

This is a new version for #1265 ;

Twig supports [Lazy-Loaded Extensions](https://symfony.com/doc/current/templating/twig_extension.html#creating-lazy-loaded-twig-extensions) since `1.35.0` and `2.4.4`.

Tests have been updated too.

(Also, @mbabker provided details and Blackfire comparison in #1261)


Templating/FilterExtension.php
```diff
@@ -16,15 +16,13 @@
 
 class FilterExtension extends AbstractExtension
 {
-    use FilterTrait;
```
emmanuelballery, line 19: This may be a BC since `FilterExtension` is not `@internal` and people may have extended it - a strange thing to do. But it's also the only way to prevent Symfony container to load the `CacheManager` when it's not necessary.

Tests/Templating/FilterRuntimeTest.php
```diff
@@ -0,0 +1,41 @@
...
+class FilterRuntimeTest extends AbstractFilterTest
+{
+    public function testInstanceOfTwigFilter(): void
+    {
+        $this->assertInstanceOf(RuntimeExtensionInterface::class, $this->createTemplatingMock());
+    }
+
+    protected function createTemplatingMock(CacheManager $manager = null): FilterRuntime
+    {
+        if (!class_exists(RuntimeExtensionInterface::class)) {
```
dmaicher: This cannot happen I think? twig is a hard dependency and the supported versions should have this interface

emmanuelballery:
Sorry @dmaicher just checked again and this question was rulled out here https://github.com/sulu/sulu/issues/5556#issuecomment-832932399_ in someone else attempt to stop loading this extension.

Feel free to close then, I'll disable this extension per project I guess.

dmaicher:
@emmanuelballery I'm not a member on this repo :wink: I just stumbled across your PR as I also noticed the non-lazy twig extension and all the services that are instantiated although mostly not used in my case. So I would like to see this implemented :+1: 

I believe this would need to go to  `3.x` as its for sure a BC break.

lsmith77:
Maybe we can figure out a non BC breaking approach and then I would be fine to have something like this in 2.x

/cc @fbourigault

dmaicher:
> Maybe we can figure out a non BC breaking approach and then I would be fine to have something like this in 2.x
> 
> /cc @fbourigault

One way of achieving this with BC would be to
- create a new Extension class like `FilterRuntimeExtension` or `LazyFilterExtension` 
- deprecate the current `FilterExtension` and the `FilterTrait` 
- add a config parameter `liip_imagine.twig_runtime_extension: false` that is false by default and deprecate not setting it to true (+ we could set it to true on the [recipe](https://github.com/symfony/recipes-contrib/blob/master/liip/imagine-bundle/1.8/config/packages/liip_imagine.yaml#L2))
- register the old extension if its false and the new one if its true

Then on 3.0 we could remove the old one and always use the lazy one with the runtime.

Or does someone have a better idea?

emmanuelballery:
I'd:
- provide the new `LazyFilterExtension`
- deprecate the rest
- give people the choice whether they want to move on by registering the new service themselves (overriding the existing one - this way we don't add extra options to remove for 3.x)

```yaml
liip_imagine.templating.filter_extension:
    class: Liip\ImagineBundle\Templating\LazyFilterExtension
    tags: [ 'twig.extension' ]
```

Then in 3.x register `LazyFilterExtension` ourselves ;

lsmith77:
both approaches sound fine for me.

FYI if you rebase, PHPCS fixer should work again.

Resources/config/imagine.xml
```diff
@@ -190,6 +190,17 @@
...
+        <!--
+        @todo remove previous service and add these 2 in 3.x
+        <service id="liip_imagine.templating.filter_extension" class="Liip\ImagineBundle\Templating\LazyFilterExtension" public="false">
+            <tag name="twig.extension" />
+        </service>
+        <service id="liip_imagine.templating.filter_runtime" class="Liip\ImagineBundle\Templating\LazyFilterRuntime" public="false">
+            <tag name="twig.runtime" />
+            <argument type="service" id="liip_imagine.cache.manager" />
+        </service>
+        -->
```
emmanuelballery: Not sure if I had to do this.

In 2.x people can override this themselves:

```yaml
    liip_imagine.templating.filter_extension:
        class: Liip\ImagineBundle\Templating\LazyFilterExtension
        tags: [ 'twig.extension' ]
    liip_imagine.templating.filter_runtime:
        class: Liip\ImagineBundle\Templating\FilterRuntime
        arguments: [ '@liip_imagine.cache.manager' ]
        tags: [ 'twig.runtime' ]
```

Templating/LazyFilterExtension.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getFilters()
+    {
+        return [
+            new TwigFilter('imagine_filter', [FilterRuntime::class, 'filter']),
+        ];
+    }
+}
```
dmaicher, line 36: I think we are missing

```php
    /**
     * {@inheritdoc}
     */
    public function getName()
    {
        return 'liip_imagine';
    }
```

or?

Templating/FilterExtension.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Liip\ImagineBundle\Templating;
 
+@trigger_error('The ' . FilterExtension::class . ' class is deprecated since version 2.2 and will be removed in 3.0; use ' . LazyFilterExtension::class . ' instead.', E_USER_DEPRECATED);
```
dmaicher:
```suggestion
@trigger_error('The ' . FilterExtension::class . ' class is deprecated since version 2.7 and will be removed in 3.0; use ' . LazyFilterExtension::class . ' instead.', E_USER_DEPRECATED);
```

?

Templating/FilterTrait.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Liip\ImagineBundle\Templating;
 
+@trigger_error('The ' . FilterTrait::class . ' trait is deprecated since version 2.7 and will be removed in 3.0; use Twig instead.', E_USER_DEPRECATED);
```
dmaicher: we should also add `@deprecated` class annotations for both I think

dmaicher (commented):
I would have preferred the other approach with a simpler upgrade path for the users but this is also fine for me :+1:

Templating/LazyFilterExtension.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Templating;
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+
+class LazyFilterExtension extends AbstractExtension
```
dmaicher: I think we should make this class final. There is no point in extending it

Templating/LazyFilterExtension.php
emmanuelballery: Missed the fact that Twig 1 was still supported :/

dmaicher (approved):

Templating/FilterTrait.php
emmanuelballery: I followed what was done in the other files, but I agree with you.

emmanuelballery:
@dmaicher Maybe I did not completely understand your first suggestion. I was under the impression you were asking to add a configuration and that this configuration would then be obsolete in `3.x` (more work for this bundle maintainers). I chose this approach because IMO only advanced users are interested in preventing Twig from loading more container's stuff than necessary.

- The upgrade path for this bundle is only to delete what's deprecated in `3.x` and uncoment the new config ;
- The upgrade path for users is to override the config now and remove (if they want) their configuration after :

```yaml
    liip_imagine.templating.filter_extension:
        class: Liip\ImagineBundle\Templating\LazyFilterExtension
        tags: [ 'twig.extension' ]
    liip_imagine.templating.filter_runtime:
        class: Liip\ImagineBundle\Templating\FilterRuntime
        arguments: [ '@liip_imagine.cache.manager' ]
        tags: [ 'twig.runtime' ]
```

I'm completly ok to work on your approach if maintainers wants me to 👍🏻

dmaicher:
> @dmaicher Maybe I did not completely understand your first suggestion. I was under the impression you were asking to add a configuration and that this configuration would then be obsolete in `3.x` (more work for this bundle maintainers). I chose this approach because IMO only advanced users are interested in preventing Twig from loading more container's stuff than necessary.
> 
>     * The upgrade path for this bundle is only to delete what's deprecated in `3.x` and uncoment the new config ;
> 
>     * The upgrade path for users is to override the config now and remove (if they want) their configuration after :
> 
> 
> ```yaml
>     liip_imagine.templating.filter_extension:
>         class: Liip\ImagineBundle\Templating\LazyFilterExtension
>         tags: [ 'twig.extension' ]
>     liip_imagine.templating.filter_runtime:
>         class: Liip\ImagineBundle\Templating\FilterRuntime
>         arguments: [ '@liip_imagine.cache.manager' ]
>         tags: [ 'twig.runtime' ]
> ```
> 
> I'm completly ok to work on your approach if maintainers wants me to 👍🏻

The downside of your approach is that users might see a deprecation notice since they are using the old filter. Then what do they do? They would need to find this PR (or check the CHANGELOG or so) to find out that they need to register a service themselves to fix it. Then once 3.0 is released they need to remember to remove that custom service definition again as it works without then.

So my idea was to make this a bit simpler. They get a deprecation because of the added config option. They set it to true and for 99.9% of all users this works fine. Only if someone extended the filter or so they need to rework it.
Then once 3.0 is released they get an error because of the removed config option. They remove the option and all works.
EDIT: also for new installations using the recipe we could immediately use the new filter runtime for those users (config option true)

But in the end its not for me to decide :blush: I'm fine with both options

dmaicher:
@lsmith77 how can we move this forward? :blush: Can I help somehow?

lsmith77:
@dbu / @Tobion

dmaicher:
Friendly ping :blush: Can a maintainer review this?

Templating/FilterExtension.php
dbu: with the current version of the PR, this is no BC break anymore, right?

Resources/config/imagine.xml
dbu: i think it would be elegant to move this definition into its own file, and the above one for the old extension into a separate file too.

and add a configuration option to select which twig mode to use. (none/lazy/legacy, defaulting to legacy)

the deprecation note can then also tell the configuration option to switch to the lazy extension.

dbu:
sorry for the long silence. i have no some time reserved each week to keep LiipImagineBundle alive.

i added some minor comments, but like where this is going. can you please have a look at them?

and can you please rebase the branch on 2.x to get the latest changes and ci-adjustments in?

Resources/config/imagine.xml
dmaicher: yeah choosing the runtime extension via a config setting (+ trigger deprecation for legacy non-lazy way) would also be my preferred way

emmanuelballery:
God ... I messed up 😞

emmanuelballery:
Sorry for the noise ... @dmaicher is it what you were thinking of?

630016fb7a65ad937d5e51c25d5c6d8e77b61604

I am really rusty regarding bundle configuration...

DependencyInjection/Configuration.php
```diff
@@ -195,6 +195,26 @@ public function getConfigTreeBuilder()
                     ->end()
                 ->end()
             ->end()
+            ->scalarNode('twig_mode')
+                ->defaultValue('legacy')
+                ->info('Twig mode: none/lazy/legacy (default)')
```
dmaicher, line 200: What is the difference between `none` and  `legacy`? Wondering if we need a `none` mode?

Or should this be used if someone does not need the twig extension/runtime at all and we should not register any services then?

emmanuelballery: I took this suggestion from @dbu's comment.
- `none` : no runtime, dependencies will be loaded with the container ;
- `lazy`: with runtime, dependencies will be loaded only when needed ;
- `legacy` is only here temporarily to inform users they now have the choice (triggering a deprecation) and will be removed in `3.0` - same as `none` regarding dependencies loading ;

dbu: my idea of `none` was to not load the twig extension at all ;-)

is there a use case for loading the twig extension in legacy mode, that is without the lazy loading? (apart from having it until version 3 for BC, of course). if there is a use case, i'd call the modes none/lazy/eager and default to eager. 

but with the code deprecations, my understanding is that eager loading will be removed? then imho it would make sense to have none => do not load twig at all , lazy => what you did, legacy => the current legacy

dbu:
can you please rebase on 2.x? that should improve the build situation...

i think the 8.0 test failure is because of our deprecation warnings. afaik we need to mark the tests for deprecated things with `@group legacy` to have phpunit not complain: https://symfony.com/doc/current/components/phpunit_bridge.html#mark-tests-as-legacy

DependencyInjection/Configuration.php
dmaicher: Ok makes sense to me :blush: Let's keep `none/lazy/legacy` then and do not load any twig services with `none`. For the next major we can remove `legacy` then.

DependencyInjection/Configuration.php
```diff
@@ -195,6 +195,26 @@ public function getConfigTreeBuilder()
                     ->end()
                 ->end()
             ->end()
+            ->scalarNode('twig_mode')
```
franmomu:
```suggestion
            ->enumNode('twig_mode')
                ->values('none', 'lazy', 'legacy')
```

We could use `enumNode` and there is no need of the first `validate` call.

dmaicher (approved):

dbu (approved):
great, i am happy with this now!

one last thing: can you please document the new twig_mode option in https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/configuration.rst?

emmanuelballery:
I used this code in my projects and found some stupid mistakes ... sorry.

- `enumNode` in `Configuration` needed an `array` as parameters ;
- my `imagine_twig_mode_lazy.xml` was not referencing the new `Runtime` class ;
- I added documentation ;

dbu (approved):
awesome, thanks a lot!

dbu:
thanks for having tested it. the mistake with the configuration would hopefully also have shown when github runs the tests. (it is always waiting for my confirmation in PRs from non repository members, afaik because some assholes started mining bitcoin on ci runners...)

can you please look at the cs fixer?

and try the `@group legacy` annotations?

> i think the test failure is because of our deprecation warnings. afaik we need to mark the tests for deprecated things with @group legacy to have phpunit not complain: https://symfony.com/doc/current/components/phpunit_bridge.html#mark-tests-as-legacy

dbu:
cheers! i will look into releasing a new minor version hopefully this wednesday.

emmanuelballery:
@dbu I may have broken the configuration doc ... 😞 I guess I had to add 2 spaces before each new line for it to work.

https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/configuration.rst

image

dbu:
thanks for the heads up. fixing this in #1396

PR #1217, opened 2019-08-27
Fix Mime deprecations for Symfony 4.3

franmomu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1200
| License | MIT
| Doc PR | 

An attempt to remove the deprecations related to mime type and extension guessers based on https://github.com/symfony/symfony/commit/74ca91deaa014a4e776671bfc523bd10077a52eb#diff-b7fc65c7d852312152e353f395fc70a8. Any feedback is more than welcome 🙂 

https://github.com/liip/LiipImagineBundle/issues/1166 can be closed as well.

PR #1379, opened 2021-06-08
Fix service definitions for mime and extension guessers

### LINKED ISSUE #1287: Incompatibility with Symfony 5: Usage of class that does not exist anymore
**Symfony version(s) affected**: 5.1.0 and 5.0.9

**Description**

This error message is thrown when I call `bin/console`:

    In getLiipImagine_MimeTypeGuesserService.php line 20:
    Attempted to load class "MimeTypeGuesser" from namespace "Symfony\Component\HttpFoundation\File\MimeType".
    Did you forget a "use" statement for another namespace?

The classes specified in... https://github.com/liip/LiipImagineBundle/blob/16c28a8c84aee044504a3fe5db9f5282e53e5f95/Resources/config/imagine.xml#L368-L374 ...do not exist in SF 5.x anymore. So it seems liip/imagine-bundle is not fully compatible with sf 5 yet. 

**How to reproduce** 

These commands run WITHOUT any problem:

    symfony new --version=4.4 foo
    cd foo
    composer req liip/imagine-bundle          (answer interactive question with 'y'!)
    bin/console -vvv

These commands lead to the error message mentioned above:

    symfony new --version=5.1 foo
    cd foo
    composer req liip/imagine-bundle          (answer interactive question with 'y'!)
    bin/console -vvv

**Additional context**  

    Exception trace:
    at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/getLiipImagine_MimeTypeGuesserService.php:20
    ContainerYn76KfG\getLiipImagine_MimeTypeGuesserService::do() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/App_KernelDevDebugContainer.php:169
    ContainerYn76KfG\App_KernelDevDebugContainer->load() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/getLiipImagine_Binary_MimeTypeGuesserService.php:23
    ContainerYn76KfG\getLiipImagine_Binary_MimeTypeGuesserService::do() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/App_KernelDevDebugContainer.php:169
    ContainerYn76KfG\App_KernelDevDebugContainer->load() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/getLiipImagine_Filter_ManagerService.php:49
    ContainerYn76KfG\getLiipImagine_Filter_ManagerService::do() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/App_KernelDevDebugContainer.php:169
    ContainerYn76KfG\App_KernelDevDebugContainer->load() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/getLiipImagine_Command_CacheRemoveService.php:24
    ContainerYn76KfG\getLiipImagine_Command_CacheRemoveService::do() at /Volumes/sd_card/foo/var/cache/dev/ContainerYn76KfG/App_KernelDevDebugContainer.php:169
    ContainerYn76KfG\App_KernelDevDebugContainer->load() at /Volumes/sd_card/foo/vendor/symfony/dependency-injection/Container.php:441
    Symfony\Component\DependencyInjection\Container->getService() at /Volumes/sd_card/foo/vendor/symfony/dependency-injection/Argument/ServiceLocator.php:40
    Symfony\Component\DependencyInjection\Argument\ServiceLocator->get() at /Volumes/sd_card/foo/vendor/symfony/console/CommandLoader/ContainerCommandLoader.php:45
    Symfony\Component\Console\CommandLoader\ContainerCommandLoader->get() at /Volumes/sd_card/foo/vendor/symfony/console/Application.php:520
    Symfony\Component\Console\Application->has() at /Volumes/sd_card/foo/vendor/symfony/console/Application.php:716
    Symfony\Component\Console\Application->all() at /Volumes/sd_card/foo/vendor/symfony/framework-bundle/Console/Application.php:142
    Symfony\Bundle\FrameworkBundle\Console\Application->all() at /Volumes/sd_card/foo/vendor/symfony/console/Descriptor/ApplicationDescription.php:91
    Symfony\Component\Console\Descriptor\ApplicationDescription->inspectApplication() at /Volumes/sd_card/foo/vendor/symfony/console/Descriptor/ApplicationDescription.php:68
    Symfony\Component\Console\Descriptor\ApplicationDescription->getCommands() at /Volumes/sd_card/foo/vendor/symfony/console/Descriptor/TextDescriptor.php:202
    Symfony\Component\Console\Descriptor\TextDescriptor->describeApplication() at /Volumes/sd_card/foo/vendor/symfony/console/Descriptor/Descriptor.php:55
    Symfony\Component\Console\Descriptor\Descriptor->describe() at /Volumes/sd_card/foo/vendor/symfony/console/Helper/DescriptorHelper.php:65
    Symfony\Component\Console\Helper\DescriptorHelper->describe() at /Volumes/sd_card/foo/vendor/symfony/console/Command/ListCommand.php:75
    Symfony\Component\Console\Command\ListCommand->execute() at /Volumes/sd_card/foo/vendor/symfony/console/Command/Command.php:258
    Symfony\Component\Console\Command\Command->run() at /Volumes/sd_card/foo/vendor/symfony/console/Application.php:929
    Symfony\Component\Console\Application->doRunCommand() at /Volumes/sd_card/foo/vendor/symfony/framework-bundle/Console/Application.php:99
    Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /Volumes/sd_card/foo/vendor/symfony/console/Application.php:264
    Symfony\Component\Console\Application->doRun() at /Volumes/sd_card/foo/vendor/symfony/framework-bundle/Console/Application.php:82
    Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /Volumes/sd_card/foo/vendor/symfony/console/Application.php:140
    Symfony\Component\Console\Application->run() at /Volumes/sd_card/foo/bin/console:43

mbabker:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | Fixes #1287
| License | MIT
| Doc PR | N/A

In #1217 several deprecations from `symfony/http-foundation` for Symfony 4.3 were addressed, but the related service configurations weren't updated, leaving the `liip_imagine.mime_type_guesser` and `liip_imagine.extension_guesser` services in a broken state when `symfony/http-foundation:^5.0` is used.

This PR aims to fix the service definitions by conditionally changing those two services to be aliases of the `mime_types` service the FrameworkBundle provides if a version of the FrameworkBundle that offers that service is installed.  This should retain B/C for Symfony 4.2 and earlier by using the existing definition which uses the classes from `symfony/http-foundation` while offering forward compatibility by mapping the services to the replacement API for the deprecated classes.  When support for Symfony 4.2 and earlier are removed, this compiler pass can be removed and the two services converted into aliases unconditionally.

### DIFF
--- a/DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
+++ b/DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+/**
+ * Changes the `liip_imagine.mime_type_guesser` and `liip_imagine.extension_guesser` services to be aliases of the
+ * `mime_types` service provided by the FrameworkBundle when available.
+ *
+ * This compiler pass can be removed when dropping support for Symfony 4.2 and earlier.
+ *
+ * @internal
+ */
+final class MaybeSetMimeServicesAsAliasesCompilerPass extends AbstractCompilerPass
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('mime_types')) {
+            $container->removeDefinition('liip_imagine.mime_type_guesser');
+            $container->removeDefinition('liip_imagine.extension_guesser');
+
+            $container->setAlias('liip_imagine.mime_type_guesser', 'mime_types');
+            $container->setAlias('liip_imagine.extension_guesser', 'mime_types');
+
+            $message = 'Replaced the "%s" and "%s" service definitions with aliases to "%s"';
+
+            $this->log($container, $message, 'liip_imagine.mime_type_guesser', 'liip_imagine.extension_guesser', 'mime_types');
+        }
+    }
+}
--- a/LiipImagineBundle.php
+++ b/LiipImagineBundle.php
@@ -16,6 +16,7 @@ use Liip\ImagineBundle\Async\Topics;
 use Liip\ImagineBundle\DependencyInjection\Compiler\DriverCompilerPass;
 use Liip\ImagineBundle\DependencyInjection\Compiler\FiltersCompilerPass;
 use Liip\ImagineBundle\DependencyInjection\Compiler\LoadersCompilerPass;
+use Liip\ImagineBundle\DependencyInjection\Compiler\MaybeSetMimeServicesAsAliasesCompilerPass;
 use Liip\ImagineBundle\DependencyInjection\Compiler\MetadataReaderCompilerPass;
 use Liip\ImagineBundle\DependencyInjection\Compiler\NonFunctionalFilterExceptionPass;
 use Liip\ImagineBundle\DependencyInjection\Compiler\PostProcessorsCompilerPass;
@@ -47,6 +48,7 @@ class LiipImagineBundle extends Bundle
         $container->addCompilerPass(new PostProcessorsCompilerPass());
         $container->addCompilerPass(new ResolversCompilerPass());
         $container->addCompilerPass(new MetadataReaderCompilerPass());
+        $container->addCompilerPass(new MaybeSetMimeServicesAsAliasesCompilerPass());
 
         if (class_exists(AddTopicMetaPass::class)) {
             $container->addCompilerPass(AddTopicMetaPass::create()


DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Changes the `liip_imagine.mime_type_guesser` and `liip_imagine.extension_guesser` services to be aliases of the
+ * `mime_types` service provided by the FrameworkBundle when available.
+ *
+ * This compiler pass can be removed when dropping support for Symfony 4.2 and earlier.
+ *
+ * @internal
+ */
+final class MaybeSetMimeServicesAsAliasesCompilerPass extends AbstractCompilerPass
```
mbabker, line 25: Purposefully final and internal to make it clear this is a B/C layer that can (and should) go away when older Symfony version support is dropped.

lsmith77:
@dbu / @Tobion

DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('mime_types')) {
+            $container->removeDefinition('liip_imagine.mime_type_guesser');
+            $container->removeDefinition('liip_imagine.extension_guesser');
```
dbu, line 34: could this be a BC break if people customized one of these services?

dbu (commented):
thanks for the PR.

i am a bit confused why you say we can remove the compiler pass when we drop support for symfony <= 4.2. if we then only revert this change, won't things be broken again? or do you mean that after that, we can just remove the service definitions for liip_imagine.mime_type_guesser and liip_imagine.extension_guesser alltogether?

DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
mbabker: Yes, but as is these definitions already are invalid when `symfony/http-foundation:>=5.0` is installed as the `Symfony\Component\HttpFoundation\File\MimeType` namespace just does not exist anymore.  So, there are 3 realistic (IMO) options:

1) This PR, which has a potential B/C break in changing a service definition to a service alias (and, I guess the same B/C concerns as the next item...)

2) Change the service definitions to use the right classes conditionally directly in the container extension class (which, depending on the B/C policy could also be a break because it will change the classes provided by these services and cause existing typehints or instanceof checks to fail if they aren't expecting the counterparts from the `symfony/mime` component)

3) Introduce new services that always point to the new classes and deprecate the existing services, which adds a lot of maintenance burden in having the bundle try to resolve to the correct services based on the active Symfony version, and if anyone is trying to use these services in their own applications they have to go through the effort of finding the right service for their Symfony version

IMO, I'd say this PR is the "best" option moving forward.

mbabker:
> i am a bit confused why you say we can remove the compiler pass when we drop support for symfony <= 4.2. if we then only revert this change, won't things be broken again? or do you mean that after that, we can just remove the service definitions for liip_imagine.mime_type_guesser and liip_imagine.extension_guesser alltogether?

The service definitions would be updated to the below when Symfony 4.3 is the minimum supported version:

```xml
<service id="liip_imagine.mime_type_guesser" alias="mime_types" />
<service id="liip_imagine.extension_guesser" alias="mime_types" />
```

(Or, those two service IDs could just be completely deprecated and the core `mime_types` service be used directly going forward)

The compiler pass is needed to handle cross-version compatibility while Symfony 3.4 is still supported.  Essentially, once you get to a newer minimum Symfony version, there's no need for the compiler pass because you can set it directly in the XML and not need any conditional changes later.

DependencyInjection/Compiler/MaybeSetMimeServicesAsAliasesCompilerPass.php
dbu: thanks for the clarification. yeah, lets do it this way. i will add a note in UPGRADE.md

PR #1381, opened 2021-06-10
Fix deprecation of PHPUnit 8

peter-gribanov:
| Q | A
| --- | ---
| Branch? | 2.0 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | 
| License | MIT
| Doc PR | 

Fix error

> expectExceptionMessageRegExp() is deprecated in PHPUnit 8 and will be removed in PHPUnit 9. Use expectExceptionMessageMatches() instead.

### DIFF
--- a/Tests/AbstractTest.php
+++ b/Tests/AbstractTest.php
@@ -87,6 +87,15 @@ abstract class AbstractTest extends TestCase
         ];
     }
 
+    public function expectExceptionMessageMatchesBC(string $regularExpression): void
+    {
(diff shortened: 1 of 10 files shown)


dbu:
thanks! i checked if we really need this, as we should support as narrow a range of phpunit as possible, but for as long as we support PHP 7.1 we have have to, phpunit 8 is only available from PHP 7.2 or better.

PR #1382, opened 2021-06-21
Remove urldecode path

Seb33300:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | https://github.com/liip/LiipImagineBundle/issues/1143
| License | MIT
| Doc PR | 

It is not required to `urldecode` since it is already decoded.
And this is causing issue when the filename contains `+` which is decoded as a space.

### DIFF
--- a/Controller/ImagineController.php
+++ b/Controller/ImagineController.php
@@ -15,7 +15,6 @@ use Imagine\Exception\RuntimeException;
 use Liip\ImagineBundle\Config\Controller\ControllerConfig;
 use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
 use Liip\ImagineBundle\Exception\Imagine\Filter\NonExistingFilterException;
-use Liip\ImagineBundle\Imagine\Cache\Helper\PathHelper;
 use Liip\ImagineBundle\Imagine\Cache\SignerInterface;
 use Liip\ImagineBundle\Imagine\Data\DataManager;
 use Liip\ImagineBundle\Service\FilterService;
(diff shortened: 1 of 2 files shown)


Controller/ImagineController.php
```diff
@@ -82,7 +81,6 @@ public function __construct(
      */
     public function filterAction(Request $request, $path, $filter)
     {
-        $path = PathHelper::urlPathToFilePath($path);
```
dbalabka, line 85: Could you please describe the case when it is not required to convert URL into File system path? 
It would be great to cover it with test.

dbalabka (changes requested):

Controller/ImagineController.php
Seb33300: If you look into the `urlPathToFilePath` method you will see this is only a `urldecode` applied to the `$path`.
But at this point, `$path` is already decoded when received in `filterAction`.

lsmith77:
@dbu / @Tobion

dbu:
the path is `rawurlencoded` by either PathHelper::filePathToUrlPath (when cache exists) or by `UrlGenerator::doGenerate` from Symfony (when no cache exists and we go to the ImagineController). the symfony router immediately reverts many of the encodings and changes `%2B` to `+`, which is why you saw errors with `+` in filenames.

we are fixing this in a way that should work for all cases :crossed_fingers: in #1391

dbu:
thanks a lot for the pull request nonetheless. it helped us with investigating the problems!

PR #1387, opened 2021-08-03
Change registering the form theme to prepend the configuration

mbabker:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | N/A
| License | MIT
| Doc PR | N/A

Per the [TwigBundle documentation](https://symfony.com/doc/current/form/form_themes.html#applying-themes-to-all-forms), the latest registered from themes have the highest priority.  Right now, this bundle is appending its form theme to the end of the list when the container is built, which effectively makes it the highest priority form theme and something that cannot be overridden.

By changing the container extension to prepend the form theme param to the TwigBundle config, we can get the form theme loaded in at a lower priority that will allow an app's own form theme(s) to override it if so desired.

`twig.form.resources` param before this change from the app I tested this change with:

```php
[
    'twig.form.resources' => [
        0 => 'form_div_layout.html.twig',
        1 => '@MisdPhoneNumber/Form/phone_number.html.twig',
        2 => 'form/app_form_layout.html.twig',
        3 => '@LiipImagine/Form/form_div_layout.html.twig',
    ],
]
```

After this change:

```php
[
    'twig.form.resources' => [
        0 => 'form_div_layout.html.twig',
        1 => '@MisdPhoneNumber/Form/phone_number.html.twig',
        2 => '@LiipImagine/Form/form_div_layout.html.twig',
        3 => 'form/app_form_layout.html.twig',
    ],
]
```

### DIFF
--- a/DependencyInjection/LiipImagineExtension.php
+++ b/DependencyInjection/LiipImagineExtension.php
@@ -20,12 +20,13 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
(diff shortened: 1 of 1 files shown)


dbu (approved):
makes sense to me :+1: 

thanks for this improvement.

PR #1391, opened 2021-10-05
fix handling of + in image names, added some functional tests

dbu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | fix #1382
| License | MIT
| Doc PR | -

Added some functional tests for handling of special characters, and fixing to not break for `+`.


Imagine/Cache/Helper/PathHelper.php
```diff
@@ -20,12 +20,14 @@ class PathHelper
...
     public static function filePathToUrlPath(string $path): string
     {
+        // this is not exactly the same as what symfony does in UrlGenerator::doGenerate where it strtr a bunch of encodings back to unencoded
+        // however, as we need to rawurldecode anyways, it should not hurt
         return implode('/', array_map('rawurlencode', explode('/', $path)));
     }
 
     public static function urlPathToFilePath(string $url): string
     {
-        // used urldecode instead of rawurlencode for BC safety to support "+" in URL
```
dbu, line 28: this did the opposite of what was intended: if there is a `+` in the filename, we would end up with ` `  instead.

dbalabka: @dbu thanks for this fix. I've missed this fact in #1162

dbu: no worries, it is quite confusing what is going on, and the encoding the the symfony route generator does is ... surprising.

i added some unit and functional tests to help us better understand what is happening.

Issue #168, opened 2013-04-02
cache busting


lsmith77:
figure out some solution to make it possible to add some cache busting capability, see #167 

i guess it should not be that hard to implement. imho the cleanest solution would be to simply append a top level directory (rather than adding something to the file name) with some hash. we would add some basic hash routing to all loaders and then the cache busting could be enabled on a per filter (plus maybe on a global level?). the generated route would be adjusted accordingly to have the relevant placeholder in the pattern.

jowe1:
@haavg it works like you said although the documentation says:
<img src="{{ asset('/relative/path/to/image.jpg') | imagine_filter('my_thumb') }}" />
on
http://symfony.com/doc/master/bundles/LiipImagineBundle/basic-usage.html

Please fix the documentation or explain why that is ...?!

fritzmg:
👍 for cache busting ability. I am in need for that as well currently and I have to resort to manually renaming the file.

michellesanver:
@fritzmg @jowe1 @lsmith77 I assume this issue is still relevant? If so I'll at least look at the documentation and see what can be further done about the cache busting to open a more clear workable issue about that.

dbu:
it seems this is not a huge concern to most people, as nobody implemented it. i will close it.

dbu:
actually accidentally implemented it while working on symfony asset versioning integration #1397 :joy:

PR #1397, opened 2021-10-20
lazy twig extension: handle asset versioning, added imagine_filter_cache, better test coverage

dbu:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | -
| Fixed tickets | fix #1392, #1291, #1240, #1172
| License | MIT
| Doc PR | :heavy_check_mark: 

While looking over things, i came up with a different approach that is less invasive than `parse_url` and as a bonus actually nicely handles asset versioning. this accidentally even provides a simplistic solution for #168 (with just one global version, which could be helpful when changing filter sets)

### DIFF
--- a/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
+++ b/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\Asset\VersionStrategy\StaticVersionStrategy;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Inject the Symfony framework assets version parameter to the
+ * LiipImagineBundle twig extension if possible.
+ *
+ * We extract the version parameter from the StaticVersionStrategy service
+ * definition. If anything is not as expected, we log a warning and do nothing.
+ *
+ * The expectation is for the user to configure the assets version in liip
+ * imagine for custom setups.
+ *
+ * Anything other than StaticVersionStrategy needs to be implemented by the
+ * user in CacheResolveEvent event listeners.
+ */
+class AssetsVersionCompilerPass extends AbstractCompilerPass
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!class_exists(StaticVersionStrategy::class)
+            // this application has no asset version configured
+            || !$container->has('assets._version__default')
+            // we are not using the new LazyFilterRuntime
+            || !$container->hasDefinition('liip_imagine.templating.filter_runtime')
+        ) {
+            return;
+        }
+        $runtimeDefinition = $container->getDefinition('liip_imagine.templating.filter_runtime');
+        if (null !== $runtimeDefinition->getArgument(1)) {
+            // the asset version has been set explicitly
+            return;
+        }
+
+        $versionStrategyDefinition = $container->findDefinition('assets._version__default');
+        if (!is_a($versionStrategyDefinition->getClass(), StaticVersionStrategy::class, true)) {
+            $this->log($container, 'Symfony assets versioning strategy "'.$versionStrategyDefinition->getClass().'" not automatically supported. Configure liip_imagine.twig.assets_version if you have problems with assets versioning');
+
+            return;
+        }
+        $version = $versionStrategyDefinition->getArgument(0);
+        $format = $versionStrategyDefinition->getArgument(1);
+        $format = $container->resolveEnvPlaceholders($format);
+        if ($format && !$this->str_ends_with($format, '?%%s')) {
+            $this->log($container, 'Can not handle assets versioning with custom format "'.$format.'". asset twig function can likely not be used with the imagine_filter');
+
+            return;
+        }
+
+        $runtimeDefinition->setArgument(1, $version);
+    }
+
+    /**
+     * Can be replaced with the built-in method when dropping support for PHP < 8.0
+     */
+    private function str_ends_with(string $haystack, string $needle): bool
+    {
+        return mb_substr($haystack, -mb_strlen($needle)) === $needle;
+    }
+}
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -195,19 +195,25 @@ class Configuration implements ConfigurationInterface
                     ->end()
                 ->end()
             ->end()
-            ->enumNode('twig_mode')
-                ->defaultValue('legacy')
-                ->info('Twig mode: none/lazy/legacy (default)')
-                ->values(['none', 'lazy', 'legacy'])
-                ->validate()
-                    ->ifTrue(function ($v) {
-                        return 'legacy' === $v;
-                    })
-                    ->then(function ($v) {
-                        @trigger_error('Twig "legacy" mode has been deprecated and will be removed in 3.0. Use "none" or "lazy".', E_USER_DEPRECATED);
+            ->arrayNode('twig')
+                ->addDefaultsIfNotSet()
+                ->children()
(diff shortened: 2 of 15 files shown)


dbu:
@Schyzophrenic @GKFX you looked into this in the past, so maybe you have some feedback on this approach?

Schyzophrenic:
@dbu I had a very brief look at this. I could be wrong but if you clean the path of everything after the ? I assume you won't be able to handle the http queries of a distant host such as an image provider where the parameters are critical to serve the right image. 
I recall someone (you?) mentioned this could lead to security risks and that further discussions are required to decide whether this should be supported though. 
It looks cleaner than before otherwise.

dbu:
i inject the configured version string and check if the url ends with exactly that string. if it does not, i don't touch the url.

Schyzophrenic:
> i inject the configured version string and check if the url ends with exactly that string. if it does not, i don't touch the url.

So I was actually thinking of a URL like this one: https://images2.productserve.com/?w=200&h=200&bg=white&trim=5&t=letterbox&url=ssl%3Awww.cdiscount.com%2Fpdt2%2F9%2F2%2F5%2F1%2F700x700%2FNIN4902370537925.jpg&feedId=38753&k=421fa9c064b91b3719c999b1a900cfb30d7e23de
In that case, you don't know in advance how the version will actually look like. 
I am not very familiar with this though, so I could say something irrelevant...

dbu:
that should now work again when using `twig.mode: lazy`. (i am not changing the legacy twig extension anymore.)

Schyzophrenic:
> that should now work again when using `twig.mode: lazy`. (i am not changing the legacy twig extension anymore.)

You're right I see it now! Thanks for the precision

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -0,0 +1,62 @@
...
+        ) {
+            return;
+        }
+        $runtimeDefinition = $container->getDefinition('liip_imagine.templating.filter_runtime');
+        if (null !== $runtimeDefinition->getArgument(1)) {
+            // the asset version has been set explicitly
+            return;
+        }
+
+        $versionStrategyDefinition = $container->getDefinition('assets._version__default');
```
dbu: getting the default symfony versioning service to copy the configuration from it, if available.

is there a better solution for this?

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $runtimeDefinition->setArgument(1, $version);
+    }
+
+    /**
+     * Can be replaced with the built-in method when dropping support for PHP < 8.0
+     */
+    private function str_ends_with(string $haystack, string $needle): bool
+    {
+        return mb_substr($haystack, -mb_strlen($needle)) === $needle;
```
stof, line 71: no need to depend on mbstring for that (As UTF-8 is self-synchronizing, you won't do a mess when using non-mbstring functions to look for the end of the string being `?%%s')

DependencyInjection/Configuration.php
```diff
@@ -195,19 +195,25 @@ public function getConfigTreeBuilder()
                     ->end()
                 ->end()
             ->end()
-            ->enumNode('twig_mode')
```
stof, line 198: don't you need to use `beforeNormalization()` to rewrite `twig_mode` to `twig.mode` ? Or is it not released yet ?

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
stof: I don't see another solution

stof: should probably use `findDefinition` to handle the case of `assets._version__default` being an alias (changing `hasDefinition` into `has` in the check above)

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -0,0 +1,62 @@
...
+            return;
+        }
+        $runtimeDefinition = $container->getDefinition('liip_imagine.templating.filter_runtime');
+        if (null !== $runtimeDefinition->getArgument(1)) {
+            // the asset version has been set explicitly
+            return;
+        }
+
+        $versionStrategyDefinition = $container->getDefinition('assets._version__default');
+        if (!is_a($versionStrategyDefinition->getClass(), StaticVersionStrategy::class, true)) {
```
stof, line 49: if you want to be safe, you should probably resolve parameters in the class name here (the DI component still supports them, even though it is considered a bad extension point and so core services stopped using that pattern). As that code is meant to detect custom strategies too, you cannot assume that it is a core definition.

DependencyInjection/Configuration.php
dbu: exactly, we did not release the changes that introduced twig_mode. therefore i don't care about BC for that. if somebody is using a dev version of the bundle, they at least get an error when upgrading and can fix the configuration.

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
dbu: ah, here we probably don't strictly need it. we use mbstring in a couple of places already. i add ext-mbstring to composer.json and leave it here as well (php-cs-fixer also wants to have the mbstring variants, would need to disable that rule otherwise, and that could mess in other places where we do need it)

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
dbu: good idea!

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -0,0 +1,63 @@
...
+            $this->log($container, 'Symfony assets versioning strategy '.$versionStrategyDefinition->getClass().' not automatically supported. Configure liip_imagine.twig.assets_version if you have problems with assets versioning');
+
+            return;
+        }
+        $version = $versionStrategyDefinition->getArgument(0);
+        $format = $versionStrategyDefinition->getArgument(1);
+        $format = $container->resolveEnvPlaceholders($format);
```
dbu, line 56: @stof is this what you meant with resolving the parameters?

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
dbu: do i need the thing below, or on this line `$container->resolveEnvPlaceholders($versionStrategyDefinition->getClass())`?

dbu: or both?

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
stof: no. I said parameters, not env placeholders (env placeholders cannot be validated at compile-time, as their value can change at runtime by definition).

I meant `$container->getParameterBag()->resolveValue($format)` (although you might not need it as it is a BEFORE_REMOVING pass, and so parameters might already have been resolved by the DI component, to be checked)

dbu: to debug, i changed the [FrameworkExtension](https://github.com/symfony/symfony/blob/ee6bbb24c5fa30bf110de6ee976dc2585a744cd5/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1233) to set the format with a parameter and the [assets.php](https://github.com/symfony/symfony/blob/ee6bbb24c5fa30bf110de6ee976dc2585a744cd5/src/Symfony/Bundle/FrameworkBundle/Resources/config/assets.php#L69) service definition to use a class name parameter.

both get resolved before this compiler pass runs. so i only removed the wrong resolveEnvPlaceholders i added but don't add the parameter bag resolving.

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
dbu: i experimented with this and saw that parameters are resolved when we get to this compiler pass. probably indeed PassConfig::TYPE_BEFORE_REMOVING

PR #1404, opened 2021-10-28
default redirect with 302 instead of 301

### LINKED ISSUE #1260: Images not resolving when browser caching is enabled
Hello,

when I enable browsing cache using nginx

```
location ~* .(js|jpg|jpeg|gif|png|css|tgz|gz|rar|bz2|doc|pdf|ppt|tar|wav|bmp|rtf|swf|ico|flv|txt|woff|woff2|svg)$ {
expires 30d;
add_header Pragma "public";
}
```

images are not resolving(only resolving when I open developer tools and tick "Disavle cache when dev tools open" in Chrome)

image

dbu:
| Q | A
| --- | ---
| Branch? | 2.0
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | fix #1260
| License | MIT
| Doc PR | :heavy_check_mark: 

If we do a permanent redirect, the browser will usually cache this information and not call the controller anymore. 

If the template contains the link to the controller, the permanent redirect saves unnecessary calls to the controller.

However, if the cached image is deleted, we again generate the URL to the controller - but the browser will remember the permanent redirect and directly request the image that is not existing. #1260 describes nginx configuration to work around this, but the default configuration should be robust rather than optimized and only working with specific nginx configuration.


jusurb:
@dbu how is this not a BC break? it changes the default setting

dbu:
BC is often tricky. what is a bugfix to somebody is a BC break to others.

i decided to do this change because using 302 instead of 301 does not break anything for a client. doing 301 breaks things in a sneaky way when clearing the cache, and the issue that brought this up showed that several people had been very confused by it.

jusurb:
> i decided to do this change because using 302 instead of 301 does not break anything for a client

I would agree with you, if this was a client side javascript library 😄 

i would consider it as a BC break as default behaviour changed and it should be at least mentioned as such in the changelog

dbu:
i will add an explicit note about it in the changelog, agree that that is a good idea.

we use https://github.com/github-changelog-generator/github-changelog-generator that automatically collects all pull requests. convenient but less details than when updating changelog manually as the changes are merged.

jusurb:
Thank you 🙂

PR #1414, opened 2021-11-04
Bump to Symfony 4.4 and 5.3, fix deprecations for Symfony 5.4

franmomu:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| License | MIT
| Doc | 

This PR 
- bumps symfony packages to supported versions (`4.4` and `5.3`).
- adds a job to test Symfony 5.4
- removes deprecations using Symfony 5.4


DependencyInjection/Configuration.php
```diff
@@ -44,7 +44,7 @@ public function __construct(array $resolversFactories, array $loadersFactories)
     }
 
     /**
-     * {@inheritdoc}
+     * @return TreeBuilder
```
dbu: can we avoid this change? i dislike adding redundancy in phpdoc. 

or are the changes in this PR narrowing down the return types?

Templating/FilterExtension.php
```diff
@@ -24,9 +24,9 @@ class FilterExtension extends AbstractExtension
     use FilterTrait;
 
     /**
-     * {@inheritdoc}
+     * @return TwigFilter[]
      */
-    public function getFilters()
+    public function getFilters(): array
```
dbu: this is technically a BC break, in case somebody extends this class and overwrites getFilters. it sounds like a weird idea to extend the class, which is why we made the new LazyFilterExtension final.

as its deprecated anyways, i'd rather not change the method signature here.

dbu (commented):
thanks.

what were the actual deprecations with symfony 5.4?

i guess it still makes sense to do this in the 2.x branch, even though i created the 3.x branch. might still take a moment to get 3.x ready, and the cleanups seem BC.

Templating/FilterExtension.php
franmomu: oops sorry, I thought I only had changed `LazyFilterExtension`, this file was not meant to be changed.

DependencyInjection/Configuration.php
franmomu: The thing is that in Symfony 5.4, you get these deprecations: https://github.com/franmomu/LiipImagineBundle/runs/4102778452?check_suite_focus=true#step:11:124

```
1x: Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "Liip\ImagineBundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in ConfigurationTest::testImplementsConfigurationInterface from Liip\ImagineBundle\Tests\DependencyInjection
```
here it is explained https://wouterj.nl/2021/09/symfony-6-native-typing#symfony-bundle-maintainers-compatibility-plan

dbu: ah, i see. do our supported PHP versions allow to declare the return type even if the base class does not have it? then i would do that rather. at least for the Configuration class this should be reasonably safe.

if that is not possible, i'd rather supress those warnings in 2.x and change it in 3.x only, where we drop support for older PHP versions as well.

franmomu: Apparently it does: https://3v4l.org/vMpRt#v7.1.0

I've also changed the one in `LiipImagineExtension` that should be also safe I think.

Imagine/Filter/RelativeResize.php
```diff
@@ -45,6 +45,8 @@ public function __construct($method, $parameter)
 
     /**
      * {@inheritdoc}
+     *
+     * @return ImageInterface
```
dbu: why change it exactly for this filter?

i think for version 3.x we want to declare return types on the FilterInterface, but for version 2.x i would leave things alone. the interface already defines the return type annotation.

Form/Type/ImageType.php
```diff
@@ -56,15 +56,15 @@ public function configureOptions(OptionsResolver $resolver)
     }
 
     /**
-     * {@inheritdoc}
+     * @return string
```
dbu, line 59: this also leads to a symfony warning if we don't do it?

Imagine/Filter/RelativeResize.php
franmomu: To be honest, this one was automatically added by 

`SYMFONY_PATCH_TYPE_DECLARATIONS=force=phpdoc ./vendor/bin/patch-type-declarations`

I'll revert it

Form/Type/ImageType.php
franmomu: yep https://github.com/franmomu/LiipImagineBundle/runs/4102778452?check_suite_focus=true#step:11:132

Imagine/Filter/RelativeResize.php
dbu: thanks

DependencyInjection/Configuration.php
dbu: thanks

dbu (approved):
thanks, looks good to me now.

dbu:
i guess i wait with releasing until symfony 5.4.0 is released, just in case...

PR #1416, opened 2021-11-04
Bump PHPStan to level 4

franmomu:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

I'm targeting `2.x` since it fixes some PHPDoc or should I target `3.x`?


dbu:
i think i would prefer doing this against 3.x and not pushing too much into 2.x anymore. the upside would be that instead of annotations, we can then use actual type ceclarations in many places. wdyt?

i am moving the code into src/ and tests/ directories to follow best practices (and making live easier when grepping through code ;-) ) in #1417

franmomu:
> i think i would prefer doing this against 3.x and not pushing too much into 2.x anymore. the upside would be that instead of annotations, we can then use actual type ceclarations in many places. wdyt?
> 
> i am moving the code into src/ and tests/ directories to follow best practices (and making live easier when grepping through code ;-) ) in #1417

yeah, sure, I'll wait for that PR to be merged and I'll rebase

dbu:
i merged it :-)

franmomu:
~I'll create another PR adding type declarations and then come back to this one.~

~So it is fine if I bump php to `7.4` right?~ nevermind, I'll add them when I find something that is needed to be changed.

dbu:
i'd love for us to do static types on everything in 3.x

src/Async/ResolveCache.php
```diff
@@ -25,7 +25,7 @@ class ResolveCache implements \JsonSerializable
     private $path;
 
     /**
-     * @var array|\string[]|null
+     * @var array|string[]|null
```
dbu:
```suggestion
     * @var string[]|null
```

src/Events/CacheResolveEvent.php
```diff
@@ -45,7 +45,7 @@ class CacheResolveEvent extends BCEvent
     /**
      * Resource url.
      *
-     * @var null
+     * @var string|null
```
dbu: i think we should add the type declaration to the property, then we can remove the annotation completely.

src/Imagine/Filter/PostProcessor/PngquantPostProcessor.php
```diff
@@ -28,7 +28,7 @@
 class PngquantPostProcessor extends AbstractPostProcessor
 {
     /**
-     * @var string Quality to pass to pngquant
+     * @var array|string Quality to pass to pngquant
```
dbu: according to line 96, this should be int|int[]

dbu (commented):
thanks!

do you want to convert to static typing? though we can also leave as is and do that in a separate PR.

dbu:
do you want to rebase this PR? or first wrap up the last namespaces with static typing?

franmomu:
> do you want to rebase this PR? or first wrap up the last namespaces with static typing?

I think I'll finish the static typing and then maybe we could increase the level even more easily.

.phpstan/baseline.neon
```diff
@@ -3,3 +3,16 @@ parameters:
...
+        - message: "#^Call to an undefined method Symfony\\\\Component\\\\Config\\\\Definition\\\\Builder\\\\NodeParentInterface\\:\\:validate\\(\\)\\.$#"
+          count: 1
+          path: ../src/DependencyInjection/Configuration.php
+
+        # BC Layer for Symfony < 5.1
+        - message: "#^If condition is always true\\.$#"
+          count: 1
+          path: ../src/Controller/ImagineController.php
+
+        - message: "#^Unreachable statement \\- code above always terminates\\.$#"
```
dbu, line 13: is this a mistake of phpstan or of us? or some BC layer?

.phpstan/baseline.neon
```diff
@@ -3,3 +3,16 @@ parameters:
         - message: '#^Access to undefined constant static\(Liip\\ImagineBundle\\Config\\Filter\\Type\\FilterAbstract\)::NAME\.$#'
           paths:
               - ../src/Config/Filter/Type/FilterAbstract.php
+
+        - message: "#^Call to an undefined method Symfony\\\\Component\\\\Config\\\\Definition\\\\Builder\\\\NodeParentInterface\\:\\:validate\\(\\)\\.$#"
```
dbu: could we get rid of this by providing a stub to phpstan? (i guess not, i suppose the class name would clash with the current symfony NodeParentInterface)

if not, can you please add a comment for which symfony version this is?

franmomu:
I was going to increase to level 5, there are only 9 issues, but I'll create another PR.

.phpstan/baseline.neon
franmomu: Is the same BC as above, I've added a comment.

https://github.com/liip/LiipImagineBundle/blob/56a72358b3338a22d59bdbcb4a4c57221fa4a524/src/Controller/ImagineController.php#L110-L126

The `if` is always true with the version installed.

.phpstan/baseline.neon
franmomu: I've added an `assert`

.phpstan/baseline.neon
dbu: thanks!

hm, maybe phpstan should know about `version_compare` - it is kind of obvious that this is done for a reason i'd say...

ah, i think i found the solution: https://github.com/phpstan/phpstan/issues/2636 - we should list the Kernel::VERSION constant in dynamicConstantNames

.phpstan/baseline.neon
dbu: nice!

.phpstan/baseline.neon
franmomu: nice, thanks! didn't know about that one, I had to change one thing, I'll add the comment there.

src/Controller/ImagineController.php
```diff
@@ -117,7 +117,7 @@ private function getFiltersBc(Request $request): array
             }
         }
 
-        $runtimeConfig = $request->query->get('filters', []);
+        $runtimeConfig = $request->query->all()['filters'] ?? [];
```
franmomu: without this change, these are these errors:

```
 ------ ------------------------------------------------------------------------
  Line   Controller/ImagineController.php
 ------ ------------------------------------------------------------------------
  122    Call to function is_array() with string will always evaluate to false.
  126    Unreachable statement - code above always terminates.
 ------ ------------------------------------------------------------------------
```

`InputBag::get` is supposed to return always `string`.

dbu: this is the BC part for symfony < 5.1, because the `all` method accepts a parameter only since version 5.1.

imo the right thing here would be to keep the code and ignore the phpstan complaint, with a comment that this is the BC < symfony 5.1 part. due to the version check, we do the right thing.

.phpstan/baseline.neon
dbu: me neither, but i thought this must be a common issue and had a good guess what to search for :-D

src/Controller/ImagineController.php
dbu: oh sorry, i think i was not clear: i meant to ignore the new warning we get once we tell phpstan that the kernel version is dynamic :sweat_smile:

franmomu: yep, these are the new ones because `InputBag::get` always returns `string`.

```
 ------ ------------------------------------------------------------------------
  Line   Controller/ImagineController.php
 ------ ------------------------------------------------------------------------
  122    Call to function is_array() with string will always evaluate to false.
  126    Unreachable statement - code above always terminates.
 ------ ------------------------------------------------------------------------
```

dbu: ah, its about the is_array now. tricky tricky... 
thanks for the improvements

dbu (approved):

Issue #1406, opened 2021-10-28
remove deprecated code and documentation


dbu:
to prepare for version 3, remove all deprecated code and the documentation about it.

hint: search repository for "deprecated"

mbabker:
Could you add a `3.x` branch so folks could start working on these types of issues?  I assume with the new issues and 3.0 milestone, you're ready to start working on the new major version.

dbu:
i created the 3.x branch now.

stephanvierkant:
And look for 'BC'. Two examples I've found:

https://github.com/liip/LiipImagineBundle/blob/89e7563103ba82c5495f728f041d7135ee4bb5ff/src/Imagine/Cache/Resolver/ProxyResolver.php#L67

https://github.com/liip/LiipImagineBundle/blob/89e7563103ba82c5495f728f041d7135ee4bb5ff/src/Imagine/Cache/CacheManager.php#L218

And this BC is still required, please add some more details when to remove it.

tacman:
I attempted to remove these for version 3, and indeed many tests failed.

Not sure how to proceed.

dbu:
the only one i still find is in ProxyResolver. i did a bit of digging to see where this comes from, and its this pull request: https://github.com/liip/LiipImagineBundle/pull/687

in this instance, i think the "old" use case of just having a list of domains to randomly pick seems valid to me. the more i look at #687 the more i think we should even drop this in version 3 and just have a list of domains that are randomly picked. the replace functionality was never documented, so i think almost everyone is using just the base functionality, if they even use the proxy resolver at all.

PR #1418, opened 2021-11-04
Remove deprecated code

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | #1406
| License | MIT
| Doc | 

Cleanup related to deprecated code


dbu (approved):
Awesome, so many things that can be deleted!

I fixed the UPGRADE.md in the 3.x branch - can you please rebase on 3.x and add the necessary information to the UPGRADE.md file for the BC breaks? i think we should keep that file up to date as we delete code. if we try to only do it later, we will forget some of the BC breaks.

composer.json
```diff
@@ -39,7 +39,6 @@
         "aws/aws-sdk-php": "^2.4",
         "doctrine/cache": "^1.11|^2.0",
```
dbu: do we still need this or can we get rid of it, now that CacheResolver has been removed?

dbu (approved):
thanks for the cleanup!

composer.json
franmomu: you're right, there are no references to doctrine cache, thanks!

dbu:
thanks a lot!

PR #1422, opened 2021-11-09
Drop support PHP < 7.4 and add type declarations to Imagine folder

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

Adding type declarations to the whole code would be quite painful to review, this PR has two commits:
- drop support for PHP < 7.4 
- add type declarations to `Imagine` folder

I'll add a couple of comments in the code

### DIFF
--- a/src/Imagine/Cache/Resolver/FormatExtensionResolver.php
+++ b/src/Imagine/Cache/Resolver/FormatExtensionResolver.php
@@ -16,15 +16,9 @@ use Liip\ImagineBundle\Imagine\Filter\FilterConfiguration;
 
 class FormatExtensionResolver implements ResolverInterface
 {
-    /**
-     * @var ResolverInterface
-     */
-    private $resolver;
+    private ResolverInterface $resolver;
 
-    /**
-     * @var FilterConfiguration
-     */
-    private $filterConfig;
+    private FilterConfiguration $filterConfig;
 
     public function __construct(ResolverInterface $resolver, FilterConfiguration $filterConfig)
     {
@@ -35,37 +29,28 @@ class FormatExtensionResolver implements ResolverInterface
     /**
      * {@inheritdoc}
      */
-    public function resolve($path, $filter)
+    public function resolve(string $path, string $filter): string
     {
         $path = $this->replaceExtension($path, $filter);
 
         return $this->resolver->resolve($path, $filter);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function store(BinaryInterface $binary, $targetPath, $filter)
+    public function store(BinaryInterface $binary, string $path, string $filter): void
     {
-        $targetPath = $this->replaceExtension($targetPath, $filter);
+        $path = $this->replaceExtension($path, $filter);
 
-        return $this->resolver->store($binary, $targetPath, $filter);
+        $this->resolver->store($binary, $path, $filter);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function isStored($path, $filter)
+    public function isStored(string $path, string $filter): bool
     {
         $path = $this->replaceExtension($path, $filter);
 
         return $this->resolver->isStored($path, $filter);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function remove(array $paths, array $filters)
+    public function remove(array $paths, array $filters): void
     {
         $newPaths = [];
         foreach ($paths as $path) {
@@ -77,7 +62,7 @@ class FormatExtensionResolver implements ResolverInterface
             }
         }
 
-        return $this->resolver->remove($newPaths, $filters);
+        $this->resolver->remove($newPaths, $filters);
     }
 
     private function replaceExtension(string $path, string $filter): string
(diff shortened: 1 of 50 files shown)


src/Imagine/Cache/Resolver/FormatExtensionResolver.php
```diff
@@ -35,37 +29,28 @@ public function __construct(ResolverInterface $resolver, FilterConfiguration $fi
...
-     * {@inheritdoc}
-     */
-    public function store(BinaryInterface $binary, $targetPath, $filter)
+    public function store(BinaryInterface $binary, string $path, string $filter): void
     {
-        $targetPath = $this->replaceExtension($targetPath, $filter);
+        $path = $this->replaceExtension($path, $filter);
 
-        return $this->resolver->store($binary, $targetPath, $filter);
+        $this->resolver->store($binary, $path, $filter);
```
franmomu, line 43: the return value doesn't seem to be used, only some of the resolvers were returning a value, the same for `ResolverInterface::remove`.

src/Imagine/Cache/Resolver/PsrCacheResolver.php
```diff
@@ -202,22 +173,19 @@ private function generateIndexKey($cacheKey)
...
-     *
-     * @return string
-     */
-    private function sanitizeCacheKeyPart($cacheKeyPart)
+    private function sanitizeCacheKeyPart(?string $cacheKeyPart): string
     {
+        if (null === $cacheKeyPart) {
+            return '';
+        }
+
```
franmomu: this is because:

https://github.com/liip/LiipImagineBundle/pull/1422/files#diff-3ffe7291fc61244af59d9d29370815bc777fd8ef61d1b5035851ee7fd9c82f00R100

```php
$this->removePathAndFilter(null, $filter);
```

So the `string` passed can be `null`

src/Imagine/Cache/Resolver/FormatExtensionResolver.php
dbu: what kind of value returned those that returned anything?

src/Imagine/Cache/Resolver/PsrCacheResolver.php
dbu: does the place calling it with null end up with a safe cache key, or do we have to fix something on that side? it feels a bit strange, but maybe its ok

tests/Imagine/Filter/Loader/DownscaleFilterLoaderTest.php
```diff
@@ -61,8 +62,10 @@ public function provideSizes()
             ->willReturn($initialSize);
         $image
             ->method('resize')
-            ->willReturnCallback(function ($box) use (&$resultSize) {
+            ->willReturnCallback(function ($box) use ($image, &$resultSize): ImageInterface {
```
dbu: should we add a type declaration for $box as well?

dbu (commented):
thanks a lot, this will make the code much more readable. just some minor questions.

i'd love to challenge the large number of non-final classes and protected fields/methods, but in a separate PR. this one affects a lot of things already.

src/Imagine/Cache/Resolver/FormatExtensionResolver.php
franmomu: so in general, these methods are called from:

https://github.com/liip/LiipImagineBundle/blob/1fb3dd3dd6f99a9fdfc906189f59cbb821e4a215/src/Imagine/Cache/CacheManager.php#L219-L222

and

https://github.com/liip/LiipImagineBundle/blob/1fb3dd3dd6f99a9fdfc906189f59cbb821e4a215/src/Imagine/Cache/CacheManager.php#L253-L255

The ones that were returning values (`PsrCacheResolver`, `ProxyResolver` and `FormatExtensionResolver`) receive a `ResolverInterface` value in the constructor, so they were returning the value from the inner resolver.

dbu: ah i see. yeah then likely it is all because of no typing and not noticing that we never actually return anything.

tests/Imagine/Filter/Loader/DownscaleFilterLoaderTest.php
franmomu: sure, added it, I may have missed some. 

My idea was when we finishing adding type declarations, check the missing ones with phpstan

franmomu:
> thanks a lot, this will make the code much more readable. just some minor questions.
> 
> i'd love to challenge the large number of non-final classes and protected fields/methods, but in a separate PR. this one affects a lot of things already.

Yeah, I've been tempted to change some of the `protected` variables to `private`, but I think it's better in another PR.

src/Imagine/Cache/Resolver/PsrCacheResolver.php
franmomu: I've split this method into two (in a new commit):

https://github.com/liip/LiipImagineBundle/blob/1fb3dd3dd6f99a9fdfc906189f59cbb821e4a215/src/Imagine/Cache/Resolver/PsrCacheResolver.php#L152-L182

- `removePathAndFilter` now receives a `string $path` and `string $filter`
- `removeFilter` that only receives a `string $filter`

dbu: much cleaner, thanks!

PR #1423, opened 2021-11-09
Add more type declarations

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | yes/no
| Fixed tickets | 
| License | MIT
| Doc | 

Following https://github.com/liip/LiipImagineBundle/pull/1422

I think I haven't found any weird thing.

The remaining folders after this one are:
- Binary
- Config
- Factory
- Model
- tests

### DIFF
--- a/src/DependencyInjection/Compiler/MetadataReaderCompilerPass.php
+++ b/src/DependencyInjection/Compiler/MetadataReaderCompilerPass.php
@@ -11,6 +11,8 @@
 
 namespace Liip\ImagineBundle\DependencyInjection\Compiler;
 
+use Imagine\Image\Metadata\DefaultMetadataReader;
+use Imagine\Image\Metadata\ExifMetadataReader;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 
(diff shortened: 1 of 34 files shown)


src/DependencyInjection/Compiler/MetadataReaderCompilerPass.php
```diff
@@ -19,20 +19,11 @@
...
-     * @var string
-     */
-    private static $metadataReaderServiceId = 'liip_imagine.meta_data.reader';
+    private static string $metadataReaderServiceId = 'liip_imagine.meta_data.reader';
 
-    /**
-     * @var string
-     */
-    private static $metadataReaderDefaultClass = 'Imagine\Image\Metadata\DefaultMetadataReader';
+    private static string $metadataReaderDefaultClass = 'Imagine\Image\Metadata\DefaultMetadataReader';
```
dbu: we should use the `::class` pseudo constant for these

src/LiipImagineBundle.php
```diff
@@ -49,7 +49,7 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new ResolversCompilerPass());
         $container->addCompilerPass(new MetadataReaderCompilerPass());
 
-        /** @var $extension LiipImagineExtension */
+        /** @var LiipImagineExtension $extension */
```
dbu: should we convert the inline `@var` comments to `\assert($extension instanceof LiipImagineExtension);` to trigger warnings when this is not the case?

dbu (approved):
great! i noticed nothing incorrect, but found 2 things we might want to improve while we cleanup with newer language features...

appreciate the work you do. and good idea to do it by namespace and not all together in 1 monster PR.

dbu:
thank you!

PR #1424, opened 2021-11-11
Add type declarations

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

Almost the last one, after this, the only one missing directory is `tests` (apart from what I could've missed, but we'll catch them using static analysis).

I'll comment a couple of things I've found in the code

### DIFF
--- a/src/Binary/BinaryInterface.php
+++ b/src/Binary/BinaryInterface.php
@@ -13,18 +13,9 @@ namespace Liip\ImagineBundle\Binary;
 
 interface BinaryInterface
 {
-    /**
-     * @return string
-     */
-    public function getContent();
+    public function getContent(): string;
 
-    /**
-     * @return string
-     */
-    public function getMimeType();
+    public function getMimeType(): ?string;
 
-    /**
-     * @return string
-     */
-    public function getFormat();
+    public function getFormat(): ?string;
 }
--- a/src/Binary/Loader/AbstractDoctrineLoader.php
+++ b/src/Binary/Loader/AbstractDoctrineLoader.php
@@ -11,30 +11,19 @@
 
 namespace Liip\ImagineBundle\Binary\Loader;
 
-use Doctrine\Common\Persistence\ObjectManager as LegacyObjectManager;
 use Doctrine\Persistence\ObjectManager;
 use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
 
 abstract class AbstractDoctrineLoader implements LoaderInterface
 {
-    /**
-     * @var ObjectManager|LegacyObjectManager
-     */
-    protected $manager;
+    protected ObjectManager $manager;
 
-    /**
-     * @var string
-     */
-    protected $class;
+    protected string $modelClass;
 
-    /**
-     * @param ObjectManager|LegacyObjectManager $manager
-     * @param string                            $class
-     */
-    public function __construct($manager, $class = null)
+    public function __construct(ObjectManager $manager, string $modelClass)
     {
         $this->manager = $manager;
-        $this->class = $class;
+        $this->modelClass = $modelClass;
     }
 
     /**
@@ -42,14 +31,14 @@ abstract class AbstractDoctrineLoader implements LoaderInterface
      */
     public function find($path)
     {
-        $image = $this->manager->find($this->class, $this->mapPathToId($path));
+        $image = $this->manager->find($this->modelClass, $this->mapPathToId($path));
 
         if (!$image) {
             // try to find the image without extension
             $info = pathinfo($path);
             $name = $info['dirname'].'/'.$info['filename'];
 
-            $image = $this->manager->find($this->class, $this->mapPathToId($name));
+            $image = $this->manager->find($this->modelClass, $this->mapPathToId($name));
         }
 
         if (!$image) {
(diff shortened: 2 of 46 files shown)


src/Binary/Loader/AbstractDoctrineLoader.php
```diff
@@ -22,16 +22,12 @@ abstract class AbstractDoctrineLoader implements LoaderInterface
      */
     protected $manager;
```
dbu: kinda out of scope for this PR, but could we get rid of the LegacyObjectManager?

src/Binary/Loader/AbstractDoctrineLoader.php
```diff
@@ -22,16 +22,12 @@ abstract class AbstractDoctrineLoader implements LoaderInterface
...
-     */
-    protected $class;
+    protected string $class;
 
     /**
      * @param ObjectManager|LegacyObjectManager $manager
-     * @param string                            $class
      */
-    public function __construct($manager, $class = null)
+    public function __construct($manager, string $class)
```
dbu: this parameter used to be optional. would the code later fail if we pass null? and what class is this? if it is the class of the entity, we could name the parameter $entityClass?

src/Config/Filter/Type/Background.php
```diff
@@ -20,35 +20,23 @@ final class Background extends FilterAbstract
...
      * @param string|null $transparency possible values 0..100
      * @param string|null $position     position of the input image on the newly created background image. Valid values: topleft, top, topright, left, center, right, bottomleft, bottom, and bottomright
      */
     public function __construct(
-        string $color = null,
-        string $transparency = null,
-        string $position = null,
+        ?string $color,
+        ?string $transparency,
+        ?string $position,
```
dbu, line 39: i think we should keep defaulting the constructor arguments to null, to be consistent with the other config models

tests/Imagine/Data/DataManagerTest.php
```diff
@@ -113,6 +113,10 @@ public function testThrowsIfMimeTypeWasNotGuessedOnFind(): void
 
         $dataManager = new DataManager($mimeTypeGuesser, $this->createExtensionGuesserInterfaceMock(), $config);
         $dataManager->addLoader('the_loader', $loader);
+
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessage('The mime type of image cats.jpeg was not guessed');
```
dbu, line 118: good catch!

dbu (approved):
thanks! some minor details:

src/Binary/BinaryInterface.php
```diff
@@ -13,18 +13,9 @@
...
-     * @return string
-     */
-    public function getContent();
+    public function getContent(): string;
 
-    /**
-     * @return string
-     */
-    public function getMimeType();
+    public function getMimeType(): ?string;
```
franmomu, line 18: I've left this as nullable because this code:

https://github.com/liip/LiipImagineBundle/blob/8a57c556227cc9854ac7334ae28a3c752c8fef3d/Tests/Imagine/Data/DataManagerTest.php#L174

and 

https://github.com/liip/LiipImagineBundle/blob/8a57c556227cc9854ac7334ae28a3c752c8fef3d/Imagine/Data/DataManager.php#L131-L143

In the last case, we can move the exception right after:

`$mimeType = $this->mimeTypeGuesser->guess($binary);`

I'll add an extra commit doing this and then we decide what to do

dbu: ah, i was wondering why that is nullable.

i think its better to have an exception than having to check the return value for null. then if it is acceptable that the mime type is unknown, the exception can be caught, but the user of the interface does not stumble accidentally over the null. i wonder if we should have a specific exception for this, instead of the generic LogicException. it is not really a "logic" problem imho.

if we change this, we need to add a note to the upgrade guide, i think.

franmomu: I think I'll open an issue about this and leave it as nullable for now so we can finish the static typing.

src/Binary/Loader/AbstractDoctrineLoader.php
franmomu: I've changed it based on the `ObjectManager::find` call:

https://github.com/doctrine/persistence/blob/1ed5427e2ea50adb17aa6bb40037a8aa6e99a9fe/lib/Doctrine/Persistence/ObjectManager.php#L11-L27

It could be also a document I guess, I've changed to $modelClass

src/Config/Filter/Type/Background.php
franmomu: I removed them because the last parameter is not nullable, I didn't thought about that it could make sense using named parameters.

franmomu: Hmm in PHP 8.1 it will throw a deprecation warning: https://3v4l.org/DKJc7 (I'm creating a PR testing with 8.1)

dbu: interesting. not sure how they thought named parameters vs this new warning should work...

from the way how we use those config models, does it make sense that the fields are nullable? especially color... i assume that the filter defines the defaults - maybe we could move those to this class, or share constants with the default values to make it explicit? that would not fix the warning of php 8.1 though. 

reordering the parameters might make sense actually, with Size being a different type than the strings users would notice an error if they use the config class themselves and not silently swap strings and get unexpected results.

PR #1426, opened 2021-11-11
Test PHP 8.1 on CI

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc |

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.4', '8.0']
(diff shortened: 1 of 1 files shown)


dbu:
thanks, good point.
can you please do this against the 2.x branch? i think we want to know if we are compatible with 8.1. if  there is anything to adjust, i expect it to be minor and hope we can do a patch release in 2.x so people who move fast are not blocked.

franmomu:
Sure, I've done https://github.com/liip/LiipImagineBundle/pull/1427

PR #1430, opened 2021-11-11
Allow Symfony 6.0

franmomu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

I'll take a look later at the errors.


dbu:
if we do this in version 3 only, it will create quite the pressure to quickly release 3.0 :-)

i wonder how difficult it would be to make it work in the 2.x branch. i might try once you found how to solve the build errors.

franmomu:
This is the easiest way I've found to fix the build, the problem was some service to not be available through the container in tests.

dbu:
> This is the easiest way I've found to fix the build, the problem was some service to not be available through the container in tests.

i like this approach.

now there is a problem with a method signature that got a type declaration in symfony 6.

franmomu:
Closing in favor of https://github.com/liip/LiipImagineBundle/pull/1431

PR #1434, opened 2021-11-23
Refactor exception handling of `Binary\Loader\ChainLoader` class

robfrawley:
| Q | A
| --- | ---
| Branch? | 3.x 
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| References | https://github.com/liip/LiipImagineBundle/pull/1432#discussion_r751265173 / https://github.com/liip/LiipImagineBundle/pull/1432#discussion_r753720665
| License | MIT
| Doc | n/a

This is a __work in progress__ toward refactoring the error handling logic of [`Liip\ImagineBundle\Binary\Loader\ChainLoader`](https://github.com/liip/LiipImagineBundle/blob/3.x/src/Binary/Loader/ChainLoader.php#L1) as was originally discussed in code review comments from #1432; see the discussion beginning with *[dbu's #1432 comment](https://github.com/liip/LiipImagineBundle/pull/1432#discussion_r751265173)* and, more specifically, *[my (robfrawley) #1432 comment](https://github.com/liip/LiipImagineBundle/pull/1432#discussion_r753720665)* in response. 

Additional information will be added *as my schedule permits* and *during this PR's progression*.

For the time being, I would *greatly appreciate* any thoughts from @dbu, @franmomu, or any other maintainer/contributor/user as to the current state of this pull request as well as any alternate possible directions to investigate.

### DIFF
--- a/src/Binary/Loader/ChainLoader.php
+++ b/src/Binary/Loader/ChainLoader.php
@@ -11,17 +11,19 @@
 
 namespace Liip\ImagineBundle\Binary\Loader;
 
+use Liip\ImagineBundle\Exception\Binary\Loader\ChainAttemptNotLoadableException;
+use Liip\ImagineBundle\Exception\Binary\Loader\ChainNotLoadableException;
 use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
 
 class ChainLoader implements LoaderInterface
 {
     /**
-     * @var LoaderInterface[]
+     * @var array<string, LoaderInterface>
      */
     private array $loaders;
 
     /**
-     * @param LoaderInterface[] $loaders
+     * @param array<string, LoaderInterface> $loaders
      */
     public function __construct(array $loaders)
     {
@@ -37,36 +39,14 @@ class ChainLoader implements LoaderInterface
     {
         $exceptions = [];
 
-        foreach ($this->loaders as $loader) {
+        foreach ($this->loaders as $configName => $loader) {
             try {
                 return $loader->find($path);
-            } catch (\Exception $e) {
-                $exceptions[$e->getMessage()] = $loader;
+            } catch (NotLoadableException $loaderException) {
+                $exceptions[] = new ChainAttemptNotLoadableException($configName, $loader, $loaderException);
             }
         }
 
-        throw new NotLoadableException(self::getLoaderExceptionMessage($path, $exceptions, $this->loaders));
-    }
-
-    /**
-     * @param array<string, LoaderInterface> $exceptions
-     * @param array<string, LoaderInterface> $loaders
-     */
-    private static function getLoaderExceptionMessage(string $path, array $exceptions, array $loaders): string
-    {
-        $loaderMessages = array_map(static function (string $name, LoaderInterface $loader) {
-            return sprintf('%s=[%s]', (new \ReflectionObject($loader))->getShortName(), $name);
-        }, array_keys($loaders), $loaders);
-
-        $exceptionMessages = array_map(static function (string $message, LoaderInterface $loader) {
-            return sprintf('%s=[%s]', (new \ReflectionObject($loader))->getShortName(), $message);
-        }, array_keys($exceptions), $exceptions);
-
-        return vsprintf('Source image not resolvable "%s" using "%s" %d loaders (internal exceptions: %s).', [
-            $path,
-            implode(', ', $loaderMessages),
-            \count($loaders),
-            implode(', ', $exceptionMessages),
-        ]);
+        throw new ChainNotLoadableException($path, ...$exceptions);
     }
 }
--- a/src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
+++ b/src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Exception\Binary\Loader;
+
+use Liip\ImagineBundle\Binary\Loader\LoaderInterface;
+
+final class ChainAttemptNotLoadableException extends NotLoadableException
+{
+    private string          $configName;
+    private LoaderInterface $loaderInst;
+
+    public function __construct(string $configName, LoaderInterface $loaderInst, NotLoadableException $loaderException)
+    {
+        $this->configName = $configName;
+        $this->loaderInst = $loaderInst;
+
+        parent::__construct($this->compileFailureText(), 0, $loaderException);
+    }
+
+    public function getLoaderConfigName(): string
+    {
+        return $this->configName;
+    }
+
+    public function getLoaderObjectInst(): LoaderInterface
+    {
+        return $this->loaderInst;
+    }
+
+    public function getLoaderObjectName(): string
+    {
+        return (new \ReflectionObject($this->getLoaderObjectInst()))->getShortName();
+    }
+
+    public function getLoaderPriorError(): string
+    {
+        return $this->getPrevious()->getMessage();
+    }
+
+    private function compileFailureText(): string
+    {
+        return sprintf('%s=[%s]', $this->getLoaderObjectName(), $this->getLoaderConfigName());
+    }
+}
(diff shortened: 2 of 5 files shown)


src/Binary/Loader/ChainLoader.php
```diff
@@ -37,36 +39,14 @@ public function find($path)
     {
         $exceptions = [];
 
-        foreach ($this->loaders as $loader) {
+        foreach ($this->loaders as $configName => $objectInst) {
```
franmomu:
```suggestion
        foreach ($this->loaders as $configName => $loader) {
```
Just a suggestion that keeping `$loader` as variable seems more understandable to me.

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Liip\ImagineBundle\Exception\Binary\Loader;
+
+use Liip\ImagineBundle\Binary\Loader\LoaderInterface;
+
+class ChainAttemptNotLoadableException extends NotLoadableException
+{
+    private string $configName;
+    private LoaderInterface $objectInst;
+
+    public function __construct(string $configName, LoaderInterface $objectInst, NotLoadableException $loaderException)
```
franmomu:
```suggestion
    public function __construct(string $configName, LoaderInterface $loader, NotLoadableException $loaderException)
```
same here

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Exception\Binary\Loader;
+
+use Liip\ImagineBundle\Binary\Loader\LoaderInterface;
+
+class ChainAttemptNotLoadableException extends NotLoadableException
```
franmomu:
```suggestion
final class ChainAttemptNotLoadableException extends NotLoadableException
```
I think we should make them `final` by default if it is not designed to be extended.

franmomu (commented):
Nice refactor!

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
dbu: i agree, makes it much more explicit to know for backwards compatibility

dbu:
i like this refactoring, thanks!

i agree with the comment from @franmomu about naming the loader variable $loader and making things final where possible.

src/Binary/Loader/ChainLoader.php
robfrawley: Sounds good; done.

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
robfrawley: Excellent idea; done.

src/Binary/Loader/ChainLoader.php
robfrawley: Sounds good; done.

*(Note: I do agree that the use of `loader` (especially for user-facing code) is much clearer using the suggested (and original) variable name (instead of `objectInst`); in fact, I never actually intended to change that variable at all! ;-) I think the loader variable was altered when I was refactoring the new exception class properties and accidentally cast too wide a "find/replace" net, such that the ChainLoader::$loader property was mistakenly included in the change. Regardless, it's fixed now. I was working on a machine without my normal environment, PHPStorm, and was instead using SublimeText, so such unexpected changes shouldn't normally be an issue. Thanks for the catch, though.)*

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
robfrawley: @franmomu @dbu Are you both comfortable with changing `objectInst` to `loaderInst` (instead of the suggested variable name of `loader`) with regard to the *exception class __only__*? Specifically:

- `ChainAttemptNotLoadableException::$objectInst` *->* __the original property name__
- `ChainAttemptNotLoadableException::$loaderInst` *->* __my suggested property name__

This better aligns with the methods as they are currently defined for the chain-attempt exception, IMHO. For example:

- `ChainAttemptNotLoadableException::getLoaderConfigName()`
- `ChainAttemptNotLoadableException::getLoaderObjectInst()`
- `ChainAttemptNotLoadableException::getLoaderObjectName()`

*(Not that it would matter to anyone else, but my suggestion also better satisfies my OCD which prefers same-length groups of properties and methods, when feasible to do so without convoluting their meaning...)*

Let me know if this is objectionable to either of you; I'll perform the original change if you both are not agreeable to my suggested alternative.

robfrawley:
I've gone ahead and squashed the five commits created while working toward this PR, as well as rebased my changes over the latest `3.x` branch. I think this should be just about ready, pending the outcome of https://github.com/liip/LiipImagineBundle/pull/1434#discussion_r757095083 and any additional thoughts that anyone may have after performing any final reviews. As such, I've removed the `[WIP]` label from the PR title and I request a final quick check from any interested maintainers and contributors.

*(At this point, I'll squash and rebase every time any additional changes are required so the PR is always in a clean and easily mergeable state; hopefully such doesn't happen too many times, though :-))*

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return $this->configName;
+    }
+
+    public function getLoaderObjectInst(): LoaderInterface
+    {
+        return $this->loaderInst;
+    }
+
+    public function getLoaderObjectName(): string
```
dbu: if i understand correctly, this is the class name without the namespace. how about calling the method `getLoaderClassName`?

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+
+        parent::__construct($this->compileFailureText(), 0, $loaderException);
+    }
+
+    public function getLoaderConfigName(): string
+    {
+        return $this->configName;
+    }
+
+    public function getLoaderObjectInst(): LoaderInterface
```
dbu: yeah naming things... i would call the thing just `$loader` and this method `getLoader`. i prefer concise names. that it is an instantiated object is defined by the return type declaration.

dbu (commented):
i would prefer to do some renaming here as i feel both "object" and "inst" do not help understanding the code, so just "loader" would be my preferred solution.

take it as a suggestion, if you really want the current name, we can also keep it.

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Liip\ImagineBundle\Exception\Binary\Loader;
+
+use Liip\ImagineBundle\Binary\Loader\LoaderInterface;
+
+final class ChainAttemptNotLoadableException extends NotLoadableException
+{
+    private string          $loaderIndex;
+    private LoaderInterface $loaderClass;
```
dbu, line 19: i would prefer to call this $loader. $loaderClass suggests its the class name of the loader, but it is the actual instance.

then again, looking at the rest of the code: we do not use the class anywhere afterwards. how about doing the reflection thing in the constructor and have this field be `private string $loaderClassName`?

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+
+        parent::__construct($this->compileMessageTxt(), 0, $loaderException);
+    }
+
+    public function getLoaderIndex(): string
+    {
+        return $this->loaderIndex;
+    }
+
+    public function getLoaderClass(): LoaderInterface
```
dbu, line 34: this is not called from outside the class, i think we can drop the method

src/Exception/Binary/Loader/ChainAttemptNotLoadableException.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return $this->loaderClass;
+    }
+
+    public function getLoaderClassName(): string
```
dbu, line 39: but this one should stay as is, it gives you the classname of the loader, so i like the name.

if we move reflection to the constructor, this will become a simple getter for `$this->loaderClassName`

dbu (commented):
sorry to continue to argue about the naming :see_no_evil:

i think we can further simplify the exception class and remove some methods we don't need.

PR #1437, opened 2021-12-01
[WIP] Change phpunit workflow to allow disabling specific test environments when incompatibilities are encountered.

robfrawley:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes/no
| Fixed tickets | https://github.com/liip/LiipImagineBundle/pull/1436#issuecomment-983360655
| License | MIT
| Doc | n/a

I am testing some different changes to resolve the above-titled issue; no single solution to the problem has been determined appropriate for this project yet. As such, __*do not merge!*__
__

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -102,7 +102,7 @@ jobs:
         run: vendor/bin/simple-phpunit -v
 
       - name: Install php-coveralls
-        run: composer require --no-interaction --no-progress php-coveralls/php-coveralls:^2.5
+        run: composer global require --no-interaction --no-progress php-coveralls/php-coveralls:^2.5
 
       - name: Upload coverage results to Coveralls
         uses: nick-invision/retry@v2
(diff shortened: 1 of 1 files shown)


robfrawley:
Looks like we need to wait on php-coveralls/php-coveralls#328 for a resolution to our failing workflows when running with any Symfony `6.x` release.

@dbu: For the time being, do you want me to update the workflow matrix to disable coveralls reporting on incompatible environments? If so, we could do that in such a way as to quickly be able to enable/disable certain environments easily, something that will likely be useful again in the future when a different incompatibility pops up...

dbu:
thanks for looking into this!

how about disabling the coveralls step completely when php version is 8, until coveralls fixes their signature to be compatible with symfony? or even mark just 1 build in the matrix to actually do the coveralls things. i don't think we need coverage for each case.

.github/workflows/phpunit.yml
Montiwa11, line 1: [](url)

Montiwa11 (changes requested):
robfrawley:bugfix-allow-symfony-and-php8-compatibility-with-php-coveralls

dbu:
@robfrawley is this pull request still relevant? the only change in it is locking the coveralls version

Issue #1415, opened 2021-11-04
type declarations


dbu:
switch all classes and interfaces to return type declarations over `@return` and parameter&property type declarations.

this means bumping the minimum PHP version to 7.4. imo this is worth it.

at the same time, make everything we can final and methods private over protected, unless they are explicitly meant as extension points.

stephanvierkant:
I'll work on this! I'll use Rector for this. Should I add the config to the repository or not?

dbu:
thanks for offering to help!

to do the cleanup changes, rector sounds like the right tool. about adding the configuration: could we enable some rules in php-cs-fixer to detect incomplete property types rather than adding another tool into the CI?

tacman:
I'd like to see the rector file in the repo, as it's not part of the CI.  It will make it easier as rector as more rules for us to adjust to what we use.  Especially as we drop support for older versions as they reach EOL.

PR #1439, opened 2021-12-07
Added return types

stephanvierkant:
| Q | A
| --- | ---
| Branch? | 3.x 
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | #1415
| License | MIT
| Doc | 

Added or changed some return types. Most of the methods already had a return type. I couldn't find any methods where a parameter type declaration could be added.

### DIFF
--- a/src/Events/CacheResolveEvent.php
+++ b/src/Events/CacheResolveEvent.php
@@ -15,24 +15,12 @@ use Symfony\Contracts\EventDispatcher\Event;
 
 class CacheResolveEvent extends Event
 {
-    /**
-     * Resource path.
-     */
     protected string $path;
 
-    /**
-     * Filter name.
-     */
     protected string $filter;
 
-    /**
-     * Resource url.
-     */
     protected ?string $url;
 
-    /**
-     * Init default event state.
-     */
     public function __construct(string $path, string $filter, ?string $url = null)
     {
         $this->path = $path;
@@ -40,50 +28,32 @@ class CacheResolveEvent extends Event
         $this->url = $url;
     }
 
-    /**
-     * Sets resource path.
-     */
     public function setPath(string $path): void
     {
         $this->path = $path;
     }
 
-    /**
-     * Returns resource path.
-     */
     public function getPath(): string
     {
         return $this->path;
     }
 
-    /**
-     * Sets filter name.
-     */
     public function setFilter(string $filter): void
     {
         $this->filter = $filter;
     }
 
-    /**
-     * Returns filter name.
-     */
     public function getFilter(): string
     {
         return $this->filter;
     }
 
-    /**
-     * Sets resource url.
-     */
     public function setUrl(?string $url): void
     {
         $this->url = $url;
     }
 
-    /**
-     * Returns resource url.
-     */
-    public function getUrl(): ?string
+    public function getUrl(): string
     {
         return $this->url;
     }
--- a/src/Imagine/Cache/Resolver/ProxyResolver.php
+++ b/src/Imagine/Cache/Resolver/ProxyResolver.php
@@ -56,7 +56,7 @@ class ProxyResolver implements ResolverInterface
         $this->resolver->remove($paths, $filters);
     }
 
-    protected function rewriteUrl(string $url): string
+    protected function rewriteUrl(string $url): ?string
     {
         if (empty($this->hosts)) {
             return $url;
--- a/src/Binary/Loader/FileSystemLoader.php
+++ b/src/Binary/Loader/FileSystemLoader.php
@@ -37,7 +37,7 @@ class FileSystemLoader implements LoaderInterface
     /**
      * {@inheritdoc}
      */
-    public function find($path)
+    public function find($path): FileBinary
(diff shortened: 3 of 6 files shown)


src/Imagine/Cache/CacheManager.php
```diff
@@ -145,9 +145,9 @@ public function isStored(string $path, string $filter, ?string $resolver = null)
      *
      * @throws NotFoundHttpException if the path can not be resolved
      *
-     * @return string The url of resolved image
+     * @return string|null The url of resolved image
```
dbu: is the reason that this is nullable the signature of $postEvent->getUrl()? maybe it would be better to make that one have to return a string.

it seems better to me to either return an url or throw an exception, rather than the caller having to verify if it got back an url.

src/Imagine/Filter/FilterManager.php
```diff
@@ -79,7 +79,7 @@ public function apply(BinaryInterface $binary, array $config): BinaryInterface
         return $this->applyPostProcessors($this->applyFilters($binary, $config), $config);
     }
 
-    public function applyFilters(BinaryInterface $binary, array $config): BinaryInterface
+    public function applyFilters(BinaryInterface $binary, array $config): Binary
```
dbu: i'd rather keep this at the BinaryInterface. consumers should not rely on the specific implementation, to stay flexible.

if for some reason we would later want to have other binary implementations (e.g. for streaming, or whatever) it would be a BC break to change. if we only declare the interface to be returned, we are more flexible.

src/Service/FilterService.php
```diff
@@ -92,7 +92,7 @@ public function warmUpCache(
         return $warmedUp;
     }
 
-    public function getUrlOfFilteredImage(string $path, string $filter, ?string $resolver = null, bool $webpSupported = false): string
+    public function getUrlOfFilteredImage(string $path, string $filter, ?string $resolver = null, bool $webpSupported = false): ?string
```
dbu: can we avoid this being nullable?

dbu (commented):
thanks for the pull request.

as we are doing a new major version, i think that we should also change the LoaderInterface to have typed parameters (and then adjust all implementations accordingly)

i want to challenge if we need to make return types nullable - i would prefer to go the other direction of making more things not nullable.

src/Imagine/Cache/CacheManager.php
```diff
@@ -77,7 +77,7 @@ public function addResolver(string $filter, ResolverInterface $resolver): void
      *
      * @param string $path The path where the resolved file is expected
      */
-    public function getBrowserPath(string $path, string $filter, array $runtimeConfig = [], ?string $resolver = null, int $referenceType = UrlGeneratorInterface::ABSOLUTE_URL): string
+    public function getBrowserPath(string $path, string $filter, array $runtimeConfig = [], ?string $resolver = null, int $referenceType = UrlGeneratorInterface::ABSOLUTE_URL): ?string
```
dbu: do we really need to make this return nullable?

dbu:
@stephanvierkant do you plan to follow up on the feedback? i would love to get these improvements in once we figured out the remaining questions.

src/Imagine/Cache/Resolver/ProxyResolver.php
```diff
@@ -56,7 +56,7 @@ public function remove(array $paths, array $filters): void
         $this->resolver->remove($paths, $filters);
     }
 
-    protected function rewriteUrl(string $url): string
+    protected function rewriteUrl(string $url): ?string
```
dbu, line 59: the only way how this can return null is when there is a regex error in the `preg_replace` call. is the calling side of this handling that? if not, i'd prefer to check the return value of preg_replace in this method and use preg_last_error_msg to throw an exception. then we can leave this as returning a non-nullable string.

src/Events/CacheResolveEvent.php
```diff
@@ -15,75 +15,45 @@
...
     public function setUrl(?string $url): void
     {
         $this->url = $url;
     }
 
-    /**
-     * Returns resource url.
-     */
-    public function getUrl(): ?string
+    public function getUrl(): string
```
dbu, line 56: i don't think you can change this to non-nullable as the url property of the class is nullable.

cbeyer:
Any news ?

src/Events/CacheResolveEvent.php
stephanvierkant: I've just looked at ‘protected ?string $url;’

dbu: the correct return type should look like this:

```suggestion
    public function getUrl(): ?string
```

dbu:
i would love to get this moving again, but have little time. @cbeyer if you want to pick up the branch from stephan and do a new pull request with the conflicts resolved and possibly more strict typing cleanup, please do and i will be happy to review the changes.

dbu:
@tacman this would be a crucial bit to modernize the code for 3.x

PR #1440, opened 2021-12-09
Gif should not be transformed to Webp format

Yoann-TYT:
| Q | A
| --- | ---
| Branch? | 2.x or 3.x 
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

When we configure the use of webp

```yaml
...
    webp:
        generate: true
...
```

and we try to generate image from a gif, we get this error

`Animated picture is currently only supported on gif`

Now, it's possible to generate animated webp. I will propose an evolution on `imagine/imagine` but for the moment, this PR fix the bug. 

Thanks a lot for your review !


dbu (commented):
thanks for the pull request.

i agree with the aim of not trying to do webp for an animated gif if we don't correctly support that conversion. however, this implementation will not use webp for any gif, also non-animated ones.

is there a way to detect if the gif is animated or not? it feels a bit iffy to rely on the file extension only, but looking at the bundle code, it seems that is common practice in other places as well.
if we can't ask if the gif is animated, could we fallback to not create webp when the exception you encountered is thrown?

Yoann-TYT:
Thanks for your response !

I agree with you.. A gif could be static. If we check the code, we consider all gifs are animated. 

https://github.com/liip/LiipImagineBundle/blob/25098c758380933ba4f2d4ffc68614c4c1c90a72/Imagine/Filter/FilterManager.php#L158

To avoid that, we can firstly check the driver used. 
If it's "gd", we can determine that isn't animated because this driver can't manage animated images. 
If it's gmagick or imagick, we can use deconstructimages method to count frames. If we get more than one, it's a animated gif, else not.
I think it's possible, but I'm afraid about performance :-/ 

> if we can't ask if the gif is animated, could we fallback to not create webp when the exception you encountered is thrown?

Yes, I think we can do that in FilterManager.

I prefer the second option not because is easiest but better for performance. :) 

What do you prefer, the first or the second ? 

Thanks !

dbu:
thanks for checking those options. 

pre-loading the image to check for frames seems like it can incur quite some overhead when the image file is large, even more if the file is stored in a remote filesystem.

i prefer option two here, fallback to not convert to webp if the conversion runs into an issue. (that actually seems like it would have been a good idea in general when we introduced webp - though by now the converter is probably quite stable). can we do that without having to redo all the operations on the image and just save the result in a different format?

this will also lead to less changes if you can actually make animated webp work somehow.

dbu:
@Yoann-TYT do you have time to look into my feedback?

Yoann-TYT:
@dbu I will do that this week !

Imagine/Filter/FilterManager.php
```diff
@@ -160,7 +160,16 @@ private function exportConfiguredImageBinary(BinaryInterface $binary, ImageInter
         }
 
         $filteredFormat = $config['format'] ?? $binary->getFormat();
-        $filteredString = $image->get($filteredFormat, $options);
+        try {
+            $filteredString = $image->get($filteredFormat, $options);
+        } catch (\Exception $exception) {
+            if ('webp' === $filteredFormat && true === $options['animated']) {
```
dbu: i think we need to check if the option is set as well:
```suggestion
            if ('webp' === $filteredFormat && array_key_exists('animated', $options) && true === $options['animated']) {
```

dbu:
thanks a lot. i squashed the commits and reversed the code to avoid the `else`. i will merge #1444 if the CI agrees with my refactoring.

dbu:
thank you for the improvement. tagged as https://github.com/liip/LiipImagineBundle/releases/tag/2.7.5

PR #1444, opened 2022-01-11
animated gifs can not be transformed to the webp format

dbu:
We can't efficiently check if the input is an animated gif, therefore we
catch the exception and try to convert to gif if converting to webp
failed.

continue fix #1440

| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| License | MIT
| Doc | -

When we configure the use of webp

```yaml
...
    webp:
        generate: true
...
```

and we try to generate image from a gif, we get this error

`Animated picture is currently only supported on gif`

Now, it's possible to generate animated webp. I will propose an evolution on `imagine/imagine` but for the moment, this PR fix the bug. 

Thanks a lot for your review !

### DIFF
--- a/Imagine/Filter/FilterManager.php
+++ b/Imagine/Filter/FilterManager.php
@@ -160,7 +160,17 @@ class FilterManager
         }
 
         $filteredFormat = $config['format'] ?? $binary->getFormat();
-        $filteredString = $image->get($filteredFormat, $options);
+        try {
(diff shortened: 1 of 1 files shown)


dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.7.5

Issue #1451, opened 2022-02-09
Replace guzzle with symfony http client in 3.0


tacman:
**Is your feature request related to a problem? Please describe.**

Not really.  There's a deprecation error: Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.

**Describe the solution you'd like**

I've found the Symfony HttpClient to be as good as Guzzle, and has fewer issues when trying to upgrade to newer Symfony versions.

dbu:
this being a symfony bundle, this makes sense to me, yes. in framework-independant libraries i advocate for PSR-18, but this is a bundle that needs symfony anyways, so we can just use the client they provide.

do you want to work on that?

tacman:
Yes, I'll give it a shot.

tacman:
So it looks we're only using Guzzle in the tests, and even then, it's just to have access to an ObjectMock of type Model (which extends Guzzle\Collection).

So we can drop Guzzle altogether, once we have a replacement for the Mock Object.  Alas, this is something I don't have much experience in (but need to learn anyway).  Can you point me to a replacement?

The code in question is in AwsS3ResolverTest.


```php
    /**
     * @return MockObject&Model
     */
    protected function getS3ResponseMock()
    {
        return $this->createObjectMock(Model::class);
    }
```

dbu:
oh, so actually it is the Aws library that is hard-tied to guzzle. however, i had a look at the AwsS3Resolver and we never use the return value that we mock in these places. i think we can get rid of that model (i think #1455 would have created problems if it was actually used as intended) and of having a mock response. 

lets completely remove the `willReturn` section for the deleteObject and putObject calls in the AwsS3ResolverTest. as long as AWS does not do strict typing, that shoud be good enough for the test.

tacman:
I think the problem was in an autocomplete -- that is, Model::class used the Guzzle Model, not our Model.  That's how version 2 works, I'm trying that fix now.

PR #1456, opened 2022-02-09
use our Model, not Guzzle, for tests

tacman:
| Q | A
| --- | ---
| Branch? | 3
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1451 
| License | MIT

This removes guzzle references from our code, which weren't being used anyway.

### DIFF
--- a/tests/Imagine/Cache/Resolver/AwsS3ResolverTest.php
+++ b/tests/Imagine/Cache/Resolver/AwsS3ResolverTest.php
@@ -12,11 +12,11 @@
 namespace Liip\ImagineBundle\Tests\Imagine\Cache\Resolver;
 
 use Aws\S3\S3Client;
-use Guzzle\Service\Resource\Model;
 use Liip\ImagineBundle\Imagine\Cache\Resolver\AwsS3Resolver;
 use Liip\ImagineBundle\Imagine\Cache\Resolver\ResolverInterface;
 use Liip\ImagineBundle\Model\Binary;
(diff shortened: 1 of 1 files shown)


tests/Imagine/Cache/Resolver/AwsS3ResolverTest.php
```diff
@@ -12,11 +12,11 @@
 namespace Liip\ImagineBundle\Tests\Imagine\Cache\Resolver;
 
 use Aws\S3\S3Client;
-use Guzzle\Service\Resource\Model;
 use Liip\ImagineBundle\Imagine\Cache\Resolver\AwsS3Resolver;
 use Liip\ImagineBundle\Imagine\Cache\Resolver\ResolverInterface;
 use Liip\ImagineBundle\Model\Binary;
 use Liip\ImagineBundle\Tests\AbstractTest;
+use Liip\ImagineBundle\Tests\Fixtures\Model;
```
dbu, line 19: if the aws client ever starts to declare return types, this will not match the return type, right? i'd hope they only do that in a new major version so i don't expect it to happen soon or uncontrolled.

but as we don't use the return type, can't we simply remove mocking the return from the S3Client in this test?

tacman:
I think it's probably safer to remove mocking the return, yes.  I'm still
learning about testing, and I find mocking the hardest part.  Can you use
my changes and tweak as you see fit?

Issue #1457, opened 2022-02-09
Drop support for AWS Version 1 cache resolver in v3?


tacman:
Installing (composer install) shows this warning:

       Package amazonwebservices/aws-sdk-for-php is abandoned, you should avoid using it. Use aws/aws-sdk-php instead.

tacman:
That warning exists only because of support for AWS Version 1 cache resolver.  So this is more complicated than it looked at first glance.  I'll keep poking at it.

dbu:
i am not super familiar with AWS. can people use AWS with a newer version in this bundle and keep using version 1 for other things, or is that somehow a switch they have to flip? if users can mix api versions i see no problem in us dropping version 1 support in version 3.

dbu:
there is additionally a phpstan warning that we have too many arguments. i ignore that for now but i think we should use 3.x to drop legacy things.

PR #1458, opened 2022-02-09
Update dependencies, including dev-develop for imagine

tacman:
| Q | A
| --- | ---
| Branch? | 3
| Bug fix? | no
| New feature? | no
| BC breaks? | maybe
| Deprecations? | yes!  All the imagine ones.
| Fixed tickets | #1457 
| License | MIT

Since this bundle is a wrapper around php-imagine, and version 3 is our dev version, I've updated the require-dev components to allow the latest versions of everything.  By using the dev-develop version of imagine, all the deprecation notices from that library disappeared.

Tests pass.

However, I don't know how stable dev-develop is.  So should v3 of this bundle support dev-develop?  Maybe someone from the imagine library can jump in.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         }
     },
     "require": {
-        "php": "^8.0",
+        "php": "^7.4|^8.0",
         "ext-mbstring": "*",
         "imagine/imagine": "dev-develop",
         "symfony/filesystem": "^4.4|^5.3|^6.0",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require": {
         "php": "^8.0",
         "ext-mbstring": "*",
-        "imagine/imagine": "^1.2.4",
+        "imagine/imagine": "dev-develop",
         "symfony/filesystem": "^4.4|^5.3|^6.0",
         "symfony/finder": "^4.4|^5.3|^6.0",
         "symfony/framework-bundle": "^3.4.23|^4.4|^5.3|^6.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -24,7 +24,7 @@
     "require": {
         "php": "^7.4|^8.0",
         "ext-mbstring": "*",
-        "imagine/imagine": "^1.2.4",
+        "imagine/imagine": "dev-develop",
```
dbu, line 27: i would prefer to only use stable versions of upstream dependencies.

but happy to raise the version to the latest released major version.

dbu (commented):
thanks!

i am -1 on using a develop version of imagine, the rest seems good to me. allowing the new flysystem and log could maybe even be done in the 2.x branch as well, i don't think there are changes for consumers, only for implementers.

tacman:
Makes sense.  They've fixed the deprecation warnings in dev-develop, so
we'll just get them in our v3 until they've released, then we can make the
same adjustments.

I'll be happier when the newer versions are out and working in harmony, as
I'm a bit confused with how to get Symfony 5.4/6, Flysystem,
flysystem-bundle, and liip-imagine all working together with their latest
production releases.  I used to use SymfonyCast All About Uploading, but it
has too many outdated dependencies.

Can you point me to a demo repo that includes those 4 components and
works?  I think a lot of us have basic and similar needs -- let users
upload an image (to a local dir in dev, an s3 bucket in prod), then
dynamically show different sizes of it.

Thx.

dbu:
i am not aware of a current demo repository.

the (released from the 2.x branch) LiipImagineBundle has tests running with symfony 6 though, so i would expect it to work with symfony 6. 

getting the imagine-bundle 3 released still has a bunch of tasks that need to be done, don't hold your breath for that to happen too soon ;-)

PR #1466, opened 2022-05-19
feat: add a cwebp post-processor for WebP images

DjLeChuck:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Doc | 

Add a new `cwebp` post-processor for filters. It use [`cwebp`](https://developers.google.com/speed/webp/docs/cwebp) to optimize the input `webp` file.

Available options are: 
* q
  * Specify the compression factor for RGB channels between **0** and **100**. The default is **75**.
  * In case of lossy compression , a small factor produces a smaller file with lower quality. Best quality is achieved by using a value of **100**.
  * In case of lossless compression (specified by the **-lossless** option), a small factor enables faster compression speed, but produces a larger file. Maximum compression is achieved by using a value of **100**.
* alphaQ
  * Specify the compression factor for alpha compression between **0** and **100**. Lossless compression of alpha is achieved using a value of **100**, while the lower values result in a lossy compression.
* m
  * Specify the compression method to use. This parameter controls the trade off between encoding speed and the compressed file size and quality. Possible values range from **0** to **6**. When higher values are used, the encoder will spend more time inspecting additional encoding possibilities and decide on the quality gain. Lower value can result in faster processing time at the expense of larger file size and lower compression quality.
* alphaFilter
  * Specify the predictive filtering method for the alpha plane. One of **none**, **fast** or **best**, in increasing complexity and slowness order. Internally, alpha filtering is performed using four possible predictions (none, horizontal, vertical, gradient). The **best** mode will try each mode in turn and pick the one which gives the smaller size. The **fast** mode will just try to form an a priori guess without testing all modes.
* alphaMethod
  * Specify the algorithm used for alpha compression: **0** or **1**. Algorithm **0** denotes no compression, **1** uses WebP lossless format for compression.
* exact
  * Preserve RGB values in transparent area. The default is off, to help compressibility.
* metadata
  * A comma separated list of metadata to copy from the input to the output if present. Valid values: **all**, **none**, **exif**, **icc**, **xmp**.
  * Note that each input format may not support all combinations.

### DIFF
--- a/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
@@ -10,16 +10,6 @@ use Symfony\Component\Process\Exception\ProcessFailedException;
 class CwebpPostProcessor extends AbstractPostProcessor
 {
     /**
-     * Specify the level of near-lossless image preprocessing. This option adjusts pixel values to help
-     * compressibility, but has minimal impact on the visual quality. It triggers lossless compression mode
-     * automatically. The range is **0** (maximum preprocessing) to **100** (no preprocessing). The typical value is
-     * around **60**. Note that lossy with **-q 100** can at times yield better results.
-     *
-     * @var int
-     */
-    protected $nearLossless;
-
-    /**
      * Specify the compression factor for RGB channels between **0** and **100**. The default is **75**.
      *
      * In case of lossy compression , a small factor produces a smaller file with lower quality. Best quality is
@@ -90,7 +80,6 @@ class CwebpPostProcessor extends AbstractPostProcessor
     public function __construct(
         string $executablePath = '/usr/bin/cwebp',
         string $temporaryRootPath = null,
-        int $nearLossless = null,
         int $q = null,
         int $alphaQ = null,
         int $m = null,
@@ -101,7 +90,6 @@ class CwebpPostProcessor extends AbstractPostProcessor
     ) {
         parent::__construct($executablePath, $temporaryRootPath);
 
-        $this->nearLossless = $nearLossless;
         $this->q = $q;
         $this->alphaQ = $alphaQ;
         $this->m = $m;
@@ -152,15 +140,6 @@ class CwebpPostProcessor extends AbstractPostProcessor
     private function getProcessArguments(array $options = []): array
     {
         $arguments = [$this->executablePath];
-
-        if ($nearLossless = $options['nearLossless'] ?? $this->nearLossless) {
-            if (!\in_array($nearLossless, range(0, 100), true)) {
-                throw new InvalidOptionException('The "nearLossless" option must be an int between 0 and 100', $options);
-            }
-
-            $arguments[] = '-near_lossless';
-            $arguments[] = $nearLossless;
-        }
 
         if ($q = $options['q'] ?? $this->q) {
             if (!\in_array($q, range(0, 100), true)) {
--- a/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
 namespace Liip\ImagineBundle\Imagine\Filter\PostProcessor;
 
 use Liip\ImagineBundle\Binary\BinaryInterface;
@@ -17,7 +26,6 @@ class CwebpPostProcessor extends AbstractPostProcessor
      *
      * In case of lossless compression (specified by the **-lossless** option), a small factor enables faster
      * compression speed, but produces a larger file. Maximum compression is achieved by using a value of **100**.
-     *
      *
      * @var int
      */
@@ -151,8 +159,7 @@ class CwebpPostProcessor extends AbstractPostProcessor
                 }
 
                 return $value >= 0 && $value <= 100;
-            })
-        ;
+            });
 
         $resolver
             ->setDefault('alphaQ', $this->alphaQ)
@@ -163,8 +170,7 @@ class CwebpPostProcessor extends AbstractPostProcessor
                 }
 
                 return $value >= 0 && $value <= 100;
-            })
-        ;
+            });
 
         $resolver
             ->setDefault('m', $this->m)
@@ -175,25 +181,21 @@ class CwebpPostProcessor extends AbstractPostProcessor
                 }
 
                 return $value >= 0 && $value <= 6;
-            })
-        ;
+            });
 
         $resolver
             ->setDefault('alphaFilter', $this->alphaFilter)
             ->setAllowedTypes('alphaFilter', ['null', 'string'])
-            ->setAllowedValues('alphaFilter', [null, 'none', 'fast', 'best'])
-        ;
+            ->setAllowedValues('alphaFilter', [null, 'none', 'fast', 'best']);
 
         $resolver
             ->setDefault('alphaMethod', $this->alphaMethod)
             ->setAllowedTypes('alphaMethod', ['null', 'int'])
-            ->setAllowedValues('alphaMethod', [null, 0, 1])
-        ;
+            ->setAllowedValues('alphaMethod', [null, 0, 1]);
 
         $resolver
             ->setDefault('exact', $this->exact)
-            ->setAllowedTypes('exact', ['null', 'bool'])
-        ;
+            ->setAllowedTypes('exact', ['null', 'bool']);
 
         $resolver
             ->setDefault('metadata', $this->metadata)
@@ -210,8 +212,7 @@ class CwebpPostProcessor extends AbstractPostProcessor
                 }
 
                 return true;
-            })
-        ;
+            });
     }
 
     /**
--- a/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
+++ b/Imagine/Filter/PostProcessor/CwebpPostProcessor.php
@@ -68,12 +68,12 @@ class CwebpPostProcessor extends AbstractPostProcessor
     protected $exact;
 
     /**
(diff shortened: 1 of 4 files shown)


Imagine/Filter/PostProcessor/CwebpPostProcessor.php
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * A comma separated list of metadata to copy from the input to the output if present. Valid values: **all**,
+     * **none**, **exif**, **icc**, **xmp**.
+     *
+     * Note that each input format may not support all combinations.
+     *
+     * @var string
+     */
+    protected $metadata;
```
dbu, line 86: i would prefer to define this as an array of string. the comma separated list is the way of telling it to the cwebp binary, but not a good representation of the configuration in php.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * @param int|string[] $options
+     *
+     * @return string[]
+     */
+    private function getProcessArguments(array $options = []): array
+    {
+        $arguments = [$this->executablePath];
+
```
dbu, line 227: this method could use the symfony OptionsResolver class to be quite a bit more consise, i think. other processors had fewer options, so it showed less. do you want to try that?

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
```diff
@@ -0,0 +1,230 @@
...
+        if ($options['exact'] ?? $this->exact) {
+            $arguments[] = '-exact';
+        }
+
+        if ($metadata = $options['metadata'] ?? $this->metadata) {
+            $metadataValues = explode(',', $metadata);
+
+            foreach ($metadataValues as $metadataValue) {
+                if (!\in_array($metadataValue, ['all', 'none', 'exif', 'icc', 'xmp'], true)) {
```
dbu: what does the combination of all or none with any other value mean? should we detect those as well, or do we leave that to cwebp to report?

Resources/config/imagine.xml
```diff
@@ -424,5 +436,10 @@
             <tag name="liip_imagine.filter.post_processor" post_processor="mozjpeg" />
         </service>
 
+        <service id="liip_imagine.filter.post_processor.cwebp" class="Liip\ImagineBundle\Imagine\Filter\PostProcessor\CwebpPostProcessor">
+            <argument>%liip_imagine.cwebp.binary%</argument>
```
dbu, line 442: don't you need to list all parameters here to get them injected to the class?

Resources/config/imagine.xml
```diff
@@ -28,6 +28,18 @@
...
+
+        <parameter key="liip_imagine.cwebp.binary">/usr/bin/cwebp</parameter>
+        <parameter key="liip_imagine.cwebp.near_lossless">100</parameter>
+        <parameter key="liip_imagine.cwebp.q">75</parameter>
+        <parameter key="liip_imagine.cwebp.alpha_q">100</parameter>
+        <parameter key="liip_imagine.cwebp.m">4</parameter>
+        <parameter key="liip_imagine.cwebp.alpha_filter">fast</parameter>
+        <parameter key="liip_imagine.cwebp.alpha_method">1</parameter>
+        <parameter key="liip_imagine.cwebp.exact">false</parameter>
+        <parameter key="liip_imagine.cwebp.metadata">none</parameter>
```
dbu: this is consistent with the other post processors, so lets do it this way for now. but i think it would be much better if this was using the symfony DI configuration system... something for the next major version.

dbu (changes requested):
thanks for the contribution! i really like the documentation you added on the options :+1: 

can you please add a short documentation to https://github.com/liip/LiipImagineBundle/tree/2.x/Resources/doc/post-processors ? i would not duplicate all details that you have in the docblocks, but explain what the cwebp post processor is and the parameter names that can be tweaked.

dbu:
i fixed the cs problems that are not related to this PR. please rebase this branch on origin/2.x to get those fixes in.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: I don't know what to do about this. The `cwebp` documentation is not clear either: `Note that each input format may not support all combinations.`

Nothing is report, it just seems to be silently processed by the binary.

Resources/config/imagine.xml
DjLeChuck: Oops, my bad. I have fixed it.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: Done!

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: Yep, will definitely try it. It was my first idea, but I change my mind to be consistent with others processors.

Resources/config/imagine.xml
DjLeChuck: Yes, I'm not proud of this either. I was just doing same as others post-processors to remain consistent.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: Here we go, done!

DjLeChuck:
@dbu thank you for your review! 😄 

All of your requests are done, except for the metadata validation (see my comment).

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
```diff
@@ -0,0 +1,283 @@
...
+        int $alphaQ = null,
+        int $m = null,
+        string $alphaFilter = null,
+        int $alphaMethod = null,
+        bool $exact = null,
+        array $metadata = []
+    ) {
+        parent::__construct($executablePath, $temporaryRootPath);
+
+        $this->q = $q;
```
dbu, line 106: my idea was to create the options resolver with the arguments and then set the properties from the resolved options. that way, they are validated and we never assign something invalid to the object properties.

but i realize we need to validate twice, first for the constructor and then with the options specified in a call. what does optionsresolver do when one of the defaults does not match the allowed values? (the allowed types are ensured by the static types on the constructor) if it reports a problem, the current implementation seems good to me.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
```diff
@@ -0,0 +1,283 @@
...
+
+                return \in_array($value, ['none', 'fast', 'best'], true);
+            })
+        ;
+
+        $resolver
+            ->define('alphaMethod')
+            ->default($this->alphaMethod)
+            ->allowedTypes('null', 'int')
+            ->allowedValues(static function ($value) {
```
dbu: could you provide `[null, 0, 1]` instead of a callback? similar above for alphaFilter? (m and the ones with values from 1-100 are fine with the callback. and metadata is a more complex structure, so lets keep that as is too)

dbu (commented):
thanks for the updates!

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
dbu: thanks for checking. that is good enough then.

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: Done. :)

Imagine/Filter/PostProcessor/CwebpPostProcessor.php
DjLeChuck: I'm not a big fan of calling the OptionResolver in the constructor, I prefer to keep it simple (even if yes it a part of his job to ensure data are valid). 

But anyway, if the default parameter if wrong, an error will occurred as expected!

dbu: as long as an invalid default does not make the resolver think that its actually ok to have that wrong value, i'd say we are good. thanks for checking

dbu:
can you please adjust to the cs fixer complaints? (i also usually format chained calls the way you do, but we should not change CS rules within minor versions, it would require to reformat quite a bit of other code i am afraid. so better be consistent i'd say)

and the OptionsResolver::define method is only available with symfony 5. i don't think new versions need to support symfony 4 still, lets drop symfony 4 for the next minor version so we can keep using this construct. could you do a pull request that adjusts composer.json accordingly (and if necessary ci instructions in the .github folder)?

DjLeChuck:
I'm OK for the CS Fixer, I will fix it.

My bad for the `OptionsResolver::define`, I haven't pay attention to the availability. I prefer to use a code compatible with Symfony 4+ if you're OK with that (bad idea to drop the support of a major Symfony version with a minor version of Liip, no? 😄)

dbu:
> I'm OK for the CS Fixer, I will fix it.
thanks.

> My bad for the `OptionsResolver::define`, I haven't pay attention to the availability. I prefer to use a code compatible with Symfony 4+ if you're OK with that (bad idea to drop the support of a major Symfony version with a minor version of Liip, no? smile)

its no problem to drop an outdated symfony version, composer will make sure nothing breaks for people running legacy symfony versions. they would not get the new feature, but symfony 4 is very old, so that would not worry me.

but as you already changed it and the change is compatible with symfony 6, lets keep as is then.

thanks a lot for this contribution!

DjLeChuck:
Thank you for the merge! 😊 

The Symfony 4+ compatibility will be a good thing for certain of our client projects... 😬

dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.8.0

DjLeChuck:
Hi,

Thank you for your review!

I will definitely try to use the OptionResolver. It was my first idea, but
I don't did it to be consistent with others.

I'm on vacation for now, I will address all your points on Saturday!

Le lun. 23 mai 2022 à 09:10, David Buchmann ***@***.***> a
écrit :

PR #1472, opened 2022-06-08
Fix Symfony 6.1 deprecations

willemverspyck:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? |no
| Deprecations? | yes
| License | MIT
| Doc | 

Since Symfony 6.1 setting the $defaultName is deprecated. Prefer to use "AsCommand" attribute, but that not supported yet in PHP ^7.1, so I moved it to the "setName" method.

### DIFF
--- a/Command/RemoveCacheCommand.php
+++ b/Command/RemoveCacheCommand.php
@@ -24,8 +24,6 @@ class RemoveCacheCommand extends Command
 {
     use CacheCommandTrait;
 
-    protected static $defaultName = 'liip:imagine:cache:remove';
-
     /**
      * @var FilterService
      */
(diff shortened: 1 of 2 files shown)


dbu (approved):
things keep changing... thanks for the fix!

dbu:
i merged the 2.x branch into the upcoming 3.x branch. that one has a more restrictive minimal php version requirement - if you have the time it would be great if you can do another PR for the 3.x branch to use the `AsCommand` attribute.

willemverspyck:
Thanks @dbu. I can add the attributes to version 3, but I see it still supports PHP 7.4 and attributes only work for PHP 8 and higher. Will support for PHP 7.4 be dropped for version 3?

dbu:
ah dang. given that symfony 6 only supports php 8, i think the bundle should do the same. can you also do that change in the PR to the 3.x branch?

willemverspyck:
I did the changes to the 3.x branche and removed support of PHP 7.4. Also removed support of Symfony 4.4, because the attribute "AsCommand" is supported since Symfony 5.3. Not sure if you want these changes for 3.x.

PR #1495, opened 2022-12-01
Allow using "vips" driver

### LINKED ISSUE #1494: Unable to use `vips` driver
Hello, I would like to use `vips` driver (https://packagist.org/packages/rokka/imagine-vips), but config:
```yaml
liip_imagine:
    driver: "vips"
```
fails with:
```
Invalid configuration for path "liip_imagine.driver": Invalid imagine driver specified: "vips"
```

Is there any way to get that working?

### Preconditions

1. Bundle version `2.9.0`;
2. Symfony version `6.1.9`;
1. `vips` implementation for imagine https://packagist.org/packages/rokka/imagine-vips;

### Steps to reproduce

Use config:
```yaml
liip_imagine:
    driver: "vips"
```

### Expected result
Can use `vips` driver.

### Actual result
App fails with message "Invalid configuration for path "liip_imagine.driver": Invalid imagine driver specified: "vips"".

PabloKowalczyk:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fix #1494
| License | MIT
| Doc |

### DIFF
--- a/DependencyInjection/LiipImagineExtension.php
+++ b/DependencyInjection/LiipImagineExtension.php
@@ -11,6 +11,7 @@
 
 namespace Liip\ImagineBundle\DependencyInjection;
 
+use Imagine\Vips\Imagine;
 use Liip\ImagineBundle\DependencyInjection\Factory\Loader\LoaderFactoryInterface;
 use Liip\ImagineBundle\DependencyInjection\Factory\Resolver\ResolverFactoryInterface;
 use Liip\ImagineBundle\Imagine\Cache\CacheManager;
@@ -101,13 +102,24 @@ class LiipImagineExtension extends Extension implements PrependExtensionInterfac
             $loader->load('templating.xml');
         }
 
-        $container->setParameter('liip_imagine.driver_service', 'liip_imagine.'.$config['driver']);
+        $driver = $config['driver'];
+        if ('vips' === $driver) {
+            if (!\class_exists(Imagine::class)) {
+                $vipsImagineClass = Imagine::class;
+
+                throw new \RuntimeException("Unable to use 'vips' driver without '{$vipsImagineClass}' class.");
+            }
+
+            $loader->load('imagine_vips.xml');
+        }
+
+        $container->setParameter('liip_imagine.driver_service', "liip_imagine.{$driver}");
 
         $container
             ->getDefinition('liip_imagine.controller.config')
             ->replaceArgument(0, $config['controller']['redirect_response_code']);
 
-        $container->setAlias('liip_imagine', new Alias('liip_imagine.'.$config['driver']));
+        $container->setAlias('liip_imagine', new Alias("liip_imagine.{$driver}"));
         $container->setAlias(CacheManager::class, new Alias('liip_imagine.cache.manager', false));
         $container->setAlias(DataManager::class, new Alias('liip_imagine.data.manager', false));
         $container->setAlias(FilterManager::class, new Alias('liip_imagine.filter.manager', false));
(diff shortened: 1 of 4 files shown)


DependencyInjection/LiipImagineExtension.php
```diff
@@ -101,6 +102,17 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('templating.xml');
         }
 
+        $driver = $config['driver'];
+        if ($driver === 'vips') {
```
dbu:
```suggestion
        if ('vips' === $config['driver']) {
```

we could inline the variable here

Resources/config/imagine_vips.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+
+        <service id="liip_imagine.vips" class="Imagine\Vips\Imagine" public="false">
```
dbu, line 8: are these always autoregistered or should we add tags to the service so it always works?

dbu: ah sorry, found it. the name is used in https://github.com/liip/LiipImagineBundle/blob/ee1581327e4de483e3359f6bb6dcfe7415bf3042/DependencyInjection/LiipImagineExtension.php#L104 so this is all we need.

dbu:
for phpstan we should add imagine-vips to the require-dev section in composer. (we could also ignore the error, but i prefer adding it, so we notice if something goes off.)

DependencyInjection/LiipImagineExtension.php
PabloKowalczyk: I use it in three other places instead.

PabloKowalczyk:
I've added `rokka/imagine-vips` to `require-dev` and to `suggest` also.

DependencyInjection/LiipImagineExtension.php
dbu: ah, i see. yeah lets keep it then.

DependencyInjection/LiipImagineExtension.php
```diff
@@ -101,13 +102,24 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('templating.xml');
         }
 
-        $container->setParameter('liip_imagine.driver_service', 'liip_imagine.'.$config['driver']);
+        $driver = $config['driver'];
+        if ($driver === 'vips') {
```
dbu:
```suggestion
        if ('vips' === $driver) {
```

DependencyInjection/LiipImagineExtension.php
```diff
@@ -101,13 +102,24 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('templating.xml');
         }
 
-        $container->setParameter('liip_imagine.driver_service', 'liip_imagine.'.$config['driver']);
+        $driver = $config['driver'];
+        if ('vips' === $driver) {
+            if (\class_exists(Imagine::class) === false) {
```
dbu:
```suggestion
            if (!\class_exists(Imagine::class)) {
```

dbu:
ah dang. we support older php than vips does. i will try something with the ci.

PabloKowalczyk:
> ah dang. we support older php than vips does. i will try something with the ci.

PHP 7.1 is quite old https://www.php.net/supported-versions.php :)

PR #1509, opened 2023-04-21
Routing file configuration

StevenRenaux:
| Q | A
| --- | ---
| Branch? | 2.x 
| Doc | Routing file configuration

### DIFF
--- a/Resources/doc/installation.rst
+++ b/Resources/doc/installation.rst
@@ -71,7 +71,7 @@ routing file:
 
     .. code-block:: yaml
 
-        # app/config/routing.yml
+        # app/config/route/liip_imagine.yml
         _liip_imagine:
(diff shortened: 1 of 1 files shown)


dbu (approved):
ah, i see this is how the flex recipe sets the bundle up. i agree it makes sense to follow that naming.

it would make sense to have the symfony flex instructions at the top and only specify the manual steps as alternative. do you want to contribute that to the documentation as well?

StevenRenaux:
> ah, i see this is how the flex recipe sets the bundle up. i agree it makes sense to follow that naming.
> 
> it would make sense to have the symfony flex instructions at the top and only specify the manual steps as alternative. do you want to contribute that to the documentation as well?

Yeah it's the https://github.com/liip/LiipImagineBundle/pull/1510

PR #1514, opened 2023-05-15
Add return types to silence Symfony 6.3 deprecations

mbabker:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | N/A
| License | MIT
| Doc | N/A

Symfony 6.3 added a number of return types to the framework. This PR should cover all the ones that popped up when beta testing an application.

Because all affected classes are non-final/internal, doc block annotations are used.

### DIFF
--- a/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
+++ b/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
@@ -29,7 +29,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class AssetsVersionCompilerPass extends AbstractCompilerPass
 {
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         if (!class_exists(StaticVersionStrategy::class)
             // this application has no asset version configured
--- a/DependencyInjection/Compiler/DriverCompilerPass.php
+++ b/DependencyInjection/Compiler/DriverCompilerPass.php
@@ -21,10 +21,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class DriverCompilerPass extends AbstractCompilerPass
 {
-    /**
-     * {@inheritdoc}
-     */
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         $liipImagineDriver = $container->getParameter('liip_imagine.driver_service');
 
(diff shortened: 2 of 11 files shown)


DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -29,6 +29,11 @@
  */
 class AssetsVersionCompilerPass extends AbstractCompilerPass
 {
+    /**
+     * {@inheritdoc}
+     *
+     * @return void
+     */
     public function process(ContainerBuilder $container)
```
dbu:
```suggestion
    public function process(ContainerBuilder $container): void
```
as its not expected that anyone extends these compilar passes, i would prefer to add actual return tye declarations instead of the phpdoc

LiipImagineBundle.php
```diff
@@ -38,6 +38,8 @@ class LiipImagineBundle extends Bundle
 {
     /**
      * {@inheritdoc}
+     *
+     * @return void
      */
     public function build(ContainerBuilder $container)
```
dbu:
```suggestion
    public function build(ContainerBuilder $container): void
```

DependencyInjection/LiipImagineExtension.php
```diff
@@ -59,6 +59,8 @@ public function getConfiguration(array $config, ContainerBuilder $container): ?C
     }
 
     /**
+     * @return void
+     *
      * @see \Symfony\Component\DependencyInjection\Extension.ExtensionInterface::load()
      */
     public function load(array $configs, ContainerBuilder $container)
```
dbu:
```suggestion
    public function load(array $configs, ContainerBuilder $container): void
```

DependencyInjection/Compiler/DriverCompilerPass.php
```diff
@@ -23,6 +23,8 @@ class DriverCompilerPass extends AbstractCompilerPass
 {
     /**
      * {@inheritdoc}
```
dbu: imho we should remove these empty phpdoc everywhere (could be a separate MR), as they do not add information.

dbu (changes requested):
thanks. i would prefer to add actual return type declarations as the compiler passes are not extension points.

mbabker:
> thanks. i would prefer to add actual return type declarations as the compiler passes are not extension points.

Updated

dbu (approved):
awesome, thanks a lot!

dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.11.0

Issue #1524, opened 2023-08-28
Support for JsonManifestVersionStrategy


wouterSkepp:
**Is your feature request related to a problem? Please describe.**
Our projects use the `Symfony\Component\Asset\VersionStrategy\JsonManifestVersionStrategy`, which get populated by our build system (Webpack Encore, in this instance). Each file gets its own versioning string. The bundle does support `StaticVersionStrategy` but, to my knowledge only handles a single versioning string for all assets. It does not support the use of wildcards either.

**Describe the solution you'd like**
Add support for `JsonManifestVersionStrategy`

dbu:
thank you for the contribution: https://github.com/liip/LiipImagineBundle/releases/tag/2.13.0

PR #1525, opened 2023-08-28
Add support for JsonManifestVersionStrategy

wouterSkepp:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1524
| License | MIT
| Doc | 

This PR adds in support for JsonManifestVersionStrategy that is available in Symfony. 
In a JSON manifest, we have a key => value structure where the key is the original filepath, and the value the versioned filestring, e.g:
```
{
  "build/images/cats.jpg": "/build/images/cats.jpg?v=dda78c62c1f2a8793334"
  "build/images/dogs.jpg": "/build/images/dogs.e6fbd6606990f549c912.jpg"
}
```

Similar to `StaticVersionStrategy` that loads it's configuration from the container, this reads the JSON Manifest file defined 

```

framework:
    assets:
        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
        
```

The LazyFilterRuntime has been extended to load the JSON Manifest file contents, and the various methods within to handle both Static & Json versioning.

Note that the current 2.x branch has unrelated failing CI.

I don't think this PR breaks current implementations although test coverage for AssetsVersionCompilerPass currently is lacking.

### DIFF
--- a/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
+++ b/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
@@ -56,7 +56,6 @@ class AssetsVersionCompilerPass extends AbstractCompilerPass
         }
         $version = $versionStrategyDefinition->getArgument(0);
         $format = $versionStrategyDefinition->getArgument(1);
-
         $format = $container->resolveEnvPlaceholders($format);
         if ($format && !$this->str_ends_with($format, '?%%s')) {
             $this->log($container, 'Can not handle assets versioning with custom format "'.$format.'". asset twig function can likely not be used with the imagine_filter');
@@ -67,15 +66,15 @@ class AssetsVersionCompilerPass extends AbstractCompilerPass
         $runtimeDefinition->setArgument(1, $version);
 
         if (is_a($versionStrategyDefinition->getClass(), JsonManifestVersionStrategy::class, true)) {
-            $jsonManifest = file_get_contents($version);
+            $jsonManifestString = file_get_contents($version);
 
-            if (!is_string($jsonManifest)) {
+            if (!\is_string($jsonManifestString)) {
                 $this->log($container, 'Can not handle assets versioning with "'.$versionStrategyDefinition->getClass().'". The manifest file at "'.$version.' " could not be read');
 
                 return;
             }
-            $jsonManifest = \json_decode($jsonManifest, true);
-            if (!is_array($jsonManifest)) {
+            $jsonManifest = json_decode($jsonManifestString, true);
+            if (!\is_array($jsonManifest)) {
                 $this->log($container, 'Can not handle assets versioning with "'.$versionStrategyDefinition->getClass().'". The manifest file at "'.$version.' " does not contain valid JSON');
 
                 return;
--- a/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
+++ b/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
@@ -11,6 +11,7 @@
 
 namespace Liip\ImagineBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\Asset\VersionStrategy\JsonManifestVersionStrategy;
 use Symfony\Component\Asset\VersionStrategy\StaticVersionStrategy;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
@@ -18,20 +19,22 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  * Inject the Symfony framework assets version parameter to the
  * LiipImagineBundle twig extension if possible.
  *
- * We extract the version parameter from the StaticVersionStrategy service
- * definition. If anything is not as expected, we log a warning and do nothing.
+ * We extract either:
+ *  - the version parameter from the StaticVersionStrategy service
+ *  - the json manifest from the JsonManifestVersionStrategy service
+ * If anything is not as expected, we log a warning and do nothing.
  *
  * The expectation is for the user to configure the assets version in liip
  * imagine for custom setups.
  *
- * Anything other than StaticVersionStrategy needs to be implemented by the
- * user in CacheResolveEvent event listeners.
+ * Anything other than StaticVersionStrategy or JsonManifestVersionStrategy needs
+ * to be implemented by the user in CacheResolveEvent event listeners.
  */
 class AssetsVersionCompilerPass extends AbstractCompilerPass
 {
     public function process(ContainerBuilder $container): void
     {
-        if (!class_exists(StaticVersionStrategy::class)
+        if (!class_exists(StaticVersionStrategy::class) || !class_exists(JsonManifestVersionStrategy::class)
             // this application has no asset version configured
             || !$container->has('assets._version__default')
             // we are not using the new LazyFilterRuntime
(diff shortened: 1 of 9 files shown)


DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -11,27 +11,30 @@
...
- * user in CacheResolveEvent event listeners.
+ * Anything other than StaticVersionStrategy or JsonManifestVersionStrategy needs
+ * to be implemented by the user in CacheResolveEvent event listeners.
  */
 class AssetsVersionCompilerPass extends AbstractCompilerPass
 {
     public function process(ContainerBuilder $container): void
     {
-        if (!class_exists(StaticVersionStrategy::class)
+        if (!class_exists(StaticVersionStrategy::class) || !class_exists(JsonManifestVersionStrategy::class)
```
dbu: which strategy is older? as both come from the same component and we only check for class existence, we should only check for the older of the 2 strategies.

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -46,13 +49,14 @@ public function process(ContainerBuilder $container): void
...
         $versionStrategyDefinition = $container->findDefinition('assets._version__default');
-        if (!is_a($versionStrategyDefinition->getClass(), StaticVersionStrategy::class, true)) {
+        if (!is_a($versionStrategyDefinition->getClass(), StaticVersionStrategy::class, true) && !is_a($versionStrategyDefinition->getClass(), JsonManifestVersionStrategy::class, true)) {
             $this->log($container, 'Symfony assets versioning strategy "'.$versionStrategyDefinition->getClass().'" not automatically supported. Configure liip_imagine.twig.assets_version if you have problems with assets versioning');
 
             return;
         }
         $version = $versionStrategyDefinition->getArgument(0);
         $format = $versionStrategyDefinition->getArgument(1);
+
```
dbu:
```suggestion
```

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -61,6 +65,24 @@ public function process(ContainerBuilder $container): void
...
+        if (is_a($versionStrategyDefinition->getClass(), JsonManifestVersionStrategy::class, true)) {
+            $jsonManifest = file_get_contents($version);
+
+            if (!is_string($jsonManifest)) {
+                $this->log($container, 'Can not handle assets versioning with "'.$versionStrategyDefinition->getClass().'". The manifest file at "'.$version.' " could not be read');
+
+                return;
+            }
+            $jsonManifest = \json_decode($jsonManifest, true);
```
dbu: i would prefer differently named variables over changing the type from string to array

Templating/LazyFilterRuntime.php
```diff
@@ -57,33 +63,49 @@ public function filterCache(string $path, string $filter, array $config = [], st
...
 
-        $start = mb_strrpos($path, $this->assetVersion);
-        if (mb_strlen($path) - mb_strlen($this->assetVersion) === $start) {
-            return rtrim(mb_substr($path, 0, $start), '?');
+        if ($this->assetVersion) {
+            $start = mb_strrpos($path, $this->assetVersion);
+            if (mb_strlen($path) - mb_strlen($this->assetVersion) === $start) {
+                return rtrim(mb_substr($path, 0, $start), '?');
+            }
+        } elseif ($this->jsonManifest) {
```
dbu: is it impossible that in symfony both asset version and a json manifest are configured?

if it is possilbe to configure both, it makes sense to look at this second, in case somebody did configure a json manifest but not for images.

Templating/LazyFilterRuntime.php
```diff
@@ -57,33 +63,49 @@ public function filterCache(string $path, string $filter, array $config = [], st
...
+            $asset = array_search($path, $this->jsonManifest, true);
```
dbu: searching a potentially long unsorted array is inefficient.

you could array_flip the array in the constructor and then do an `array_key_exists` and a simple key lookup instead for better performance. i guess both keys and values in json manifest should be unique.

Templating/LazyFilterRuntime.php
```diff
@@ -57,33 +63,49 @@ public function filterCache(string $path, string $filter, array $config = [], st
...
+            if ($asset) {
+                return $asset;
```
dbu: can we be sure that the path from the manifest is right both for when the image needs to be generated and when it already exists?

Templating/LazyFilterRuntime.php
```diff
@@ -57,33 +63,49 @@ public function filterCache(string $path, string $filter, array $config = [], st
...
+        if (!$this->assetVersion && !$this->jsonManifest) {
+            return $resolvedPath;
         }
 
-        $separator = false !== mb_strpos($path, '?') ? '&' : '?';
+        if ($this->assetVersion) {
+            $separator = false !== mb_strpos($resolvedPath, '?') ? '&' : '?';
+
+            return $resolvedPath.$separator.$this->assetVersion;
+        } elseif ($this->jsonManifest) {
```
dbu: this is unnecessary with the check above if both do not exist.

you could
```
if (array_key_exists($path, $this->jsonManifest)) {
    return $this->jsonManifest[$path];
}

return $resolvedPath;
```

Templating/LazyFilterRuntime.php
dbu: could we simplify this? we know json manifest is set at this point, it could be `else`.
or it could be simply `if`, if we want to still do this if assetVersion was set but the strlen / start check abvoe did not result in an early return.

dbu (commented):
thanks a lot for this contribution! cool idea to support this feature.

i am not super sure if the path / resolvedPath are always the file path or if they can be the cache path as well.

dbu:
i will fix the cs failures in the 2.x branch to be clean there.

dbu:
> i will fix the cs failures in the 2.x branch to be clean there.

the 2.x branch is green again. please rebase your branch on 2.x

Templating/LazyFilterRuntime.php
wouterSkepp: Specifying both static and json manifest will give a configuration error:
`Invalid configuration for path "framework.assets": You cannot use both "version" and "json_manifest_path" at the same time under "assets".`

It is also possible to group your assets into packages, and then have each of the packages use a different strategy. This however resolves correctly.

wouterSkepp:
Thank you for your helpful review! Updated PR accordingly.

> i am not super sure if the path / resolvedPath are always the file path or if they can be the cache path as well.

I've not done a deep dive into the caching mechanism, but according to  https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/asset-versioning.rst?plain=1#L10-L19 
and from what I've seen so far is that this is always the original filepath.

Templating/LazyFilterRuntime.php
dbu: cool!

dbu:
i rebased and squashed the commits in #1529

PR #1529, opened 2023-09-07
Add support for JsonManifestVersionStrategy

dbu:
wrap up #1525

* [x] Why is JsonManifestVersionStrategy not found by phpstan?
* [x] Adjust Resources/doc/asset-versioning.rst to document the new functionality
* [x] Check file format changes @wouterSkepp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ for a given releases. Unreleased, upcoming changes will be updated here periodic
 
 # 2.x
 
+## [2.13.0](https://github.com/liip/LiipImagineBundle/tree/2.13.0)
+
+- Support JsonManifestVersionStrategy that was added in Symfony 6.
+
 ## [2.12.3](https://github.com/liip/LiipImagineBundle/tree/2.12.3)
 
 - Add alias for `Imagine\Image\ImagineInterface` to help autowiring.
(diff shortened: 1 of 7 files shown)


dbu:
@wouterSkepp could you please do a pull request against this branch to adjust the [asset versioning documentation](Resources/doc/asset-versioning.rst) ? apart from that i think we are ready to merge.

wouterSkepp:
@dbu I might have overlooked this bundles ability to change formats in my tests. (eg. the original file is a png and the filter reformats it to a .jpg)
I just stumbIed across https://github.com/liip/LiipImagineBundle/issues/1198 about it; I've not tested this PR against these cases and I don't think it's handled correctly by my current implementation. (e.g. the filtered jpg still has it's .png extension)

I'm willing to have a look at this later on but it'll take a while because of vacationing plans.

dbu:
thank you for the heads up! then i will refrain from merging until you (or somebody else) has been able to check that use case and possibly adjust this PR to handle it correctly.

dbu:
@wouterSkepp do you have time to look into this so we can wrap up the feature?

wouterSkepp:
@dbu well this took a bit longer than expected. :)

When converting images to a new format, the extension did not get changed as I suspected. Aparantly that is default behaviour (for non-versioned assets as well), unless you decorate the resolver with the FormatExtensionResolver?

For those cases I've updated the logic, and these now support extension rewriting by the FormatExtensionResolver.

dbu:
thanks a lot for the contribution!

PR #1538, opened 2023-11-20
Update configuration.rst

ThomasBeloBeesCom:
I found a mistake in this default configuration.
I have problem with the generation of thumbnail. When i try to replace 'liip_imagine.controller:filterAction' to liip_imagine.controller::filterAction that solve my problem.

So i think it's need to have '::' and not ':'.

| Q | A
| --- | ---
| Branch? | 2.x or 3.x 
| Bug fix? | yes/no
| New feature? | yes/no
| BC breaks? | yes/no
| Deprecations? | yes/no
| Fixed tickets | 
| License | MIT
| Doc |

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -40,8 +40,8 @@ The default configuration for the bundle looks like this:
             filters: []
             post_processors: []
         controller:
-            filter_action:          liip_imagine.controller:filterAction
-            filter_runtime_action:  liip_imagine.controller:filterRuntimeAction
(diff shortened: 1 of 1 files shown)


dbu (approved):
thanks for spotting this. this used to be the syntax in old versions of symfony but changed long ago. sorry you ran into this and thanks for fixing it so others don't get confused by it.

eventually we could also change from service name to full class name as is the current best practice of symfony. but for that we would need to also make sure that all liip imagine things are actually available as services with the class names.

PR #1554, opened 2024-01-03
Add aliases for manager services for improved autowiring support

mbabker:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | N/A
| License | MIT
| Doc | N/A

The `liip_imagine.cache.manager`, `liip_imagine.data.manager`, and `liip_imagine.filter.manager` services cannot be autowired at present, requiring folks to either manually create aliases in their app's configuration or to use one of the newer dependency injection attributes to specify the right service in their classes.  This PR adds the appropriate service aliases to support autowiring and remove that little bit of extra overhead needed to use these three services.

### DIFF
--- a/Resources/config/imagine.xml
+++ b/Resources/config/imagine.xml
@@ -125,6 +125,7 @@
             <argument type="service" id="liip_imagine" />
             <argument type="service" id="liip_imagine.binary.mime_type_guesser" />
         </service>
+        <service id="Liip\ImagineBundle\Imagine\Filter\FilterManager" alias="liip_imagine.filter.manager"/>
 
         <service id="liip_imagine.data.manager" class="Liip\ImagineBundle\Imagine\Data\DataManager" public="true">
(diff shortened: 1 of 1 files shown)


dbu (commented):
thanks a lot!

can you please also update the documentation of the corresponding services? i see that filter and data do not mention the service names but cache does, so there we should change it to use the CacheManager FQN as service name.

mbabker:
> can you please also update the documentation of the corresponding services?

Sorry, I forgot all about this ask with the way things have gone for me over the last several weeks.  I'll try to throw a PR together for the docs soon.

dbu:
no worries. i also forgot and just noticed now. i adjusted the doc in 13be8f6810407a34241c1efe6ffcbded4f3efc39

i want to cleanup rather radically for 3.x, i think no further doc refinement for the 2.x is needed now.

and thanks for the contribution.

PR #1559, opened 2024-01-06
add deprecation note.

tacman:
Having both amazon-s3 and aws-s3 is confusing.

### DIFF
--- a/Resources/doc/cache-resolver/amazons3.rst
+++ b/Resources/doc/cache-resolver/amazons3.rst
@@ -7,6 +7,8 @@ Amazon S3 Resolver
 The ``AmazonS3Resolver`` resolver enables cache resolution using the
 ``\AmazonS3`` storage API.
 
+This will be deprecated in Version 3, use aws-s3 resolver instead.
+
 Dependencies
 ------------
 
@@ -15,7 +17,7 @@ by executing the following command in your project directory:
 
 .. code-block:: bash
 
-    $ composer require aws/aws-sdk-php
+    composer require aws/aws-sdk-php
 
 Configuration
 -------------
--- a/Resources/doc/cache-resolver/amazons3.rst
+++ b/Resources/doc/cache-resolver/amazons3.rst
@@ -7,7 +7,7 @@ Amazon S3 Resolver
 The ``AmazonS3Resolver`` resolver enables cache resolution using the
 ``\AmazonS3`` storage API.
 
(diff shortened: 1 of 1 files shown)


Resources/doc/cache-resolver/amazons3.rst
```diff
@@ -7,6 +7,8 @@ Amazon S3 Resolver
 The ``AmazonS3Resolver`` resolver enables cache resolution using the
 ``\AmazonS3`` storage API.
 
+This will be deprecated in Version 3, use aws-s3 resolver instead.
```
dbu: we deprecate it now and remove it in version 3, lets say:

```suggestion
**Deprecated**, use the aws-s3 resolver instead. The legacy AmazonS3Resolver will be removed in version 3.
```

dbu:
do we need an S3 resolver at all in version 3? or could we do all with flysystem? or can flysystem not provide the web URL to a file on S3?

tacman:
Hmm, interesting question.  Regardless, we need more documentation and examples.

This is what I use for s3 on production and local filesystem in dev.

```yaml
# config/packages/flysystem.yaml
# Read the documentation at https://github.com/thephpleague/flysystem-bundle/blob/master/docs/1-getting-started.md
flysystem:
    storages:
        default.storage:
            adapter: 'aws'
            # visibility: public # Make the uploaded file publicly accessible in S3
            options:
                client: 'Aws\S3\S3Client' # The service ID of the Aws\S3\S3Client instance
                bucket: '%env(AWS_S3_BUCKET_NAME)%'
                streamReads: true
                prefix: '%env(S3_STORAGE_PREFIX)%'
when@dev:
    flysystem:
        storages:
            default.storage:
                adapter: 'local'
                options:
                    directory: '%kernel.project_dir%/public/images'

```

And

```yaml
# config/packages/liip_imagine.yamn
liip_imagine:
    driver: "gd"
    twig:
        mode: lazy

    loaders:
        flysystem_loader:
            flysystem:
                # this comes from flysystem.yaml
                filesystem_service: default.storage

    # default loader to use for all filter sets
    data_loader: flysystem_loader
```

dbu:
looks like modern versions of flysystem actually support that: https://flysystem.thephpleague.com/docs/usage/public-urls/

i think the way to go for 3.x is to leverage flysystem as much as we can to avoid this bundle having to care about storage and filesystem and such stuff.

PR #1568, opened 2024-02-23
Fix PHP str_replace call with parameter #3 is null deprecation

simonberger:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? |no
| BC breaks? |no
| Deprecations? |no
| License | MIT

When the cache_prefix is not set for the flysystem resolver, it results in
```
Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated
```
in the  `FlysystemV2Resolver` while fixing I saw the same should happen for the `AwsS3Resolver`.

There are multiple ways to fix this easily. I decided to do it very early by changing the default value of the configuration property, because null also already violated the doctype string of the parameter in the resolvers.

### DIFF
--- a/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
+++ b/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
@@ -92,7 +92,7 @@ class AwsS3ResolverFactory extends AbstractResolverFactory
                     ->cannotBeEmpty()
                 ->end()
                 ->scalarNode('cache_prefix')
-                    ->defaultValue(null)
+                    ->defaultValue('')
                 ->end()
                 ->arrayNode('client_config')
                     ->isRequired()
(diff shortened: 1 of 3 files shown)


dbu (commented):
thanks! we could alternatively check in the DI extension if the prefix is non-null and otherwise not do the str_replace at all.

with only changing the defaults, the user could still configure the values to null in their config. but i consider this good enough for now.

dbu:
we have a test on the configuration handling that is now failing. can you please adjust the test?

simonberger:
Sure, I'll fix the tests this evening.
If I understood the upgrade log of 3.0 correctly this part is gone anyway (I might be wrong, but at least I could not find comparable code there on a quick look), which is why I also did not consider a more in-depth PR.

dbu:
thanks. the flysystem should still be around, but legacy aws is removed.

I hope that we can actually remove all the remote storage things in favor of flysystem - could well be that we change a lot more there. so fully agree to not do more on 2.x :+1:

dbu:
thanks a lot!

dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.12.2

PR #1575, opened 2024-03-22
[Doc] Misc fixes in some doc pages

javiereguiluz:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| License | MIT
| Doc | -

Fixes these errors shown when building the docs:

```
Found invalid reference "../events.rst" in file "cache-resolver/aws_s3"
Found invalid reference "Imagine\Image\ImageInterface" in file "filters/general"
Found invalid reference "sys_get_temp_dir()" in file "filters/general"
```

### DIFF
--- a/Resources/doc/cache-resolver/aws_s3.rst
+++ b/Resources/doc/cache-resolver/aws_s3.rst
@@ -307,7 +307,7 @@ You can also use the constructor of the resolver to directly inject multiple opt
             tags:
                 - { name: "liip_imagine.cache.resolver", resolver: "aws_s3_resolver" }
 
-You can find an example with private ACL and signed url on the :doc:`events chapter <../events.rst>`.
+You can find an example with private ACL and signed url on the :doc:`events chapter </events>`.
 
 .. _`aws-sdk-php`: https://github.com/amazonwebservices/aws-sdk-for-php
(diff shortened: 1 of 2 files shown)


dbu (approved):
thank you.

you build this doc with sphinx, right? is it built-standalone or integrated with the rest of the symfony doc? if it is stand-alone, we could add a build in the ci of the bundle to get immediate feedback when we break something.

javiereguiluz:
@dbu we used Sphinx in the past, but it's hard to maintain ... so we moved from Sphinx (Python) to a custom-made and 100% PHP alternative: https://github.com/symfony-tools/docs-builder

You can see it in action here: https://github.com/symfony/symfony-docs/tree/7.1/_build

This custom-made solution used packages from other communities, like Doctrine's RST parser. Other projects like TYPO3 (https://github.com/TYPO3-Documentation) also wanted a PHP-based RST parser ... so, Doctrine, TYPO3, Symfony, etc. joined forces and there's a new tool to replace the existing one. 

We don't use it yet to build Symfony docs, but I hope to use it "soon".

dbu:
ah, glad to hear. but we won't need to change the doc format, right?

javiereguiluz:
Correct! We still use the exact same RST syntax as before ... but now, we handle it with PHP code.

Issue #1576, opened 2024-03-22
Regex in redirect optimization doc matches more than expected


e-five256:
### Preconditions

This is related to the avoid redirect documentation for at least both [2.x](https://github.com/liip/LiipImagineBundle/blob/2.x/Resources/doc/optimizations/avoid-redirects.rst) and [3.x](https://github.com/liip/LiipImagineBundle/blob/3.x/doc/optimizations/avoid-redirects.rst)

### Steps to reproduce

We used the provided web server configuration in nginx in this doc:

```
location ~* .(js|jpg|jpeg|gif|png|css|tgz|gz|rar|bz2|doc|pdf|ppt|tar|wav|bmp|rtf|swf|ico|flv|txt|woff|woff2|svg)$ {
```

And found that on URLs that ended in any of these things, for example `/FirefoxCSS`, `/ruleico`, the routes were not working. It turned out the `.` was being used as the regex match on any character, so `xCSS` and `eico` would match. I believe the goal is to match on the file type, so we changed it to `\.` to fix this issue

### Expected result

The regex to not match when the location is not a specific file / file type suffix

### Actual result

I can't quite remember the actual result, I believe it was a 404, or at the very least the pages did not work

Edit: Ah, sorry, it might not be completely correct to say that. The routes responded as webpages fine, but it broke activitypub for those routes (getting a bit specific in implementation details, but the same urls also need to respond to `"Accept: application/activity+json" ` requests and they were no longer doing so. I think either way the intent wasn't to match on non-full stop characters for that regex.

dbu:
yep, i agree that the intention was to match on literal `.` and not match any character. if you have a specific prefix path to the image controller set up, you could also add that at the beginning of the matching (which would probably also be more performant, which matters given this is applied to every request). something like `location ~* ^/my-prefix/.*\.(js|...`

do you want to do a pull request (against the 2.x branch, we will then merge that into 3.x to keep them in sync)? ideally, you try the prefix and escaping the `.` in your system to make sure the doc is working as specified (i would assume in nginx we don't need to double-escape `\` but with these things its always best to actually try to be sure :-)

dbu:
the `*` after `~` btw is to make the regex case sensitive. i think we still want that, we don't expect sudden uppercase file extensions.

e-five256:
[...]

PR #1578, opened 2024-03-23
Avoid redirects documentation - escape full stop in location regex

e-five256:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1576
| License | MIT
| Doc | 

Escapes the `.` character appearing in the suggested regex, as it has the potential of matching routes that end in a file type suffix


dbu:
thanks!

PR #1579, opened 2024-05-05
add default values to webp parameters

tacman:
Makes autowiring a bit easier.


dbu:
we can do this, it reflects the defaults in Configuration. can you please fix the cs issue? (spaces around the `=`)

just out of curiosity: when you use the service as configured by the bundle, the configuration class sets these values as default and the extension configures the parameters. what is your use case to define your own service with this? (i am thinking about a larger rewrite of the library and knowing current use cases / shortcomings is interesting for that)

dbu:
thanks for the contribution

Issue #1413, opened 2021-11-02
refactor naming and workflow of filtered images generation


dbu:
i am tempted to cleanup some of the process and also naming how the filtered image files are created. especially with webp support, things are a bit muddy to me. 

"resolve cache" is used to populate the cache. that is a strange name. we should streamline naming.

we have several classes that are not super obvious, with FilterPathContainer and FilterService.

WarmupCacheHandler calls FilterService::warmupCache but then also FilterService::getUrlOfFilteredImage (because of #1347). the only thing i see that the additional call does is triggering resolve events. i think it would be better to have warmupCache also trigger those events. ResolveCacheCommand does only call warmupCache, so it won't trigger the events if i understand correctly.

PR #1590, opened 2024-06-19
[Draft] Clean interfaces for the imagine processes

dbu:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | no, but better names
| BC breaks? | yes!
| Deprecations? | no
| Fixed tickets | #1413
| License | MIT
| Doc | TODO

Draft for clean interfaces as discussed at the Hackday at Liip.

We added the new interfaces and implementations under `Domain` to see what is new, but should move them to the root folder when things are ready.

Things that are missing in the draft
* Event system (do we want to keep it, do we need to change it?)
* Can we leverage the `Config` namespace somehow?
* One-shot transformations with a dynamic configuration

Will be done separatly in the `refactor` main branch after merging this branch:
* Adjust controller, commands and message handler to use the new interface #1593
* Refactor filters, post processors and flysystem to the new interfaces #1594
* Write ImageStore and ImageLoader with flysystem #1595


src/Domain/ImageReference.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Liip\ImagineBundle\Domain;
+
+interface ImageReference
```
dbu, line 5: i renamed this to ImageReference as its the reference to an image rather than an actual image. i noticed when i got confused in LiipTransformerStackExecutor where we also have the imagine ImageInterface which holds the actual image data and has filters applied to it.

src/Domain/Imagine/Filter/FilterExecutor.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Domain\Imagine\Filter;
+
+use Imagine\Image\ImageInterface;
+
+interface FilterExecutor
```
dbu, line 16: better name for Liip\ImagineBundle\Imagine\Filter\Loader\LoaderInterface - functionality does not need to be changed.

src/Domain/ImagineTransformer.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function warmupCache(string $sourceImageId, array $stackNames, array $targetFormats): void;
+
+    /**
+     * Remove cached images for the specified source image.
+     *
+     * @param string $sourceImageId Identifier for the image, for example a file system path
+     * @param string[] $stackNames Limit invalidation to the specified stacks. If empty, all stacks are invalidated.
+     */
+    public function invalidateCache(string $sourceImageId, array $stackNames = []): void;
```
dbu, line 52: this is not strictly about transforming, but imho makes sense to have on the main interaction point with the system.

src/Domain/ImagineTransformer.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * If the result image is already cached, it will be regenerated and overwritten.
+     *
+     * There is no cache lifetime defined. The application is expected to use active invalidation if the source image changes.
+     *
+     * @param string $sourceImageId Identifier for the image, for example a file system path
+     * @param string[] $stackNames
+     * @param string[] $targetFormats
+     */
+    public function warmupCache(string $sourceImageId, array $stackNames, array $targetFormats): void;
```
dbu, line 44: i realized we said that transformToUrl should ideally not be slow and not necessarily generates the image right away but can chose to only generate it on demand. and when that is the case, we need this separate method to force warming up the cache.

src/Domain/Stack/LiipTransformerStack.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Liip\ImagineBundle\Domain\Stack;
+
+use Liip\ImagineBundle\Domain\ImageReference;
+use Liip\ImagineBundle\Imagine\Filter\PostProcessor\PostProcessorInterface;
+
+class LiipTransformerStack implements TransformerStack
```
dbu, line 8: TODO: either finish this or remove it.

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+use Imagine\Image\ImageInterface;
+use Imagine\Image\ImagineInterface;
+use Liip\ImagineBundle\Binary\MimeTypeGuesserInterface;
+use Liip\ImagineBundle\Domain\ImageReference;
+use Liip\ImagineBundle\Domain\ImageReferenceFile;
+use Liip\ImagineBundle\Domain\Imagine\Filter\FilterExecutor;
+use Liip\ImagineBundle\Domain\PostProcessor\PostProcessor;
+use Liip\ImagineBundle\Imagine\Filter\FilterConfiguration;
+
+class LiipTransformerStackExecutor implements TransformerStackExecutor
```
dbu, line 14: this is both the container for the filters and post processors and the executor for them.

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+        ];
+
+        if ($sourceImageReference instanceof ImageReferenceFile) {
+            $image = $this->imagine->open($sourceImageReference->getPath());
+        } else {
+            $image = $this->imagine->load($sourceImageReference->getContent());
+        }
+
+        $image = $this->applyFilters($image, $config);
+        $resultImageReference = $this->exportConfiguredImageBinary($sourceImageReference, $image, $config);
```
dbu, line 67: we need this step before we apply post processors. and it also depends on the stack configuration. so i think we need all 3 things in the same place and can not separate them.

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+        foreach ($this->sanitizePostProcessors($config['post_processors'] ?? []) as $name => $options) {
+            $image = $this->postProcessors[$name]->process($image, $options);
+        }
+
+        return $image;
+    }
+
+    private function exportConfiguredImageBinary(ImageReference $imageReference, ImageInterface $image, array $config): ImageReference
+    {
+        // TODO: this is for now literal copy-paste from FilterManager
```
dbu, line 97: please challenge the things below, but i don't know if we can do much better. i'd not change the configuration if we don't really need to.

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+
+        if ('gif' === $imageReference->getFormat() && $config['animated']) {
+            $options['animated'] = $config['animated'];
+        }
+
+        $filteredFormat = $config['format'] ?? $imageReference->getFormat();
+        try {
+            $filteredString = $image->get($filteredFormat, $options);
+        } catch (\Exception $exception) {
+            // TODO: why only a problem for webp but not png/jpg? and should the stack handle this
```
dbu, line 120: we could also throw an exception (possibly containing the alternate gif image)

the problem with this is that we don't return the requested format. but from the comment below it seems we can't easily tell in advance if we do an animated gif...

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+            if ('webp' !== $filteredFormat || !\array_key_exists('animated', $options) || true !== $options['animated']) {
+                throw $exception;
+            }
+            $filteredFormat = 'gif';
+            $filteredString = $image->get($filteredFormat, $options);
+        }
+
+        $this->destroyImage($image);
+
+        return new ImageReferenceInstance( // TODO
```
dbu, line 132: need to implement some in-memory image

src/Domain/Stack/LiipTransformerStackExecutor.php
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Report all non-existing filters from the configuration.
+     *
+     * This is better than a simple array_key_exists check, which would report the issues only one at a time.
+     */
+    private function sanitizeFilters(array $filters): array
+    {
+        // TODO: is this much better than something like?
+        // if ($missing = array_diff(array_keys($filters), array_keys($this->filters))) { throw new \InvalidArgumentException('missing'.implode(',', $missing)); }
```
dbu, line 147: is the below code just legacy or is it that much more efficient?

src/Domain/Storage/ImageStore.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param string[] $stackNames Remove cached images with specified stack names. If empty, all stacks are deleted.
+     */
+    public function delete(string $imageId, array $stackNames = []): void;
+
+    /**
+     * Whether this store can create the image on the fly when the URL it returns is fetched.
+     *
+     * (e.g. the URL points to a controller or a 404 handler is used to generate the image on demand)
+     */
+    public function supportsOnDemandCreation(): bool;
```
dbu, line 27: i added this to handle both scenarios of having to generate the image when requesting the url or capable of generating on demand.

we could do a separate interface for it (but would need the method anyways because depending on the store this option could depend on whether configuration is set up)

src/Domain/ImagineTransformer.php
dbu: as per discussion in #1602: there is also a use case to make sure an image exists without forcing regeneration if the image already exists.

we could add a flag to this method to control regeneration:

```suggestion
    public function warmupCache(string $sourceImageId, array $stackNames, array $targetFormats, bool $regenerateExisting): void;
```

on the other hand, that use case would like to get the image, so maybe a separate method `public function getImage(string $sourceImageId, string $stackName, string $targetFormat): ImageReference` would be more suited for that use case.

PR #1600, opened 2024-06-30
Fix trying to open non-existant manifest file

AirBair:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

Hello,

Following #1529, `cache:clear` auto-script in fresh composer install for example is crashing because the manifest.json does not exist yet.

image

This PR add a check on the existence of the file before trying to open it.

### DIFF
--- a/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
+++ b/DependencyInjection/Compiler/AssetsVersionCompilerPass.php
@@ -68,6 +68,11 @@ class AssetsVersionCompilerPass extends AbstractCompilerPass
         $runtimeDefinition->setArgument(1, $version);
 
         if (is_a($versionStrategyDefinition->getClass(), JsonManifestVersionStrategy::class, true)) {
+            if (!file_exists($version)) {
+                $this->log($container, 'The manifest file at "'.$version.'" does not yet exist');
+
+                return;
+            }
             $jsonManifestString = file_get_contents($version);
 
             if (!\is_string($jsonManifestString)) {


DependencyInjection/Compiler/AssetsVersionCompilerPass.php
```diff
@@ -68,7 +68,7 @@ public function process(ContainerBuilder $container): void
         $runtimeDefinition->setArgument(1, $version);
 
         if (is_a($versionStrategyDefinition->getClass(), JsonManifestVersionStrategy::class, true)) {
-            $jsonManifestString = file_get_contents($version);
+            $jsonManifestString = @file_get_contents($version);
```
dbu: i really dislike ignoring errors. could we instead do

```suggestion
            if (!file_exists($version)) {
                $this->log($container, 'The manifest file at "'.$version.'" does not yet exist');
                
                return;
            }
            $jsonManifestString = file_get_contents($version);
```

dbu (changes requested):
thanks for looking into this. i suggest to explicitly check for the file to have better error reporting.

DependencyInjection/Compiler/AssetsVersionCompilerPass.php
AirBair: I didn't add it in the first place because `file_get_contents` doesn't trigger an error, only a E_WARNING which doesn't halt the script. It's the cache:clear of symfony who transform this into an error afterwards.

And since $jsonManifestString is set to false with or without the error control operator, the unreadable file is handled by the next condition.

But as you wish, the log message will be clearer with your proposed solution !

dbu: yeah, i guess the check if the contents of the file is empty would have caught it. but this seems clearer to me.

thanks for looking into the issue and providing the fix!

dbu:
tagged as https://github.com/liip/LiipImagineBundle/releases/tag/2.13.1

PR #1610, opened 2024-09-04
apply latest cs fixer

dbu:


JohJohan:
Looks like php-cs-fixer applies some stuff which isn't allowed for PHP 7.x stuff like trailing comma https://php.watch/versions/8.0/trailing-comma-parameter-use-list

Here is the php-cs-fixer rule: [trailing_comma_in_multiline](https://cs.symfony.com/doc/rules/control_structure/trailing_comma_in_multiline.html)

dbu:
yep, i was running the fixer with php 8.3 locally :facepalm: 

fixed those now.

PR #1614, opened 2024-12-12
Check if path is a file in FileSystemLoader

revoltek-daniel:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

If the resolve URL is called without a file like "media/cache/resolve/image_xs/media/image/40/93/" the FileinfoMimeTypeGuesser will fail with a InvalidArgumentException, because the mime type checker will check if the path is a file. https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Mime/FileinfoMimeTypeGuesser.php#L47

So we should check before if it is a file or catch the exception.

### DIFF
--- a/Binary/Loader/FileSystemLoader.php
+++ b/Binary/Loader/FileSystemLoader.php
@@ -12,6 +12,7 @@
 namespace Liip\ImagineBundle\Binary\Loader;
 
 use Liip\ImagineBundle\Binary\Locator\LocatorInterface;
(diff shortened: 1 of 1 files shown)


dbu (approved):
this makes sense, thank you!

dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.13.3

PR #1629, opened 2025-08-26
Support disabled acls on s3 buckets

deguif:
| Q | A
| --- | ---
| Branch? | 2.x 
| Bug fix? | yes/no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes/no
| Fixed tickets |
| License | MIT
| Doc |

AWS now recommends to disable ACLs on S3 buckets, this PR is adaptating s3 configuration in order to allow to not send acl when the bucket has ACLs disabled.
In the meantime it's possible to send `bucket-owner-full-control` so that bucket with ACL enabled can later be switched to ACL disabled without issues.

Here's the documentation from AWS : https://docs.aws.amazon.com/AmazonS3/latest/userguide/ensure-object-ownership.html

### DIFF
--- a/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
+++ b/DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
@@ -96,7 +96,6 @@ class AwsS3ResolverFactory extends AbstractResolverFactory
                 ->end()
                 ->scalarNode('acl')
                     ->defaultValue('public-read')
-                    ->cannotBeEmpty()
                 ->end()
                 ->scalarNode('cache_prefix')
                     ->defaultValue('')
--- a/Imagine/Cache/Resolver/AwsS3Resolver.php
+++ b/Imagine/Cache/Resolver/AwsS3Resolver.php
@@ -58,17 +58,17 @@ class AwsS3Resolver implements ResolverInterface
     /**
      * Constructs a cache resolver storing images on Amazon S3.
      *
-     * @param S3Client $storage    The Amazon S3 storage API. It's required to know authentication information
-     * @param string   $bucket     The bucket name to operate on
-     * @param string   $acl        The ACL to use when storing new objects. Default: owner read/write, public read
-     * @param array    $getOptions A list of options to be passed when retrieving the object url from Amazon S3
-     * @param array    $putOptions A list of options to be passed when saving the object to Amazon S3
+     * @param S3Client    $storage    The Amazon S3 storage API. It's required to know authentication information
+     * @param string      $bucket     The bucket name to operate on
+     * @param string|null $acl        The ACL to use when storing new objects. Default: owner read/write, public read
+     * @param array       $getOptions A list of options to be passed when retrieving the object url from Amazon S3
+     * @param array       $putOptions A list of options to be passed when saving the object to Amazon S3
      */
     public function __construct(S3Client $storage, $bucket, $acl = 'public-read', array $getOptions = [], $putOptions = [])
     {
         $this->storage = $storage;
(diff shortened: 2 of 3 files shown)


Imagine/Cache/Resolver/AwsS3Resolver.php
```diff
@@ -99,20 +99,19 @@ public function resolve($path, $filter)
...
     {
         $objectPath = $this->getObjectPath($path, $filter);
+        $options = [
+            'Bucket' => $this->bucket,
+            'Key' => $objectPath,
+            'Body' => $binary->getContent(),
+            'ContentType' => $binary->getMimeType(),
+        ];
+
+        if ('' !== $this->acl) {
```
deguif, line 109: we should maybe use null, but as `$acl` is type-hinted as string in constructor documentation, in order to keep BC, I used an empty string to disable sending ACL.

dbu: the constructor is less the problem than the protected field - in case somebody extended the class.

as its only documentation, i suggest we change the constructor doc to allow null. we still need to handle both null and '' here then, to be strictly BC.

not sure when/if it is going to happen, but my vision is to remove all the S3 and whatnot from this bundle and rely much more on flysystem for all of this...

DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
```diff
@@ -87,9 +87,8 @@ public function addConfiguration(ArrayNodeDefinition $builder)
                 ->scalarNode('cache')
                     ->defaultFalse()
                 ->end()
-                ->scalarNode('acl')
+                ->stringNode('acl')
```
dbu: we seem to support some very old symfony version. for the 2.x branch i'd rather not change that.

Imagine/Cache/Resolver/AwsS3Resolver.php
deguif: Do you mean, that I should allow null on the constructor but store an empty string on the protected field to be strictly BC ?

DependencyInjection/Factory/Resolver/AwsS3ResolverFactory.php
deguif: Yes, I'm going to revert this change

Imagine/Cache/Resolver/AwsS3Resolver.php
deguif: > not sure when/if it is going to happen, but my vision is to remove all the S3 and whatnot from this bundle and rely much more on flysystem for all of this...

Well for the moment, we rely on this feature, that's why I make some PRs to simplify the advanced usage we had by overriding the services and bypassing the config ;)

dbu: the flysystem remark was more for larger context - as i don't have capacity to work on that, your contributions are appreciated to improve the current setup.

> Do you mean, that I should allow null on the constructor but store an empty string on the protected field to be strictly BC ?

yes, i'd like to have it like that. other people might have their own workarounds in place, so if we can avoid BC breaks here it would be good. (one of the reasons why protected properties and non-final classes is not ideal for maintenance)

dbu:
can you please change https://github.com/liip/LiipImagineBundle/pull/1629/files#diff-102332ec3bfab7b8cfbec11947c9377467b02f0f52ed522dcc9b9f1906f0615bR71 to cast $acl to string in the constructor? and rebase on 2.x and add a changelog entry explaining what people should do?

dbu (approved):
thanks

PR #1637, opened 2025-10-13
Start moving Symfony DI config from XML to PHP format

dmaicher:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | Part of https://github.com/liip/LiipImagineBundle/issues/1634
| License | MIT
| Doc | 

Part of https://github.com/liip/LiipImagineBundle/issues/1634. This moves some first configs from XML to PHP.

PR #1639, opened 2025-10-13
drop support for unmaintained Symfony versions

dmaicher:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

See https://github.com/liip/LiipImagineBundle/pull/1637#discussion_r2426797017

I think it would make sense to do this first


dbu (approved):
wow, such old :sweat_smile: 

thanks a lot for the careful work!

the matrix of supported PHP versions is also growing big, but symfony 5.4 allows PHP 7.2, so unless it causes major pains we should keep those versions.

PR #1642, opened 2025-10-29
Allow Symfony 8

dmaicher:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| License | MIT
| Doc | 

This adds Symfony 8 compatibility.

I added a dev dependencies CI build for now. A bit annoying is the dev dependency on https://github.com/php-enqueue/enqueue-bundle which seems dead :thinking: So I removed that bundle for the Symfony 8 version on CI.


.github/workflows/phpunit.yml
```diff
@@ -86,6 +89,8 @@ jobs:
         uses: ramsey/composer-install@v2
         with:
           dependency-versions: ${{ matrix.dependencies }}
+        env:
+          SYMFONY_REQUIRE: "${{ matrix.symfony }}"
```
dmaicher, line 93: using this ENV to force a specific Symfony version is more standard nowadays

dbu (approved):
awesome! and also great that it just works out of the box.

thanks a lot for the work! i will merge but i guess we wait with tagging until symfony 8.0.0 is tagged, just to be sure not to produce a problematic version.

dbu:
yeah enqueue support for sure will be remove in the next version, symfony messenger is the way to go with a symfony bundle. if we even need an explicit message queue support - providing clear services to warm the cache should be the priority, and then a message queue integration should become so trivial that it could just be a section in the documentation.

PR #1644, opened 2025-11-24
test with php 8.5

dbu:
| Q | A
| --- | ---
| Branch? | 2.x 
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| License | MIT
| Doc | just testing with php 8.5

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
+        php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5']
         dependencies: [highest]
         symfony: ['*']
         stability: ['stable']
@@ -26,7 +26,7 @@ jobs:
             stability: 'stable'
 
           # Minimum supported dependencies with the latest supported PHP version
-          - php: '8.4'
+          - php: '8.5'
             dependencies: lowest
             symfony: '*'
             stability: 'stable'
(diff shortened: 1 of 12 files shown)


dbu:
deprecations are not errors. but aparently phpunit or symfony make the the build fail on deprecations.

but as we see those deprecations...  the aws package we use is abandoned and full of `(integer)` casts which are deprecated in php 8.5. we really should go forward with the rewrite that bases itself on flysystem, i don't want to replace the aws package in this bundle, that feels like effort in the wrong place.

dmaicher:
Indeed it can be customized so the CI job won't fail in those cases. See https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail

.github/workflows/phpunit.yml
```diff
@@ -43,13 +43,13 @@ jobs:
             stability: 'stable'
 
           # Test Symfony 7.4 dev version
-          - php: '8.4'
+          - php: '8.5'
             dependencies: highest
             symfony: '7.4.*'
             stability: 'dev'
```
dmaicher: I think we can also change this to `stable` now that 7.4 and 8.0 are released?

.github/workflows/phpunit.yml
```diff
@@ -43,13 +43,13 @@ jobs:
...
             
           # Test Symfony 8.0 dev version
-          - php: '8.4'
+          - php: '8.5'
             dependencies: highest
             symfony: '8.0.*'
             stability: 'dev'
```
dmaicher: and here

.github/workflows/phpunit.yml
dbu: yep

.github/workflows/phpunit.yml
```diff
@@ -43,16 +43,16 @@ jobs:
             stability: 'stable'
 
           # Test Symfony 7.4 dev version
```
dmaicher:
```suggestion
          # Test Symfony 7.4
```

.github/workflows/phpunit.yml
```diff
@@ -43,16 +43,16 @@ jobs:
...
-          - php: '8.4'
+          - php: '8.5'
             dependencies: highest
             symfony: '7.4.*'
-            stability: 'dev'
+            stability: 'stable'
             
           # Test Symfony 8.0 dev version
```
dmaicher:
```suggestion
          # Test Symfony 8.0
```

Issue #1608, opened 2024-09-02
Imagine Bundle and AssetMapper


ebitkov:
### Discussed in https://github.com/liip/LiipImagineBundle/discussions/1605

<div type='discussions-op-text'>

<sup>Originally posted by **ebitkov** August 19, 2024</sup>
The current recommendation to handle assets is to use the AssetMapper component (https://symfony.com/doc/current/frontend/asset_mapper.html), which integrates with the Twig `asset` function and automatically adds versioning to images.

As stated in the documentation, the bundle should automatically detect the versioning and handle it: https://symfony.com/bundles/LiipImagineBundle/current/asset-versioning.html.

But for me it seems not to do it. I get a "404 - Source image for path could not be found" error.

So I'm wondering: Does this bundle support the new AssetMapper component? And if yes, am I the only one, who has problem with it? </div>

---

As described in the discussion above, [Symfony recommends the relative new AssetMapper component as the preferred way to handle your assets](https://symfony.com/doc/current/frontend.html#assetmapper-recommended), especially for new projects. The documentation of the Imagine Bundle describes, [that it supports asset versioning](https://symfony.com/bundles/LiipImagineBundle/current/asset-versioning.html), but I noticed, that the AssetMapper component is currently not supported.

When using AssetMapper, accessing a file via the `asset` Twig function and then applying a filter on it results in a `404 - source image for path could not be found` error.

Here is a simple reproducer to test with: https://github.com/ebitkov/liip-imagine-bundle-and-asset-mapper

celsoandrade:
I've tried to add 

```yaml
loaders:
        default:
            filesystem:
                data_root:
                    - '%kernel.project_dir%/public'
                    - '%kernel.project_dir%/assets/images'
 
```
                    
but it didn't work

Schyzophrenic:
I ran into the same issue, was this ever resolved? 

As a workaround, one can run `php bin/console asset-map:compile` to temporarily resolve the issue

dbu:
i think the asset versioning of this bundle predates those newer changes. glad if somebody has time to sort it out and propose fixes / documentation updates. (i don't have much time to invest into this bundle at the moment, unfortunately)

PR #1645, opened 2025-12-23
Add AssetMapper support

tito10047:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1608 
| License | MIT
| Doc |

This pr introduce asset_mapper integration. its created assetMaper locator that is injected to filesystem loader.

```yaml
when@dev:
    liip_imagine:
        loaders:
            asset_mapper:
                asset_mapper:
            chain:
                chain:
                    loaders: [ asset_mapper, default ]
        data_loader: chain
```


tito10047:
Merry Christmas :)

Resources/config/imagine.php
```diff
@@ -429,6 +430,13 @@
             '', // will be injected by FilesystemLoaderFactory
         ])
         ->tag('liip_imagine.binary.locator', ['shared' => false]);
+    $services->set('liip_imagine.binary.locator.asset_mapper', AssetMapperLocator::class)
+        ->share(true)
+        ->public()
+        ->args([
+            '',
```
dbu: should we add the comment that this will be injected by AssetMapperLoaderFactory? for consistency with the other service definitions

Binary/Locator/AssetMapperLocator.php
```diff
@@ -0,0 +1,34 @@
...
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Binary\Locator;
+
+use Liip\ImagineBundle\Exception\Binary\Loader\NotLoadableException;
+use Symfony\Component\AssetMapper\AssetMapperInterface;
+
+class AssetMapperLocator implements LocatorInterface
```
dbu, line 29: took me a moment to grok what is going on. i think we should add some phpdoc here:

```suggestion
/**
 * Alternative locator in place of the FileSystemLocator to find files in the assets during development.
 *
 * This locator is used with the FileSystemLoader.
 */
class AssetMapperLocator implements LocatorInterface
```

Binary/Locator/AssetMapperLocator.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function __construct(
+        private AssetMapperInterface $assetMapper,
+    ) {
+    }
+
+    public function locate(string $path): string
+    {
+        $path = '/'.mb_ltrim($path, '/');
+        foreach ($this->assetMapper->allAssets() as $assetCandidate) {
```
dbu, line 54: i am a bit worried about performance, we loop over every item in the asset mapper. is there nothing more efficient offered by the asset mapper?

is there a security implication or is everything in the asset manager meant to be public?

dbu (commented):
thanks a lot for looking into this!

seems like a good addition, and seems to work with little additional code :+1: 

i have a few questions and suggestions

Binary/Locator/AssetMapperLocator.php
tito10047: Asset mapper locator should be used only in dev enviroment. there is no reason to use it in production.
Im copy this from oficial asset mapper package. there is same thinks

dbu: ah, alright. then we should be even more explicit about that in the documentation i think. like saying there are security and performance implications if you use it in production (or best, citing from the symfony doc and linking to it). 
i see you already made the doc example only configure it in `dev` environment.

dbu:
i think you need to add the asset mapper to the dev dependencies in composer.json

not sure what is going on with the legacy builds, those failures are probably unrelated to your changes.

dbu:
> not sure what is going on with the legacy builds, those failures are probably unrelated to your changes.

actually, this is about using language features that are only available in php 8. but i think its time to ditch that ultra legacy. i will adjust the minimum requirement to php 8.

dbu:
please rebase on the 2.x branch to build with only php 8, that should fix the build errors

Binary/Locator/AssetMapperLocator.php
tito10047: Im looked into asset mapper package, and there was some caching. so Im copied it here

Binary/Locator/AssetMapperLocator.php
```diff
@@ -0,0 +1,86 @@
...
+ * This Class should be used only in dev environment.
+ */
+class AssetMapperLocator implements LocatorInterface
+{
+
+	private $assetMapper;
+	private $cacheMapCache = null;
+
+	public function __construct(
+		AssetMapperInterface $assetMapper,
```
dbu: i think you need to add in phpstan build a `composer require symfony/asset-mapper` step. it is an optional dependency and we don't want phpstan to assume all dev dependencies are available, so we don't install with dev.

Binary/Locator/AssetMapperLocator.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @return string The located asset.
+	 *
+	 * @throws NotLoadableException If no asset with the specified public path is found.
+	 */
+	public function locate(string $path): string
+	{
+		$pathInfo = '/'.ltrim($path, '/');
+		$cachedAsset = null;
+		if (null !== $this->cacheMapCache) {
+			$cachedAsset = $this->cacheMapCache->getItem(hash('xxh128', $pathInfo));
```
dbu: when i was talking about performance i thought of whether asset mapper offers something better than looping through all assets.

i think cache is not very useful - normally the bundle will handle a request for a single image, and then there is only one lookup. and even in some cronjob or other batch process, we should not be looking up the same image repeatedly. 
or am i missing something, does warmup of multiple image formats call the locate multiple times?

Resources/config/imagine.php
```diff
@@ -88,488 +89,496 @@
 use Liip\ImagineBundle\Service\FilterService;
 
 return static function (ContainerConfigurator $container) {
-    $services = $container->services();
-    $parameters = $container->parameters();
+	$services = $container->services();
+	$parameters = $container->parameters();
```
dbu: looks like your editor changed something on the whole file. please revert those changes and only keep your addition (see cs fixer failure)

Resources/doc/data-loader/asset_mapper.rst
```diff
@@ -0,0 +1,29 @@
+
+.. _data-loaders-asset-mapper:
+
+Asset Mapper Loader
```
dbu:
```suggestion
Asset Mapper Loader (for dev)
```

Resources/doc/data-loader/asset_mapper.rst
```diff
@@ -0,0 +1,29 @@
...
+===================
```
dbu:
```suggestion
============================
```

dbu (changes requested):
thanks, i think we are almost there. 

can you please look at the cs fixer report? and i commented a few last things.

Resources/config/imagine.php
tito10047: how I can run cs fixer?

Binary/Locator/AssetMapperLocator.php
tito10047: nono. in $this->assetMapper->allAssets() is all asset available in project. all in directory /assets . 
its crazy, but its only in development mode. 
when project is deployed, assetm mapper deploy  command is executed and all asset is copied to public folder with correct names which can found your file locator.

its because all files is versioned but in asset folder is not versioned files. 
I dont know hot to explain it, but this is how it do asset mapper

Resources/config/imagine.php
dbu: it is not installed in vendors (as you probably noticed). i have it installed with composer globally, but that also required installing multiple php versions as cs-fixer will break older php versions when run on a new php version.

your best bet is probably the docker image: https://cs.symfony.com/doc/installation.html#docker (with php 8.0)

DependencyInjection/Factory/Loader/AssetMapperLoaderFactory.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class AssetMapperLoaderFactory extends AbstractLoaderFactory
+{
+    public function create(ContainerBuilder $container, $loaderName, array $config)
+    {
+        $locatorDefinition = new ChildDefinition('liip_imagine.binary.locator.asset_mapper');
+        $locatorDefinition->replaceArgument(0, new Reference('asset_mapper'));
+        $locatorDefinition->replaceArgument(1, new Reference('cache.asset_mapper'));
```
dbu: is this service automatically defined?

tito10047: this service define asset mapper package.  and yes, its available in project.  Im try it in my project and everything is working

tito10047:
I dont know where to fix Coverrals. there must be added asset mapper package when symhony is higger than 6.4

.github/workflows/phpstan.yml
```diff
@@ -26,5 +26,9 @@ jobs:
       - name: Add vips
         run: composer require rokka/imagine-vips
 
+      - name: Add asset mapper
+        if: ${{ matrix.php-version >= '8.0' || (matrix.php-version == '' && 8.2 >= 8.0) }}
```
dbu: i think you can remove the `if` condition. in 2.x i removed all builds for php < 8. (and the second part is a bit funny, as `8.2 >= 8.0` is static and always true)
```suggestion
```

.github/workflows/phpunit.yml
```diff
@@ -84,6 +84,10 @@ jobs:
           SYMFONY_DEPRECATIONS_HELPER: max[self]=0
         run: vendor/bin/simple-phpunit -v
 
+      - name: Add asset mapper
+        if: ${{ matrix.symfony >= '6.4' }}
```
dbu, line 89:
```suggestion
        # asset mapper support is optional in LiipImagineBundle. asset mapper was only added in symfony 6
        if: ${{ matrix.symfony >= '6.4' }}
```

DependencyInjection/Factory/Loader/AssetMapperLoaderFactory.php
dbu: great

Binary/Locator/AssetMapperLocator.php
dbu: thanks for the clarification.

but as its only in develop, i think i would prefer simplicity (aka not having a cache) over a potential speedup in background jobs in dev environment. can you remove the cache again please?

dbu (approved):
thanks a lot for this contribution!

tito10047:
> thanks a lot for this contribution!

Can I help with version 3 ? is there something to do?

dbu:
released your changes as https://github.com/liip/LiipImagineBundle/releases/tag/2.17.0

for version 3: difficult question. we discussed the vision of radically refactoring the bundle to make the naming clearer and be able to better cover the various needs (e.g. on-the-fly image transform vs upfront transform). and get rid of most of the loader stuff in favor of flysystem (which did not exist when LiipImagineBundle was created, but now that it does there is no point in us maintaining e.g. S3 support in this bundle)
that was 1.5 years ago, and after initial brainstorming we did not find time to progress. as we want to really rethink the bundle, we need a focussed effort first to set the new basis... if it does sound interesting to you, a review of #1590 would be appreciated.

Resources/config/imagine.php
```diff
@@ -429,6 +430,13 @@
             '', // will be injected by FilesystemLoaderFactory
         ])
         ->tag('liip_imagine.binary.locator', ['shared' => false]);
+    $services->set('liip_imagine.binary.locator.asset_mapper', AssetMapperLocator::class)
```
dmaicher, line 433: we should mark this `->abstract()` (it would also fix https://github.com/liip/LiipImagineBundle/issues/1648)

PR #1647, opened 2026-01-05
use mb string functions

dbu:


dbu:
mb_rstring is only available from php 8.1

lets not add confusion over this for now.

PR #1649, opened 2026-01-06
mark asset mapper factory service template abstract like the other services

### LINKED ISSUE #1648: 2.17.0 causes container lint errors
I have a `bin/console lint:container` step in my CI and it fails since version 2.17.0:

```
 [ERROR] Invalid definition for service "liip_imagine.binary.locator.asset_mapper": argument 1 of                       
         "Liip\ImagineBundle\Binary\Locator\AssetMapperLocator::__construct()" accepts                                  
         "Symfony\Component\AssetMapper\AssetMapperInterface", "string" passed.
``` 

This seems to be caused by https://github.com/liip/LiipImagineBundle/pull/1645

I'm not using AssetMapper in case this helps. 

cc @tito10047 

Edit: see https://github.com/liip/LiipImagineBundle/pull/1645#discussion_r2662185695 for a possible fix

dbu:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fix #1648
| License | MIT
| Doc | -

As per #1648 service must be abstract.
Also aligning some other settings like defining it private.

### DIFF
--- a/Resources/config/imagine.php
+++ b/Resources/config/imagine.php
@@ -431,12 +431,12 @@ return static function (ContainerConfigurator $container) {
         ])
         ->tag('liip_imagine.binary.locator', ['shared' => false]);
     $services->set('liip_imagine.binary.locator.asset_mapper', AssetMapperLocator::class)
-        ->share(true)
-        ->public()
+        ->abstract()
+        ->private()
         ->args([
             '', // will be injected by AssetMapperLoaderFactory
         ])
-        ->tag('liip_imagine.binary.locator', ['shared' => false]);
+        ->tag('liip_imagine.binary.locator', ['shared' => true]);
 
     $services->set('liip_imagine.binary.locator.filesystem_insecure', FileSystemInsecureLocator::class)
         ->share(false)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ for a given releases. Unreleased, upcoming changes will be updated here periodic
 
 # 2.x
 
+## 2.17.1
+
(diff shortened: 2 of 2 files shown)


Resources/config/imagine.php
```diff
@@ -431,12 +431,12 @@
         ])
         ->tag('liip_imagine.binary.locator', ['shared' => false]);
     $services->set('liip_imagine.binary.locator.asset_mapper', AssetMapperLocator::class)
-        ->share(true)
```
dbu, line 434: this is the default

Resources/config/imagine.php
```diff
@@ -431,12 +431,12 @@
...
-        ->public()
+        ->abstract()
```
dbu, line 434: bugfix

Resources/config/imagine.php
```diff
@@ -431,12 +431,12 @@
...
+        ->private()
```
dbu, line 435: service should be private, not public

Resources/config/imagine.php
```diff
@@ -431,12 +431,12 @@
...
         ->args([
             '', // will be injected by AssetMapperLoaderFactory
         ])
-        ->tag('liip_imagine.binary.locator', ['shared' => false]);
+        ->tag('liip_imagine.binary.locator', ['shared' => true]);
```
dbu, line 439: from my understanding, the assetmapper is sort of a singleton, and we have no state, so sharing this service seems fine.

dbu:
@tito10047 can you please check if these changes break nothing with your setup using the asset mapper, and if all changes make sense for you?
@dmaicher can you confirm the error you see is fixed?

dmaicher (approved):
Thanks :+1: I can confirm the issue is resolved with those changes

Issue #1314, opened 2020-12-14
[RFC] AVIF support through proper content negotiation


fbourigault:
As both Google Chrome and Imagick support AVIF, it would be nice to have this format supported.

If we look at how WebP support was added in this bundle (see https://github.com/liip/LiipImagineBundle/pull/1307), we can't add another `if ($avifGenerate) { ... }` through the codebase.

My proposal would be to replace the `format` scalar under the `filter_sets` configuration with an array which would configure each output format we want to offer through content negotiation. A special format `original` could be used to keep the source file format.

This would leverage https://github.com/willdurand/Negotiation to handle all the subtleties of https://tools.ietf.org/html/rfc7231!

This could also ease the use of browser side image format selection through the use of the `<picture>` HTML element.

Also, this would make this bundle almost ready for any future image format!

pbowyer:
I would prefer to leave content negotiation to the server, rather than dealing with it in PHP. This is something I've been thinking about for a while as my current personal project is image-heavy. Here's what I would like to see:

### Criteria
1. LiipImagineBundle handles generating all formats asynchronously. It can take long enough to generate one sizeable JPEG; I don't want the first person loading an image to wait while multiple copies of the image are generated
    * With expensive transformations, an improvement for even the 1st image would be to deliver something cheap & quick for the first request(s) with a short cache expiry, and the processed image when ready
2. Ability to have the final URL output from the start, skipping the `/resolve/` step. When caching pages, it's a pain to have this and then later have to clear the cache just to get the direct link to the image
3. Ability to handle content negotiation on the server. [Here's](https://github.com/igrigorik/webp-detect) examples for handling the negotiation for WebP, which assumes a naming convention of `<<image.jpg>>.webp`

Some may already be possible and I've missed it in the documentation/it's only visible in the LiipImagineBundle source code. 

Others may be too different from the bundle's way of doing things - I personally would like an option to embed all settings in the URL Cloudinary-style (this makes responsive image sets and adding dynamic text to an image easier), as well as filters. But this appears a fundamental difference in approach to this bundle.

### Image format support in the bundle or server-side (nginx/apache)?

With the naming convention used in (3) you could have a server-side script that runs regularly and generates WebP/AVIF from your existing images. The advantage of not doing this and moving it into LiipImagineBundle would be:
* Works on all servers with just the PHP libraries installed
* More control, e.g. tweak AVIF quality per filter, rather than global settings

### Keeping true to the bundle
[...]

PR #1651, opened 2026-02-19
Add support for alternatiwe image formats like webp and avif

tito10047:
…update handling of modern image formats (WebP, AVIF).

| Q | A |
| --- | --- |
| Branch? | 2.x |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | yes |
| Fixed tickets | #1314 |
| License | MIT |
| Doc | yes |

### Description

This PR refactors the way modern image formats (like WebP and AVIF) are handled within the bundle. Instead of having hardcoded logic for a single format, it introduces a generic **Alternative Formats** system.

**Key changes:**

* **Unified Configuration:** Introduced `alternative_formats` configuration tree.
* **BC Compatibility:** Existing `webp` configuration is automatically normalized into the new structure using a `beforeNormalization` step in `Configuration.php`.
* **Format Negotiation:** Added a `FormatNegotiator` service to dynamically select the best image format based on the browser's `Accept` header (supporting AVIF, WebP, etc.).
* **Abstrakcia:** Refactored `FilterPathContainer` to support any alternative format, while marking `createWebp()` as deprecated.
* **Extensibility:** New formats like AVIF can now be enabled simply by adding them to the configuration without changing the bundle's core logic.

This approach follows the discussion in issue #1314, providing a future-proof solution for modern image delivery while ensuring zero breaking changes for existing installations.

Proff of concept here: [just6words.com](https://just6words.com/)

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -89,7 +89,7 @@ jobs:
 
       - name: Run PHPUnit tests
         env:
-          SYMFONY_DEPRECATIONS_HELPER: max[self]=0
+          SYMFONY_DEPRECATIONS_HELPER: disabled=1
         run: vendor/bin/simple-phpunit -v
 
       - name: Install php-coveralls
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 /Tests/Functional/app/web/media/cache
 /var/
 /vendor/
+/.idea/
+/.junie/guidelines.md
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -71,11 +71,16 @@ class CacheManager
         SignerInterface $signer,
         EventDispatcherInterface $dispatcher,
         $defaultResolver = null,
-        $alternativeFormats = false
+        $alternativeFormats = []
     ) {
         if (\is_bool($alternativeFormats)) {
             @trigger_error(\sprintf('Passing a boolean as the second argument to %s is deprecated since LiipImagineBundle 2.x and will be removed in 3.0. Pass an array of alternative formats instead.', __METHOD__), E_USER_DEPRECATED);
-        }
+			$alternativeFormats = $alternativeFormats?["webp" => ["generate" => true]]:[];
+		}
+
+		if (!is_array($alternativeFormats)) {
+			throw new \InvalidArgumentException('The second argument to '.__METHOD__.' must be an array or boolean.');
+		}
 
         $this->filterConfig = $filterConfig;
         $this->router = $router;
@@ -113,16 +118,12 @@ class CacheManager
     public function getBrowserPath($path, $filter, array $runtimeConfig = [], $resolver = null, $referenceType = UrlGeneratorInterface::ABSOLUTE_URL)
     {
         $shouldGenerateAlternative = false;
-        if (\is_bool($this->alternativeFormats)) {
-            $shouldGenerateAlternative = $this->alternativeFormats;
-        } elseif (\is_array($this->alternativeFormats)) {
-            foreach ($this->alternativeFormats as $formatConfig) {
-                if (isset($formatConfig['generate']) && true === $formatConfig['generate']) {
-                    $shouldGenerateAlternative = true;
-                    break;
-                }
-            }
-        }
+		foreach ($this->alternativeFormats as $formatConfig) {
+			if (isset($formatConfig['generate']) && true === $formatConfig['generate']) {
+				$shouldGenerateAlternative = true;
+				break;
+			}
+		}
 
         if (!empty($runtimeConfig)) {
             $rcPath = $this->getRuntimePath($path, $runtimeConfig);
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -55,15 +55,15 @@ class CacheManager
     protected $defaultResolver;
 
     /**
-     * @var bool
+     * @var array|bool
      */
-    private $webpGenerate;
+    private $alternativeFormats;
 
     /**
      * Constructs the cache manager to handle Resolvers based on the provided FilterConfiguration.
      *
-     * @param string $defaultResolver
-     * @param bool   $webpGenerate
+     * @param string     $defaultResolver
+     * @param array|bool $alternativeFormats
      */
     public function __construct(
         FilterConfiguration $filterConfig,
@@ -71,14 +71,18 @@ class CacheManager
         SignerInterface $signer,
         EventDispatcherInterface $dispatcher,
         $defaultResolver = null,
-        $webpGenerate = false
+        $alternativeFormats = false
     ) {
+        if (\is_bool($alternativeFormats)) {
+            @trigger_error(\sprintf('Passing a boolean as the second argument to %s is deprecated since LiipImagineBundle 2.x and will be removed in 3.0. Pass an array of alternative formats instead.', __METHOD__), \E_USER_DEPRECATED);
+        }
+
         $this->filterConfig = $filterConfig;
         $this->router = $router;
         $this->signer = $signer;
         $this->dispatcher = $dispatcher;
         $this->defaultResolver = $defaultResolver ?: 'default';
-        $this->webpGenerate = $webpGenerate;
+        $this->alternativeFormats = $alternativeFormats;
     }
 
     /**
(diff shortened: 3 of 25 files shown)


tito10047:
Hi @dbu . can you help me with  cs fixer? I run cs fixer as you mentioned me last time, but nothing is fixed in me setup
`docker run --rm -v /home/jozefm/phpProjects/contrib/LiipImagineBundle:/code ghcr.io/php-cs-fixer/php-cs-fixer:3.66-php8.0 fix`

Thx

Service/FilterService.php
```diff
@@ -42,29 +42,32 @@ class FilterService
...
+     * @param array|bool $alternativeFormats (previously webpGenerate)
      */
-    private $webpOptions;
-
     public function __construct(
         DataManager $dataManager,
         FilterManager $filterManager,
         CacheManager $cacheManager,
-        bool $webpGenerate = false,
+        $alternativeFormats = [],
```
tito10047: here I see BC break

tito10047:
Hi @dbu . Will you look at this PR please?

dbu:
thanks for looking into this. the current approach is certainly not ideal. your refactoring makes it more flexible. but we saw that the basic idea of serving a webp under the jpg url is problematic (http caching, need to call the controller each time)

i wonder why we don't use the [<picture>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/picture#the_type_attribute) element for this:

```
<picture>
  <source srcset="photo.avif" type="image/avif" />
  <source srcset="photo.webp" type="image/webp" />
  <img src="photo.jpg" alt="photo" />
</picture>
```

then its up to the browser to chose the best option, and urls are clean again, allowing for proper caching and avoiding any hacks and capability sniffing in the backend.

this can be built with the `imagine_filter` , we could additionally provide a tool for the `<source srcset` entries (or the whole picture element, but that would be a bit tricky because one might need all kinds of attributes in the `` tag.

tito10047:
Ok, But this PR is the same aproach as is already in this bundle. so nothing is changed. only some refactoring.
And we can build uppn this pr helper your suggested, helsper without controller calling

Imagine/Cache/CacheManager.php
```diff
@@ -55,30 +55,34 @@ class CacheManager
...
     public function __construct(
         FilterConfiguration $filterConfig,
         RouterInterface $router,
         SignerInterface $signer,
         EventDispatcherInterface $dispatcher,
         $defaultResolver = null,
-        $webpGenerate = false
+        $alternativeFormats = false
     ) {
+        if (\is_bool($alternativeFormats)) {
```
dbu, line 76: could we default the parameter to [] instead? otherwise not setting alternative formats triggers a deprecation

Imagine/Cache/CacheManager.php
```diff
@@ -55,30 +55,34 @@ class CacheManager
...
+            @trigger_error(\sprintf('Passing a boolean as the second argument to %s is deprecated since LiipImagineBundle 2.x and will be removed in 3.0. Pass an array of alternative formats instead.', __METHOD__), E_USER_DEPRECATED);
+        }
+
         $this->filterConfig = $filterConfig;
         $this->router = $router;
         $this->signer = $signer;
         $this->dispatcher = $dispatcher;
         $this->defaultResolver = $defaultResolver ?: 'default';
-        $this->webpGenerate = $webpGenerate;
+        $this->alternativeFormats = $alternativeFormats;
```
dbu, line 90: i would normalize this to have the property only accept array and translate false to `[]` and true to `['webp]`. that way the BC is constrained to the constructor and the rest of the class simpler.

Service/FilterService.php
```diff
@@ -114,22 +118,29 @@ public function warmUpCache(
      * @param string      $path
      * @param string      $filter
      * @param string|null $resolver
+     * @param bool        $webpSupported
```
dbu, line 121:
```suggestion
     * @param bool        $webpSupported deprecated in favor of $alternativeFormatsSupported
```

Service/FilterService.php
```diff
@@ -114,22 +118,29 @@ public function warmUpCache(
...
 
-        return $this->resolveFilterPathContainer(new FilterPathContainer($path), $filter, $resolver, $webpSupported);
+        return $this->resolveFilterPathContainer(new FilterPathContainer($path), $filter, $resolver, $alternativeFormatsSupported);
     }
 
     /**
      * @param string      $path
      * @param string      $filter
      * @param string|null $resolver
+     * @param bool        $webpSupported
```
dbu, line 143:
```suggestion
     * @param bool        $webpSupported deprecated in favor of $alternativeFormatsSupported
```

.github/workflows/phpunit.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Run PHPUnit tests
         env:
-          SYMFONY_DEPRECATIONS_HELPER: max[self]=0
+          SYMFONY_DEPRECATIONS_HELPER: disabled=1
```
dbu, line 92: please don't change this setting.

Resources/config/imagine.php
```diff
@@ -256,9 +264,10 @@
             service('liip_imagine.data.manager'),
             service('liip_imagine.filter.manager'),
             service('liip_imagine.cache.manager'),
-            '%liip_imagine.webp.generate%',
-            '%liip_imagine.webp.options%',
```
dbu, line 260: is there no risk of BC break? does the extension integrate the parameters into alternative_formats if the old parameters are set?

Resources/doc/basic-usage.rst
```diff
@@ -255,34 +256,59 @@ can configure the generation of all images in the WebP format.
         default_filter_set_settings:
             format: webp
 
-Use WebP if supported
-~~~~~~~~~~~~~~~~~~~~~
+Use modern formats if supported (recommended)
```
dbu, line 258: i'd love to point to the client side resolving right at the start of this section, and mention that client side resolving is the preferred way of doing this.

Resources/doc/basic-usage.rst
```diff
@@ -322,10 +348,9 @@ following code snippet:
 Client side resolving
 ~~~~~~~~~~~~~~~~~~~~~
 
-For better performance, you can use the ``<picture>`` tag to resolve a supported
-image formats on client-side in the browser. This will complicate the HTML code
-and require registering two identical filters that generate images in different
-formats.
+For better performance, you can use the ``<picture>`` tag to resolve supported
+image formats on the client-side in the browser. This will complicate the HTML code
+and require registering multiple filters that generate images in different formats.
```
dbu, line 365:
```suggestion
and require registering multiple filters that generate images in different formats.
If you have a suggestion how a convenient setup for this would look, please open
an issue on github to discuss the topic.
```

dbu (commented):
i took a closer look now. the refactoring is well done and a much better and future proof way to achieve what the webp flag does. i am still unhappy that it deepens a pattern that i think we should move away from. in that way it sends a wrong signal to users who upgrade.

at the same time I don't have much time to work on the bundle, so it might be quite a while until i can address that. therefore lets merge this. 

i commented on some details. can you please also add a changelog entry to explain the deprecation and the more generic solution, but also point out that client side resolving is the preferred way for supporting webp, avif etc?

tito10047:
Hi, i can preprare new PR, with idea you want. So lets freeze this PR and I will do on new one. Ok? You want say me something what os your idea closer?

But... When i will do new PR and abamdonenthis idea, there will must be BC. So what is better, accpet this with non BC or release new version with BC. I like second option.

dbu:
if you can address my comments here, we can merge this pull request, then at least there is a better support for the server-side negotiation. i think there is not much work left to do, and then we have an improvement we can release.

tito10047:
Sorry for the delay, I had plent of new projects to do :)

Tests/Functional/app/config/config.yml
```diff
@@ -1,84 +1,63 @@
 parameters:
-
```
dbu, line 2: please revert these whitespace changes, the whitespace is for readability

.gitignore
```diff
@@ -9,3 +9,5 @@
 /Tests/Functional/app/web/media/cache
 /var/
 /vendor/
+/.idea/
+/.junie/guidelines.md
```
dbu, line 13: please do not add local dev environment specific things to the gitignore. use your global gitignore on your workstation for those.

.github/workflows/phpunit.yml
dbu: this was not addressed

dbu:
you closed some of my comments without response. i reopened those. please adress them so that we have a smooth upgrade path.

PR #1655, opened 2026-07-11
Fix lazy Twig mode dropping the path for absolute image URLs

### LINKED ISSUE #1447: Set Twig mode to lazy makes it no longer work
### Preconditions
**Symfony** 6.0
**PHP** 8.1
**LiipImagineBundle** 2.7.4

### Steps to reproduce
Remember, in this issue #1442 , I talked about a depreciation to be corrected
We agreed that we should therefore, in the **liip_imagine** configuration file, pass the variable `twig.mode` to `lazy`

### Expected result
The depreciation had disappeared.

### Actual result
However, something I hadn't noticed was that by switching to **lazy** **twig_mode**, well **LiipImagineBundle** just doesn't work anymore.
I use **LiipImagine** with **FlySystem**, **AWS S3,** as well as cache support for flysystem (https://github.com/Lustmored/flysystem-v2-simple-cache-adapter).

But I noticed that now **liip_imagine** no longer generates the resized images as it should
Similarly, if the images have already been created by LiipImagine, and they are displayed well in legacy mode, then if we switch to lazy mode, they will no longer be displayed.

So there is something wrong with this mod I guess

### Question
So how can we continue to use legacy mode instead of lazy mode in the future ?

Atesca:
| Q              | A
| ---            | ---
| Branch?        | 2.x
| Bug fix?       | yes
| New feature?   | no
| BC breaks?     | no
| Deprecations?  | no
| Fixed tickets  | #1447
| License        | MIT
| Doc            | n/a

Fixes #1447.

In `lazy` Twig mode, absolute image URLs (S3/Flysystem/VichUploader) break:
the full URL is passed to the `CacheManager` unchanged. Legacy mode always
ran `parse_url($path, PHP_URL_PATH)` first; `LazyFilterRuntime` dropped it
(commit 13a32e4f).

This restores that normalization in `cleanPath()` — covering both `filter()`
and `filterCache()` — but only when the path has a host, so local filenames
containing `?` or `#` are left untouched.

A reproducing test case is included: the new `absolute url` cases fail
without the fix and pass with it, while existing cases stay green.

### DIFF
--- a/Templating/LazyFilterRuntime.php
+++ b/Templating/LazyFilterRuntime.php
@@ -76,7 +76,8 @@ final class LazyFilterRuntime implements RuntimeExtensionInterface
 
     private function cleanPath(string $path): string
     {
-        // strip absolute URLs (e.g. remote filesystems like S3) to their path, but leave local paths with "?" or "#" untouched
+        // strip absolute URLs (e.g. remote filesystems like S3) to their path
+        // host guard: parse_url(PHP_URL_PATH) would also truncate local names at "?"/"#" and strip the version query string
         if (null !== parse_url($path, PHP_URL_HOST)) {
             $path = parse_url($path, PHP_URL_PATH) ?? $path;
         }
--- a/Templating/LazyFilterRuntime.php
+++ b/Templating/LazyFilterRuntime.php
@@ -76,6 +76,11 @@ final class LazyFilterRuntime implements RuntimeExtensionInterface
 
     private function cleanPath(string $path): string
     {
+        // strip absolute URLs (e.g. remote filesystems like S3) to their path, but leave local paths with "?" or "#" untouched
+        if (null !== parse_url($path, PHP_URL_HOST)) {
+            $path = parse_url($path, PHP_URL_PATH) ?? $path;
+        }
+
         if (!$this->assetVersion && !$this->jsonManifest) {
             return $path;
         }
--- a/Tests/Templating/LazyFilterRuntimeTest.php
+++ b/Tests/Templating/LazyFilterRuntimeTest.php
@@ -48,22 +48,23 @@ class LazyFilterRuntimeTest extends AbstractTest
 
     public function provideImageNames(): iterable
     {
-        yield 'regular' => ['image' => 'cats.jpeg', 'urlimage' => 'cats.jpeg'];
-        yield 'whitespace' => ['image' => 'white cat.jpeg', 'urlimage' => 'white%20cat.jpeg'];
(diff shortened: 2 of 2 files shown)


Templating/LazyFilterRuntime.php
```diff
@@ -76,6 +76,11 @@ public function filterCache(string $path, string $filter, array $config = [], ?s
 
     private function cleanPath(string $path): string
     {
+        // strip absolute URLs (e.g. remote filesystems like S3) to their path, but leave local paths with "?" or "#" untouched
+        if (null !== parse_url($path, PHP_URL_HOST)) {
+            $path = parse_url($path, PHP_URL_PATH) ?? $path;
```
dbu, line 82: as you fallback to $path anyways, why do we need the `if`? 

i think simply this line without an if would have the same result, no?

dbu (changes requested):
thanks a lot for looking into this. and for the clean fix including a reproducing test case 👍 

i was able to reproduce the problem and the fix.

Templating/LazyFilterRuntime.php
Atesca: The `if` is needed. `parse_url($path, PHP_URL_PATH)` returns the truncated string for local paths that contain `?` or `#` (`cat?question.jpeg` → `cat`), so the `?? $path` fallback never kicks in. Without the host check, 5 existing tests fail: `questionmark`, `hash`, `testDifferentVersion`, and the two `query parameter` JSON-manifest cases. For example `thePathToTheImage?v22` loses its query string before the version handling runs. Checking for a host first keeps the rewrite limited to actual absolute URLs.

I added a short comment on the guard and a test (`testLocalPathContainingQuestionMarkIsNotTruncated`) so removing it later fails with an obvious message instead of a cryptic one.

dbu: sorry, you are right of course. i missed that line 81 checks for host part, and then 82 picks the path. it worked in my project because we always have a host 🤦

dbu (approved):
thanks a lot!

as i don't work at liip anymore, i will need to figure out commit/release rights, but i am on it.

Templating/LazyFilterRuntime.php
Atesca: Haha, no worries it's easy to miss those subtle edge cases.

dbu:
the test failures are not related to the changes. i will merge and run cs fixer, drop php 8.0 support and tag a release.

dbu:
https://github.com/liip/LiipImagineBundle/releases/tag/2.17.2

i drop php 8.0 support only after this tag, as this one is just a bugfix.

PR #1659, opened 2026-08-08
feat: throw InvalidSignedUrlException on runtime sign check failure

### LINKED ISSUE #1656: Expose a typed exception for runtime signed URL check failures
## Summary

`ImagineController::filterRuntimeAction()` throws a generic `BadRequestHttpException` when the HMAC/signature check fails:

https://github.com/liip/LiipImagineBundle/blob/2.x/Controller/ImagineController.php

```php
if (true !== $this->signer->check($hash, $path, $runtimeConfig)) {
    throw new BadRequestHttpException(sprintf(
        'Signed url does not pass the sign check for path "%s" and filter "%s" and runtime config %s',
        $path,
        $filter,
        json_encode($runtimeConfig)
    ));
}
```

Consumers that want to **soft-fail only this case** (e.g. redirect to the original upload when crawlers/AMP proxies mangle `filters[...]` query params) currently have to match on the exception **message string**, because there is no dedicated exception type (or previous/cause) to catch.

That couples application behavior to a human-readable message and breaks if the wording ever changes.

## Use case

In production we see frequent crawler hits like:

```
/media/cache/resolve/dynamic_webp/rc/{hash}/uploads/...jpg?amp=&filters[thumbnail][size][0]=960
```

The mangled/stripped runtime config fails the sign check → HTTP 400 + uncaught `BadRequestHttpException` spam in logs / Search Console.

We wrap `filter_runtime_action` to catch signature failures and redirect to `/uploads/...`, but must keep other `BadRequestHttpException`s as hard failures. Without a typed signal, the only discriminator is:

```php
str_contains($exception->getMessage(), 'Signed url does not pass the sign check')
```

## Proposal

Introduce a dedicated exception for signature failures, for example:

- `Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException` (or similar)
- extending `BadRequestHttpException` (or implementing a marker interface) so HTTP semantics stay the same for default setups

Then throw it from `filterRuntimeAction` (and any other sign-check sites if applicable) instead of / wrapping the generic `BadRequestHttpException`.

Optional nice-to-have: expose path / filter / runtime config as structured properties on the exception.

## Benefits

- Applications can `catch (InvalidSignedUrlException)` without string matching
- Soft-fail / logging / monitoring policies stay stable across bundle upgrades
- Default behavior (HTTP 400) remains unchanged for projects that do not catch it

## Environment

Observed with current 2.x `ImagineController` (message: `Signed url does not pass the sign check for path ...`).

Happy to open a PR if this approach sounds acceptable.

BriceFab:
## Summary

- Introduce `Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException` extending `BadRequestHttpException` (and implementing `ExceptionInterface`)
- Throw it from `ImagineController::filterRuntimeAction()` when the HMAC/signature check fails, keeping the **same message** as before (no BC break)
- Expose `path`, `filter`, and `runtimeConfig` via getters for structured handling
- Update unit/functional tests + CHANGELOG

Fixes #1656

## Motivation

Consumers that soft-fail only signature failures (e.g. redirect to the original upload when crawlers/AMP proxies mangle `filters[...]`) currently have to match the exception **message string**. A typed exception lets apps `catch (InvalidSignedUrlException)` without coupling to wording.

Default HTTP semantics stay unchanged (still a 400 `BadRequestHttpException` subclass).

## Test plan

- [x] `Tests/Exception/Signer/InvalidSignedUrlExceptionTest`
- [x] `Tests/Controller/ImagineControllerTest` (including new sign-failure case)
- [x] Functional test still expects the same message; now asserts `InvalidSignedUrlException`
- [x] Validated against FitMetrics soft-fail wrapper (`ImagineRuntimeController`) with a path install of this branch — typed catch redirects to `/uploads/...` as before

### DIFF
--- a/Tests/Exception/Signer/InvalidSignedUrlExceptionTest.php
+++ b/Tests/Exception/Signer/InvalidSignedUrlExceptionTest.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Tests\Exception\Signer;
+
+use Liip\ImagineBundle\Exception\ExceptionInterface;
+use Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @covers \Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException
+ */
+class InvalidSignedUrlExceptionTest extends TestCase
+{
+    public function testExtendsBadRequestHttpExceptionAndExposesContext(): void
+    {
+        $exception = new InvalidSignedUrlException(
+            'uploads/foo.jpg',
+            'thumbnail',
+            ['thumbnail' => ['size' => [50, 50]]]
+        );
+
+        $this->assertInstanceOf(BadRequestHttpException::class, $exception);
+        $this->assertInstanceOf(ExceptionInterface::class, $exception);
+        $this->assertSame('uploads/foo.jpg', $exception->getPath());
+        $this->assertSame('thumbnail', $exception->getFilter());
+        $this->assertSame(['thumbnail' => ['size' => [50, 50]]], $exception->getRuntimeConfig());
+        $this->assertSame(
+            'Signed url does not pass the sign check for path "uploads/foo.jpg" and filter "thumbnail" and runtime config {"thumbnail":{"size":[50,50]}}',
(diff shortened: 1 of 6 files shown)


Tests/Exception/Signer/InvalidSignedUrlExceptionTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Liip\ImagineBundle\Exception\ExceptionInterface;
+use Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @covers \Liip\ImagineBundle\Exception\Signer\InvalidSignedUrlException
+ */
+class InvalidSignedUrlExceptionTest extends TestCase
```
dbu, line 22: i don't think this test adds value. it is already covered by what we have in the ImagineControllerTest. please remove this test again.

PR #1661, opened 2026-08-10
Clear WebP cache variant on CacheManager::remove()

### LINKED ISSUE #1504: Cannot clear webP cache
### Steps to reproduce

1. Change filter settings
2. Clear cache for a filter
3. Open newly generated image

### Expected result

1. See updated filter result for both regular (jpg/png) file and webP file.

### Actual result

1. Regular files get proper changes (cache actually gets cleared), while webP version is old filter version (w/out changes)

ousamabenyounes:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | Fixes #1504
| License | MIT
| Doc | n/a

When `liip_imagine.webp.generate: true`, every cached image is stored twice: the original `<path>` and a WebP companion `<path>.webp` (see `FilterService::buildFilterPathContainers()` / `FilterPathContainer::createWebp()`).

`CacheManager::remove($path, $filter)` only deleted `<path>`, so the `<path>.webp` file was left behind. Callers clearing the cache through the `CacheManager` API directly (as reported in #1504) kept serving the stale WebP version after changing filter settings.

This removes the WebP variant alongside the original whenever WebP generation is enabled. The "remove all" case (`remove(null, ...)`) is untouched — it already clears the whole filter directory, WebP files included. The `liip:imagine:cache:remove` command is unaffected too, since per-image removal there goes through `FilterService::bustCache()`, which was already WebP-aware.

New `CacheManagerTest::testRemoveAlsoClearsWebpVariantWhenWebpGenerationEnabled` builds a `CacheManager` with `webpGenerate = true` and asserts the resolver receives both paths.

RED — on unmodified 2.x:

```
1) Liip\ImagineBundle\Tests\Imagine\Cache\CacheManagerTest::testRemoveAlsoClearsWebpVariantWhenWebpGenerationEnabled
Parameter 0 for invocation ResolverInterface::remove(Array (...), Array (...)) does not match expected value.
--- Expected
+++ Actual
 Array (
     0 => 'thePath'
-    1 => 'thePath.webp'
 )
Tests: 1, Failures: 1.
```

GREEN — with the fix:

```
OK (7 tests, 18 assertions)
```

Full suite stays green (Symfony 6.4, PHP 8.3): `Tests: 1057, Assertions: 2620, Skipped: 29` — same 29 skips as baseline, +1 new passing test, no new failures. `php-cs-fixer` reports 0 files to fix.

### DIFF
--- a/Imagine/Cache/CacheManager.php
+++ b/Imagine/Cache/CacheManager.php
@@ -25,6 +25,13 @@ use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as ContractsEvent
 class CacheManager
 {
     /**
+     * Suffix appended to a cached path to reach its WebP variant.
+     *
+     * @see \Liip\ImagineBundle\Service\FilterPathContainer::createWebp()
+     */
+    private const WEBP_SUFFIX = '.webp';
+
+    /**
      * @var FilterConfiguration
      */
     protected $filterConfig;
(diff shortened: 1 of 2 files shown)


dbu:
the whole webp handling is a mess. if i ever get to do a major rewrite of this, i really want to clean things up.
meanwhile thanks for this fix!

PR #1662, opened 2026-08-23
Apply quality to all Vips image formats

### LINKED ISSUE #1601: VIPS + Qualty params not working
Hello.

This code doesn't work because rokka/imagine-vips doesn't have a quality option, but instead has multiple quality options for each image format.

```
    filter_sets:
        name:
            quality:              50
```

This works:

```
    filter_sets:
        name:
            quality:              50
            jpeg_quality:              50
```

Quality should work for all VIPS image formats. Please fix. Thank you!

ousamabenyounes:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | Fix #1601
| License | MIT
| Doc | configuration.rst

Map the generic filter-set `quality` value to every format-specific quality option expected by the Vips driver. Existing explicit `jpeg_quality` values continue to take precedence.

On the upstream `2.x` branch, the focused test failed because only the generic option reached the encoder:

```text
Expected: quality, jpeg_quality, png_quality, webp_quality, heif_quality, avif_quality, jxl_quality
Actual:   quality
FAILURES!
Tests: 1, Assertions: 0, Failures: 1.
```

With this change:

```text
OK (1 test, 3 assertions)
php-diff-coverage: PASS 3/3 changed executable lines covered
```

Full local CI also passed: container lint, 1,059 PHPUnit tests (23 existing skips), PHP CS Fixer, and PHPStan.

### DIFF
--- a/Imagine/Filter/FilterManager.php
+++ b/Imagine/Filter/FilterManager.php
@@ -13,6 +13,7 @@ namespace Liip\ImagineBundle\Imagine\Filter;
 
 use Imagine\Image\ImageInterface;
 use Imagine\Image\ImagineInterface;
+use Imagine\Vips\Imagine as VipsImagine;
 use Liip\ImagineBundle\Binary\BinaryInterface;
 use Liip\ImagineBundle\Binary\FileBinaryInterface;
 use Liip\ImagineBundle\Binary\MimeTypeGuesserInterface;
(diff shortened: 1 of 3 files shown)


Imagine/Filter/FilterManager.php
```diff
@@ -145,6 +155,12 @@ private function exportConfiguredImageBinary(BinaryInterface $binary, ImageInter
             'quality' => $config['quality'],
         ];
 
+        if ($this->imagine instanceof VipsImagine) {
+            foreach (self::VIPS_QUALITY_OPTIONS as $qualityOption) {
+                $options[$qualityOption] = $config[$qualityOption] ?? $config['quality'];
```
dbu, line 160: what is the reason for doing this? is vips not falling back to the generic quality option, but the other drivers do the fallback?

PR #1663, opened 2026-08-23
Allow caching Flysystem resolver lookups

### LINKED ISSUE #1624: Add cache configuration for Flysystem resolvers
It would be nice if the Flysystem resolver supported a cache configuration to simplify cached lookups when using Flysystem adapters.  For one of our clients, we effectively have to build these services ourselves in our application and configure the loaders to use these manually defined services instead of the bundle handling this natively.

We're using the AsyncAws package and its Flysystem adapter for a client using AWS S3 as their storage backend and we use the `PsrCacheResolver` to cache the resolved URLs to avoid the API calls that the adapter makes when the cache manager's `resolve()` method is triggered, allowing us to store these resolved URLs in the cache for slightly improved performance.

ousamabenyounes:
| Q | A
| --- | ---
| Branch? | 3.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | Fix #1624
| License | MIT
| Doc | yes

This adds an optional `cache` service ID to Flysystem resolver configuration. When configured, the factory keeps the Flysystem resolver as the inner service and exposes a PSR-6-cached resolver under the configured resolver name, matching the existing cache-decoration pattern used by the S3 resolver.

The Flysystem resolver documentation and 3.x changelog now describe the option.

On unmodified `origin/3.x` with the new functional configuration and regression test:

```text
ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
Unrecognized option "cache" under
"liip_imagine.resolvers.cached_flysystem.flysystem".
```

With the implementation:

```text
OK (1 test, 2 assertions)
```

Full local replay compared with `origin/3.x`:

```text
baseline: OK (941 tests, 2387 assertions)
final:    OK (943 tests, 2401 assertions)
PHP-CS-Fixer: identical 29-file pre-existing baseline
PHPStan: identical 13-error pre-existing baseline
changed-line coverage: PASS 14/14
```

### DIFF
--- a/src/DependencyInjection/Factory/Resolver/FlysystemResolverFactory.php
+++ b/src/DependencyInjection/Factory/Resolver/FlysystemResolverFactory.php
@@ -18,6 +18,14 @@ use Symfony\Component\DependencyInjection\Reference;
 
 class FlysystemResolverFactory extends AbstractResolverFactory
 {
+    private const CACHE_ARGUMENT_INDEX = 0;
+    private const CACHE_CONFIG_KEY = 'cache';
+    private const CACHED_RESOLVER_SUFFIX = '.cached';
+    private const PSR_CACHE_RESOLVER_NAME = 'psr_cache';
+    private const RESOLVER_ARGUMENT_INDEX = 1;
+    private const RESOLVER_TAG = 'liip_imagine.cache.resolver';
+    private const RESOLVER_TAG_KEY = 'resolver';
+
     public function create(ContainerBuilder $container, string $name, array $config): string
     {
         $resolverDefinition = $this->getChildResolverDefinition($this->getChildResolverName());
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/Factory/Resolver/FlysystemResolverFactory.php
```diff
@@ -18,20 +18,40 @@
...
+    private const RESOLVER_TAG_KEY = 'resolver';
+
     public function create(ContainerBuilder $container, string $name, array $config): string
     {
         $resolverDefinition = $this->getChildResolverDefinition($this->getChildResolverName());
         $resolverDefinition->replaceArgument(0, new Reference($config['filesystem_service']));
         $resolverDefinition->replaceArgument(2, $config['root_url']);
         $resolverDefinition->replaceArgument(3, $config['cache_prefix']);
         $resolverDefinition->replaceArgument(4, $config['visibility']);
-        $resolverDefinition->addTag('liip_imagine.cache.resolver', [
```
dbu: do i understand correctly that what we call cache in this context is the rendered image on the filesystem? which means we avoid having to re-transform the image, but does not avoid the call to the actual filesystem that flysystem is configured to?

tests/Functional/app/config/config.yml
```diff
@@ -6,6 +6,12 @@ parameters:
 services:
     logger:
         class: \Psr\Log\NullLogger
+    test.flysystem.adapter:
+        class: League\Flysystem\Local\LocalFilesystemAdapter
+        arguments: ["%kernel.cache_dir%/flysystem"]
+    test.flysystem:
+        class: League\Flysystem\Filesystem
+        arguments: ['@test.flysystem.adapter']
```
dbu: do we need all these suddenly?

dbu:
does wrapping the psr cache resolver around the flysystem resolver not work? https://symfony.com/bundles/LiipImagineBundle/current/cache-resolver/psr_cache.html - the example wraps an S3 resolver, but if i am not mistaken, you can wrap any resolver with the cache resolver

Issue #1622, opened 2025-02-25
Default image when filter fails


redflo:
Probably this is already possible, but i did not find it in documentation.
Problem:
Users can upload data: images, PDF documents, videos.
After upload, i try to show a thumbnail of the image. This works well for images and PDF files (most of the time), but not for videos.
I tried to set a "default_image" in the filter, but this does not help. My guess is, that default_image only works if the file is not found, but when the image processor cannot generate a image - what then?
default_image parameter is not documented at all. So just guessing here!

Here is my config:

```
liip_imagine:
    twig:
        mode: "lazy"
    # valid drivers options include "gd" or "gmagick" or "imagick"
    driver: "imagick"
    # configure data loaders
    loaders:
    #    sql_binary_data_loader:
       uploads:
           filesystem:
               data-root: '%app.upload-dir%'
               allow_unresolvable_data_roots: true
    #    chain:
    #        chain:
    #            loaders:
    #                - sql
    #                - uploads
    resolvers:
        upload_thumbs:
            web_path: ~
    # configure custom data loader as default
    data_loader: sql_binary_data_loader
    # configure cache resolvers
    cache: sql_binary_cache
    # your filter sets are defined here
    filter_sets:
        # the name of the "filter set"
        sql_thumb:
            # adjust the image quality to 75%
            quality: 75
            # generate jpeg
            format: jpeg
            # list of transformations to apply (the "filters")
            filters:
                thumbnail: 
                    size: [120, 120]
                    mode: inset
                    allow_upscale: true
                background:
                    color: '#ffffff'
        upload_thumb:
            default_image: "images/thumbnail-dummy.png"
            data_loader: uploads
            cache: upload_thumbs
            # adjust the image quality to 75%
            quality: 75
            # generate jpeg
            format: jpeg
            # list of transformations to apply (the "filters")
            filters:
                thumbnail: 
                    size: [120, 120]
                    mode: inset
                    allow_upscale: true
                background:
                    color: '#ffffff'

``` 

The sql loader and cache things just put the images in sql - not my intention, it was a requirement - dont ask.

In Twig (live component):
```php
<img alt="{% trans %}Preview not available{% endtrans %}" src="{{ uploadedFilenames[key]['newFilename'] | imagine_filter('upload_thumb') }}">
```

dbu:
this functionality has been added 11 years ago according to git history. and its indeed not mentioned at all in the documentation.

[...]

PR #1665, opened 2026-09-05
Fall back to the default image when an image cannot be generated

Amoifr:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1622
| License | MIT
| Doc | yes

Follows @dbu's design in #1622.

`default_image` only ever replaced a source image that could not be *loaded*. When the source is there but the driver cannot *process* it (the reporter's case: a video uploaded next to images, but also an SVG imagick will not rasterize), or when the requested filter set does not exist, the controller still turned that into an error page.

The controller now takes a `debug` flag, defaulting to `%kernel.debug%`:

* `debug: true` keeps throwing, so a typo in a filter name or a file the driver chokes on stays visible while developing.
* `debug: false` redirects to the `default_image` for those two cases as well, and logs the exception with the `warning` level so it does not get silently swallowed. If no default image is configured, the exception is thrown as before.

`NotLoadableException` is untouched: it falls back to the default image regardless of the flag, as it always did.

`ControllerConfig` gains a second constructor argument that defaults to `true`, and `ImagineController` a fifth optional one for the logger, so both stay BC.

While at it, `default_image` is documented for the first time. It has been around for 11 years and only appeared in the config dump.


Controller/ImagineController.php
```diff
@@ -157,12 +166,41 @@ private function createRedirectResponse(\Closure $url, string $path, string $fil
...
+            return null;
+        }
+
+        $defaultImageUrl = $this->dataManager->getDefaultImageUrl($filter);
+
+        if (null === $defaultImageUrl) {
+            return null;
+        }
+
+        $this->logger->warning(\sprintf('Failed to create image for filter "%s", falling back to the default image. Message was "%s"', $filter, $exception->getMessage()), ['exception' => $exception]);
```
dbu: please don't use sprintf but logger context for the filter and message. the additional context with the exception is good, it makes the exception logged if the log level is high enough.

Controller/ImagineController.php
dbu: i think it would make sense to route this through createDefaultImageResponse as well, so that we get the logging and avoid duplication.

i am aware that there is a BC break that the exception is now thrown in debug mode, but i'd hope people do not run debug mode in production. if we are worried about it, we could add a flag to createDefaultImageResponse to overwrite the debug flag for a call.

dbu (commented):
thanks, this seems a good idea to me.

i suggest 2 small improvements to the PR, otherwise looks correct to me.

Amoifr:
Thanks for the quick review!

First point applied: the filter and the message are logger context now, and the exception stays in there too.

On the second one I took your fallback option rather than the BC break, because that fallback is older than this PR: `NotLoadableException` already returned the default image in debug mode before I touched anything. Routing it through the helper as-is would not have changed my new code, it would have changed a long-standing behaviour for people who did not ask for it.

So `createDefaultImageResponse()` takes an `$evenInDebug` flag, which that one call site passes. You get the single code path and the logging, and nobody's debug setup changes. I added `testNotLoadableSourceStillRedirectsToDefaultImageInDebugMode` since nothing covered that guarantee.

Controller/ImagineController.php
Amoifr: Done: `filter` and `message` are context values now, with the exception alongside them.

Controller/ImagineController.php
Amoifr: Routed through the helper, with an `$evenInDebug` flag rather than the BC break. This fallback predates the debug check and already applied in debug mode, so the flag keeps it exactly as it was while still getting the logging and dropping the duplication. Covered by a new test.

Controller/ImagineController.php
```diff
@@ -151,18 +160,54 @@ private function createRedirectResponse(\Closure $url, string $path, string $fil
         try {
             return new RedirectResponse($url(), $this->controllerConfig->getRedirectResponseCode());
         } catch (NotLoadableException $exception) {
-            if (null !== $this->dataManager->getDefaultImageUrl($filter)) {
-                return new RedirectResponse($this->dataManager->getDefaultImageUrl($filter));
+            // this fallback predates the debug mode check, so it keeps applying in debug mode
```
dbu, line 163:
```suggestion
            // this fallback predates the debug mode check, so it keeps applying in debug mode
            // TODO next major version: change to throw exception in debug mode and get rid fo the evenInDebug flag
```

dbu (approved):
makes sense, thanks!

i added a note that we want to remove the evenInDebug flag in the next major.

PR #1666, opened 2026-09-08
Dispatch filter lifecycle events

### LINKED ISSUE #1467: Add telemetry hooks
**Is your feature request related to a problem? Please describe.**
We've recently started using telemetry in production and discovered how long LiipImagineBundle filter sets can sometimes take to run. On closer inspection it's because some filters are conditional (e.g. image rotation only runs if the image needs rotating) and a couple of filters we've created are slow.

Telemetry told us the overall 'use this bundle' call was slow, but it would be useful to monitor the time each filter takes.

**Describe the solution you'd like**
I'd like some hooks to be present in LiipImagineBundle that we can subscribe to and use to create telemetry spans. We're using [Tideways](https://www.tideways.com) but others may use Opentelemetry, Datadog or want to push the details to Monolog.

_**Why have I created an issue and not a pull request?**_
There are many ways this could be implemented and I haven't seen the PHP community reach any consensus on how they'd like to support telemetry in libraries (or to think about it). AFAIK there's no equivalent in PHP for https://hexdocs.pm/telemetry/readme.html.

The bundle already uses Symfony's EventDispatcher so using this wouldn't add extra dependencies. There's a balance between emitting useful events and emitting too many, but I think the execution of filters/post-processors is core enough to make sense to record.

We edited `FilterManager::applyFilters()` and `FilterManager::applyPostProcessors()` to create telemetry spans around each filter call https://github.com/liip/LiipImagineBundle/blob/2.x/Imagine/Filter/FilterManager.php#L88

**Describe alternatives you've considered**
The alternatives we've used so far are:
1. Add the spans to our custom filters. Pro: no need to change the library. Con: we don't get insight into other filters
2. Edit LiipImagineBundle directly. Pro: we get full insight into filters. Con: we have to maintain a fork with the changes (actually: we didn't fork, we edited in place whilst testing 🙈

ousamabenyounes:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | Fix #1467
| License | MIT
| Doc | Resources/doc/events.rst

This adds side-effect-free lifecycle events around each image filter and post-processor. Event listeners receive the operation name and configured options, which lets telemetry integrations measure individual operations without changing filter behavior.

The event dispatcher remains optional for callers that instantiate `FilterManager` directly.

The focused test was copied onto the unmodified `2.x` base and failed behaviorally because no lifecycle events were dispatched:

```text
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
-    0 => 'liip_imagine.pre_filter'
-    1 => 'filter'
-    2 => 'liip_imagine.post_filter'
-    3 => 'liip_imagine.pre_post_processor'
-    4 => 'post_processor'
-    5 => 'liip_imagine.post_post_processor'
+    0 => 'filter'
+    1 => 'post_processor'
```

With this change:

```text
OK (1 test, 5 assertions)
```

Full local validation suite on PHP 8.2 / Symfony 6.4:

```text
Container lint: OK
Tests: 1074, Assertions: 2693, Skipped: 23
PHP-CS-Fixer: 0 of 304 files can be fixed
PHPStan: No errors
```

### DIFF
--- a/Resources/config/imagine.php
+++ b/Resources/config/imagine.php
@@ -217,6 +217,7 @@ return static function (ContainerConfigurator $container) {
             service('liip_imagine.filter.configuration'),
             service('liip_imagine'),
             service('liip_imagine.binary.mime_type_guesser'),
+            service('event_dispatcher'),
         ]);
 
     $services->alias(FilterManager::class, 'liip_imagine.filter.manager');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ for a given releases. Unreleased, upcoming changes will be updated here periodic
 
 ## 2.18.0 (unreleased)
 
+- Dispatch filter lifecycle events for telemetry and debugging ([ousamabenyounes](https://github.com/liip/LiipImagineBundle/pull/1666))
 - Fall back to the `default_image` when a filter does not exist or the image can not be generated, unless the new `controller.debug` option is enabled ([Amoifr](https://github.com/liip/LiipImagineBundle/pull/1665))
 - Dispatch pre/post store and pre/post remove events on the CacheManager, so applications can react to the cache storage lifecycle ([ousamabenyounes](https://github.com/liip/LiipImagineBundle/pull/1664))
 - Fix bug that we did not delete the webp variant when enabled, when deleting an image through the CacheManager ([ousamabenyounes](https://github.com/liip/LiipImagineBundle/pull/1661))
--- a/Events/FilterEvent.php
+++ b/Events/FilterEvent.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
+ * (c) https://github.com/liip/LiipImagineBundle/graphs/contributors
+ *
+ * For the full copyright and license information, please view the LICENSE.md
+ * file that was distributed with this source code.
+ */
+
+namespace Liip\ImagineBundle\Events;
+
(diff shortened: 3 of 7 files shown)


Resources/config/imagine.php
```diff
@@ -217,6 +217,7 @@
             service('liip_imagine.filter.configuration'),
             service('liip_imagine'),
             service('liip_imagine.binary.mime_type_guesser'),
+            service('event_dispatcher'),
```
dbu, line 220: to avoid all BC break, we should make this optional so its null if the event_dispatcher service does not exist.

dbu (changes requested):
i am not sure how useful this is, but compared to the cost of applying any filter, the cost of calling an event dispatcher is very small, so i don't see harm in it.

i have one small correction, otherwise this looks correct to me.

Resources/config/imagine.php
ousamabenyounes: Thanks @dbu — addressed in `4952ca9` by making the `event_dispatcher` reference optional with `nullOnInvalid()`. The dependency-injection regression test now covers the missing-service case.

ousamabenyounes:
Thanks for the review @dbu. Addressed the requested backward-compatibility change in `4952ca9`: the `event_dispatcher` reference is now optional via `nullOnInvalid()`, with a regression test for the missing-service case.

Validation: all GitHub checks on `4952ca98a5bfc1b077cad9ccaee2d39090dcb537` completed successfully. Re-review welcome.

Issue #1654, opened 2026-03-27
Documentation Issue: Transparency Parameter Behavior


lucacirannacopiaincolla:
I was reviewing the documentation on how to create background filter icons, specifically regarding setting a transparent background [here](https://symfony.com/bundles/LiipImagineBundle/2.x/filters/general.html#background)

I noticed that the documentation states:
“transparency: integer — Sets the background alpha value. The value should be within a range of 0 (opaque) - 100 (fully transparent).”

However, when I set the value to 100, the background still appeared with the default white color. On the other hand, setting the transparency value to 0 actually made the background transparent, which is the opposite of what the documentation indicates.

I wanted to report this in case the documentation needs to be updated.

Thanks a lot for your work!

Best regards

dbu:
thank you for reporting this. do you know if it is the same with all image libraries supported by Imagine (GD, Imagick and Gmagick)? if there is a discrepancy between the libraries, that could explain why the documentation is wrong. if we are sure its wrong for all libraries, we can simply fix the documentation. otherwise its tricky to fix as inverting it depending on the used library, while legally being a bug fix would be experienced as an unexpected BC break for users of the library that suddenly has its transparency inverted.

PR #1667, opened 2026-09-10
Fix the documented range of the background filter transparency

Amoifr:
| Q | A
| --- | ---
| Branch? | 2.x
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #1654
| License | MIT
| Doc | yes

@dbu you asked on #1654 whether the behaviour is the same across the Imagine drivers, since that decides between fixing the documentation and fixing the code. It is the same, so this fixes the documentation.

The reporter is right: `0` is the transparent end and `100` the opaque one, the opposite of what the page says. Omitting the option leaves the background opaque.

Measured on GD, running the filter and reading the alpha of a background pixel:

| `transparency` | resulting alpha |
| --- | --- |
| omitted | 100 (opaque) |
| 0 | 0 (fully transparent) |
| 50 | 50 |
| 100 | 100 (opaque) |

And the three drivers agree on the direction, since `BackgroundFilterLoader` hands the option straight to `palette()->color($color, $alpha)` and the semantics are Imagine's from there on:

* GD: `imagecolorallocatealpha(..., round(127 * (100 - $color->getAlpha()) / 100))`, and GD's own alpha runs from 0 (opaque) to 127, so an alpha of 100 comes out opaque.
* Imagick: `$pixel->setColorValue(\Imagick::COLOR_ALPHA, $color->getAlpha() / 100)`, where the ImageMagick alpha channel is 1.0 for opaque.
* Gmagick: throws `alpha transparency is not supported` for a value below 100 on a palette without alpha, so 100 is the opaque end there too.

So the value is an alpha, which reads as an opacity, despite the option being called `transparency`. I said as much on the page rather than only flipping the two words, since the name will keep surprising people.

The filter had no test at all, so the direction is pinned by one now. It uses the real GD driver, like `GrayscaleFilterLoaderTest` and the two `FloatToIntCast...` tests already do.

One thing I could not check and would rather not write into the documentation: `Gmagick\Imagine::create()` builds the canvas with `$pixel->getcolor(false)`, without the alpha component, and only then calls `setimagebackgroundcolor()` with the full pixel. I do not have the extension to tell whether a partially transparent background actually comes out of it. If it does not, that is worth a note on the page, and you or someone with gmagick would know in a second.

### DIFF
--- a/Resources/doc/filters/general.rst
+++ b/Resources/doc/filters/general.rst
@@ -62,7 +62,9 @@ Background Options
     ``bottom``, and ``bottomright``.
 
 **transparency:** ``integer``
-    Sets the background alpha value. The value should be within a range of 0 (opaque) - 100 (fully transparent).
+    Sets the background alpha value, from ``0`` (fully transparent) to ``100`` (opaque).
+    The background is opaque when the option is omitted. Despite its name the value is an
+    opacity, because it is handed to Imagine as the alpha of the background color.
 
 
 .. _filter-grayscale:
--- a/Tests/Imagine/Filter/Loader/BackgroundFilterLoaderTest.php
+++ b/Tests/Imagine/Filter/Loader/BackgroundFilterLoaderTest.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the `liip/LiipImagineBundle` project.
+ *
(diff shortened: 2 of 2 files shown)


dbu (approved):
thanks for clarifying this

